PR #7, opened 2020-05-08
Sorting

lucguillemot:


app/domain/config-types.ts
```diff
@@ -119,14 +132,27 @@ export type BarConfig = t.TypeOf<typeof BarConfig>;
 
 const ColumnFields = t.intersection([
   t.type({
-    x: GenericField,
+    x: t.type({
+      componentIri: t.string,
+      sorting: t.type({
```
herrstucki, line 137: Beware! Making `sorting` non-optional makes existing chart configs invalid, thus breaking all published charts!

app/domain/config-types.ts
```diff
@@ -119,14 +132,27 @@ export type BarConfig = t.TypeOf<typeof BarConfig>;
...
+        sortingField: t.union([t.literal("alphabetical"), t.literal("y")]),
```
herrstucki, line 138: I was confused by `alphabetical` here (because it implies order). Shouldn't the _field_ be named `label` or `categoryLabel` instead?

app/domain/config-types.ts
```diff
@@ -119,14 +132,27 @@ export type BarConfig = t.TypeOf<typeof BarConfig>;
...
+        sortingOrder: t.union([t.literal("asc"), t.literal("desc")]),
+      }),
+    }),
     y: GenericField,
   }),
   t.partial({
     segment: t.type({
-      componentIri: t.string,
+      componentIri: t.union([t.string, t.undefined]),
```
herrstucki, line 146: Not sure if it's correct to let the `componentIri` be optional here. The whole `segment` field is already optional, and I don't think it makes much sense to have `segment` options without a componentIri.

lucguillemot: The reason I did that is to have a default sorting type and sorting order. In the stacked column chart, if selecting a dimension for the "partition", then "None", then a dimension again, the app was previously crashing because no default was set for the sorting options. I updated the action `CHART_OPTION_CHANGED` consequently.
I actually now updated this type back to `string`, and use in the app either a `dimensionIri` or `FIELD_VALUE_NONE`... I think we should actually go all the way and make segment mandatory (not a `partial`).

lucguillemot: Oh... but I see your previous point about not breaking previous chart configs...

app/domain/config-types.ts
lucguillemot: I now made all sorting-related options optional. Thanks for the reminder 😅 .

app/components/chart-columns.tsx
```diff
@@ -84,7 +84,10 @@ export const ChartColumns = memo(
   }) => {
     return (
       <>
-        {fields.segment && fields.segment.type === "stacked" ? (
+        {/* FIXME: These checks should probably be handled somewhere else */}
+        {fields.segment?.componentIri &&
+        fields.segment?.componentIri !== FIELD_VALUE_NONE &&
```
herrstucki, line 89: Just wondering: is this check even necessary? I would assume that when "none" is selected the whole `segment` field is removed from the config?

app/components/chart-options-selector.tsx
```diff
@@ -91,12 +99,12 @@ const ActiveFieldSwitch = ({
   );
 
   // It's a dimension which is not mapped to an encoding field, so we show the filter!
-  // FIXME: activeField and encodingField should match? remove type assertion
+  // FIXME: activeField and encodingField should match? to remove type assertion
   if (
     !encodings.map((e) => e.field).includes(activeField as EncodingField) &&
```
herrstucki, line 104: Maybe simplify this to `encodings.find(e => e.field === activeField)` to save a temporary array allocation.

app/components/chart-options-selector.tsx
```diff
@@ -198,7 +219,86 @@ const EncodingOptionsPanel = ({
...
+  disabled = false,
+}: {
+  state: ConfiguratorStateConfiguringChart;
+  field: string;
+  encodingSortingOptions: EncodingSortingOption[];
+  disabled?: boolean;
+}) => {
+  const activeSortingType = get(
+    state,
+    `chartConfig.fields.${field}.sorting.sortingType`,
```
herrstucki, line 235: Just checking because I remember there was an issue with this in the past: will that not break if `field` contains a `.` for some reason? Probably better use an array as path: `[ "chartConfig", "fields", field, "sorting", "sortingType"]`

app/components/charts-generic/areas/areas-state.tsx
```diff
@@ -56,6 +71,9 @@ const useAreasState = ({
 
   const width = useWidth();
 
+  const hasSegment =
+    fields.segment && fields.segment.componentIri !== FIELD_VALUE_NONE;
```
herrstucki, line 75: See comment above: is there a point in having a `segment` field with value none?

janwaechter:
As discussed this morning, I think we can leave the sorting of grouped column-charts as an option for the user — however, I'm not sure if the current wording is ideal in the case of sorting a grouped column-chart by value. For example in the screenshot below the options are «Largest Bottom» and «Largest Top», which does not seem to make too much sense in this case (the charts being grouped horizontally).

screenshot_grouped_sorted-by-value

janwaechter:
@lucguillemot I assume that some of these «slices» are so small that they are not really visible (e.g. the four first elements/colors of the legend) — correct?

screenshot_values-too-small-to-display

janwaechter:
@lucguillemot When activating the partitioning of a column-chart, the individual dimension values are automatically sorted by «Total size» with «Largest Bottom» selected — this is great.
However, when I now manually switch to sorted by «Name», the option «Z->A» seems to be the default option. I'm not sure if this makes sense, esp. as for the x-Axis «A→Z» seems to be default direction when sorting by «Name»? 
Thus, I would probably suggest to always sort by «A→Z» by default…unless I'm missing something? 😅

lucguillemot:
Thank you for your comments @janwaechter. 
I changed the wording for grouped columns to "Largest first" and "Largest last". 
In the pie chart, the dimensions that are not displayed have negative values. This a problem we should solve, at least visually (disabling the items in the right panel? grey out the items in the legend?), but I think it belongs top another PR, what do you think?

janwaechter:
Ah yes, in that case the should probably be another story/issue — I was thinking that maybe they are just too small to be shown… 😅
Thank you for clarifying @lucguillemot ! 😊

PR #11, opened 2020-05-18
Color customization

lucguillemot:


janwaechter:
Some initial comments regarding the color-picker implementation:
- [ ] Would it be possible to auto-complete a shortened hex-code from e.g. `#f66` to `#ff6666`?
- [ ] Add a hover effect to the color-swatches, e.g. same outline as in highlight/active state.
- [ ] The checkbox and the color swatch (in the filter-panel) should be vertically center-aligned.

herrstucki:
@janwaechter the 1st suggestion would be super annoying when you want to keep typing 😅 eg. #f66789

janwaechter:
@herrstucki Ah sorry, I think probably need to clarify — I don’t mean «auto-complete», but rather on «Enter» or «Confirmation».

PR #17, opened 2020-05-27
Localize date and number formatting

herrstucki:
Replaces all uses of date/number formatting functions with localized ones. I created React hooks for both because the locale can then easily be read from context.

NB:
- Numbers are currently all formatted with the de-CH locale.
- Currencies are not handled since there's no formal definition of how they would be expressed in the Data Cube
- Date/time formatting is approximated to the shortest useful form using a multi-scale time format as described in [d3-time-format](https://github.com/d3/d3-time-format/blob/master/README.md), but since we lack working real-world datasets with anything other than day and year interval, it's not very thoroughly tested.
- Date/time is formatted differently on time axes (where they are abbreviated) and everywhere else. This is probably not bullet-proof in some edge cases and will need to be verified in the future.

### DIFF
--- a/app/components/charts-generic/axis/axis-width-time.tsx
+++ b/app/components/charts-generic/axis/axis-width-time.tsx
@@ -2,20 +2,18 @@ import { axisBottom } from "d3-axis";
 import { select, Selection } from "d3-selection";
 import * as React from "react";
 import { useEffect, useRef } from "react";
-import { useChartTheme } from "../use-chart-theme";
-import { useChartState } from "../use-chart-state";
-import { LinesState } from "../lines/lines-state";
-import { AreasState } from "../areas/areas-state";
-import { formatDateAuto } from "../../../domain/helpers";
+import { useFormatShortDateAuto } from "../../../domain/helpers";
 import { estimateTextWidth } from "../../../lib/estimate-text-width";
-import { max } from "d3-array";
+import { AreasState } from "../areas/areas-state";
+import { LinesState } from "../lines/lines-state";
+import { useChartState } from "../use-chart-state";
+import { useChartTheme } from "../use-chart-theme";
 
 export const AxisTime = () => {
   const ref = useRef<SVGGElement>(null);
+  const formatDateAuto = useFormatShortDateAuto();
 
-  const { xScale, yScale, bounds, xUniqueValues } = useChartState() as
-    | LinesState
-    | AreasState;
+  const { xScale, yScale, bounds } = useChartState() as LinesState | AreasState;
 
   const {
     labelColor,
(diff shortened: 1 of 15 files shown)


app/components/charts-generic/axis/axis-width-time.tsx
```diff
@@ -27,28 +25,23 @@ export const AxisTime = () => {
...
+  // Roughly equivalent to estimateTextWidth("99.99.9999", 12);
+  const maxLabelLength = 70;
 
   // This could be useful: use data points as tick values,
   // but it does not solve the problem of overlapping.
   // const tickValues =
   //   bounds.chartWidth / (maxLabelLength + 20) > xUniqueValues.length
   //     ? xUniqueValues
   //     : null;
-  const ticks = Math.min(
```
lucguillemot, line 41: This comparison is useful to prevent time labels for non existing data points, like months when there are only yearly values.

herrstucki: I found that it's not a very useful heuristic, since time is continuous and not ordinal, so the axis should offer a reference and not the distinct values. So I personally don't mind seeing months between the year ticks. After all, we show ticks at regular intervals other continuous axes too!

Also it gets more tricky: Let's say we have only two data points, at e.g. 2017-01-08 and 2018-01-12. If we limit the tick number to 2, the axis will only show one at 2008! We could argue to show the exact start and end but I think showing month intervals is more readable than using the same precision like the data.

Before (using the FOAG dataset):

Bildschirmfoto 2020-05-28 um 15 44 20

After:

Bildschirmfoto 2020-05-28 um 15 44 12

It may not be the _perfect_ way to label this axis but I think it's better to have more ticks.

lucguillemot: The analogy with a continuous quantitative axis is very convincing! You're right, let's have additional ticks to serve as a time reference. Maybe @janwaechter also has an opinion on this?

janwaechter: Interesting idea. I also agree that it looks better, but I also fear that it might lead people to assume that there are more data-points visible then there actually are, no?

Honestly, I guess in this case they shouldn't even use a line-chart, but rather a column chart. 😄

herrstucki: @janwaechter Yeah, the problem is that we can't switch the chart type easily (keeping the filters and settings) at the moment ;) Some filters obviously have more data points and some have none or just one (which also doesn't get displayed in a line chart!).

I firmly disagree though that ticks on a time axis should align with the data points. I think this would make the axis super confusing and too precise (and there would be no way to leave ticks out).

A solution to indicate _where_ data points are could be to display dots on top of the lines (like we do on hover). This would also address the "single data point" problem, at least it's not completely unusable.

janwaechter: @herrstucki True that, you're absolutely right — the problem are not the ticks themselves, but rather the uncertainty about where the data points are on said line…adding point-markers could help. Not sure if I like them from a «visual» perspective, though. 😅

PR #24, opened 2020-10-07
Fix misleading ui

lucguillemot:


app/components/chart-annotations-selector.tsx
```diff
@@ -22,9 +25,16 @@ export const ChartAnnotationsSelector = ({
...
   }, [state.activeField]);
+  const locale = useLocale();
 
   const af = state.activeField === "title" ? "title" : "description";
 
+  // Reorder locales so the input field for
+  // the current locale is on top
+  const oldPosition = locales.findIndex((l) => l === locale);
+  const _locales = [...locales];
+  const orderedLocales = moveItemInArray(_locales, oldPosition, 0);
```
herrstucki, line 36: This process seems a bit complicated to me. Can't we just do the following?

```ts
orderedLocales = [locale, ...locales.filter(l => l === locale)]
```

herrstucki (approved):
LGTM!

app/components/chart-annotations-selector.tsx
lucguillemot: Indeed 😅

PR #44, opened 2020-11-20
Interactive filters time

lucguillemot:


app/charts/area/areas-state.tsx
```diff
@@ -100,33 +103,73 @@ const useAreasState = ({
...
     [data, getX]
   );
+  const allDataGroupedMap = group(sortedData, getGroups);
+  const allDataWide = getWideData({
+    groupedMap: allDataGroupedMap,
+    getSegment,
+    getY,
+    xKey,
+  });
+  const xUniqueValues = sortedData
```
herrstucki, line 125: I might be wrong: but this looks like we're getting _all_ individual timestamps, which will most likely be slow/break on datasets that use a different interval than years.

Can't we use a comination of the extent and the time interval to make the brush "snap"? I don't think we should respect sparse datasets too much (i.e. ones where in-between values are missing like [2010, 2015, 2016, 2017])

app/package.json
```diff
@@ -29,6 +29,7 @@
     "cors": "^2.8.5",
     "d3-array": "^2.7.1",
     "d3-axis": "^2.0.0",
+    "d3-brush": "^2.1.0",
```
herrstucki, line 32: Sidenote: I think we need to make sure that we're using individual d3 package versions that belong together. Not all the types are up-to-date as well … I tried before but failed because @types/d3 wasn't at v6 back then, but it should be now.

PR #45, opened 2020-11-24
Interactive filters editor

lucguillemot:


app/configurator/components/field.tsx
```diff
@@ -2,7 +2,7 @@ import { t } from "@lingui/macro";
 import { I18n } from "@lingui/react";
 import { Box, Flex } from "@theme-ui/components";
 import get from "lodash/get";
-import { ChangeEvent, ReactNode, useCallback } from "react";
+import React, { ChangeEvent, ReactNode, useCallback } from "react";
```
herrstucki, line 5: Please don't use the default React export anymore, it will be removed in the future.

app/configurator/components/chart-annotator.tsx
```diff
@@ -1,31 +1,76 @@
...
+          <AnnotatorTabField
+            value={"description"}
+            icon="description"
+            label={getFieldLabel("description")}
+          ></AnnotatorTabField>
+        </ControlSectionContent>
+      </ControlSection>
+
+      {/* Filters */}
+      {segmentDimension && (
```
herrstucki, line 55: Why are we checking for the segment dimension here instead of the chart type?

Does the setting depend on if an actual segment dimension has been set before? What happens when I enable the filter and _then_ disable the segmentation?

PR #64, opened 2021-03-25
Use latest available cube version automatically

herrstucki:
After a visualization is created, it should not be locked to the data cube version current at the time of creation but automatically try to pick and display the latest version of the cube (preferring published over draft cubes).


herrstucki:
Note: I couldn't test the behavior yet because publishing new cubes is currently broken :sweat_smile:

PR #75, opened 2021-04-28
Unversioned Filter Values

herrstucki:


herrstucki:
This is currently blocked because adding a lookup dimension for `schema:sameAs` on all filtered dimensions as in https://github.com/visualize-admin/visualization-tool/pull/75/commits/dbb23d1fc37a23e7fa29891183602bf9e7c83db7 will make the query return nothing when any of the dimensions don't have values with a sameAs counterpart – which is the case for:

- dimensions with literal values
- for managed dimensions (that are inherently unversioned)
- cubes that were not generated by the Cube Creator (maybe)

To solve this:

1. Either dimensions need to be marked explicitly as having versioned values with sameAs
2. Somehow the query library could make lookup dimensions and their filter optional.

l00mi:
What changed since you reported, that all new cubes should dimensions have be marked with type `cube:SharedDimension` which means they are not versioned per definition.

On the other dimensions if they come from the cube creator they are always versioned, by contract. But it is true that other cubes generated by other means could provide non-versioned cubes. @ktk and @bergos would it make sense to additionally explicitly mark a cube or a dimension as "versioned"?

l00mi:
@herrstucki would you prefer this "per" Cube, or "per" Dimension? or both?

l00mi:
As proposed we add now schema:version per Dimension if not SharedDimension and not a Literal. See https://s.zazuko.com/AsXbi for an example.

PR #77, opened 2021-05-02
feat: Time Range Filters

### LINKED ISSUE #69: Filter on date
This problem is maybe already known: When you want to filter the data on a time variable (year, date), only the first and last date of the series are available in the filter.  The dates between the oldest and the most recent are however displayed in the graph. See dataset Penguins.
image

herrstucki:
Fixes #69


herrstucki:
@lucguillemot The case where a temporal dimension is only used as a filter is not handled. I'm not exactly sure what to do there because we can't just put all possible date values into a dropdown menu …

PR #138, opened 2021-10-13
feat: Display more information in the dataset preview

### LINKED ISSUE #135: Show more data cube metadata
- contact email (cf. https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/304)
- link to landingpage (for methodology)

bprusinowski:
Closes #135.

This PR adds the email & landing page URL to the dataset preview.

- If there is an `contactEmail` attached to the cube, then the the email field will contain a mailto: href; the label will be either a `contactName` (if exists), otherwise it'll be the `contactEmail`.
- If there is a `landingPage` attached to the cube, then the landingPage field will contain a href and a label (both equal to the `landingPage`).

### DIFF
--- a/app/components/chart-footnotes.tsx
+++ b/app/components/chart-footnotes.tsx
@@ -52,26 +52,11 @@ export const ChartFootnotes = ({
             <>
               <Box sx={{ display: "inline", mx: 1 }}>·</Box>
               <Link
+                variant="inline"
+                sx={{ mt: 2 }}
                 href={shareUrl}
                 target="_blank"
                 rel="noopener noreferrer"
-                sx={{
-                  display: "inline",
-                  textDecoration: "none",
-                  color: "primary",
-                  textAlign: "left",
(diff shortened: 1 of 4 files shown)


herrstucki:
@bprusinowski The landing page info should be explicitly part of the metadata. In fact, it's already parsed here: https://github.com/visualize-admin/visualization-tool/blob/20e5ec2d4e8b59053e4ba98ef0ee8a69b1fea84c/app/rdf/parse.ts#L60 Beware that it can be optional (contact point potentially as well?), so make sure to display "–" when that's the case, like in the other fields.

bprusinowski:
@herrstucki, I guess the links shouldn't look like this – should the style from e.g. "Download data" below a published chart be applied there?

Screenshot 2021-10-13 at 14 39 19

ptbrowne:
Thanks 🎉

PR #139, opened 2021-10-13
Update Cube Versions

### LINKED ISSUE #76: Enable Automatic Upgrading of Cube Versions
Currently, a cube's version is locked to the one selected when the visualization was created.

The logic is already implemented in https://github.com/visualize-admin/visualization-tool/blob/6d72bb37deb55fcbe3f6da4ce246a1bb71284366/app/rdf/queries.ts#L96-L162

It is disabled because this functionality depends on #75 to work (otherwise versioned dimension values are used as filters which will break newer cubes). See https://github.com/zazuko/cube-creator/issues/658

### LINKED ISSUE #116: Visualize does not display the last version of the data
When new data are added to a cube, an already published charts does not see the new data.
This a chart published with the first version of the cube (only data of 2019) : [2019](https://www.test.visualize.admin.ch/de/v/rDnxj5kg9_Vb)
This a chart published after some new data (2020) were added in the same cube : [2019+2020](https://www.test.visualize.admin.ch/de/v/gdDLrftmJuu3)

Both charts should show the same data, but the first chart does not see the new data.
If the first chart were updated, the title will not update, that's another problem the user has to solve himself.

jstcki:
Closes  #76
Closes #116

Also implements changes related to how cubes are marked as expired.

### DIFF
--- a/app/rdf/parse.ts
+++ b/app/rdf/parse.ts
@@ -23,6 +23,13 @@ export const getQueryLocales = (locale: string): string[] => [
   "",
 ];
 
+export const isCubePublished = (cube: Cube): boolean =>
+  cube
+    .out(ns.schema.creativeWorkStatus)
+    .terms.some((t) =>
+      t.equals(ns.adminVocabulary("CreativeWorkStatus/Published"))
+    );
+
 export const parseCube = ({
   cube,
   locale,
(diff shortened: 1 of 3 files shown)


herrstucki:
Going to merge, so it's ready to test. Opened another issue #143 for adding tests.

PR #145, opened 2021-10-18
feat: Consider InteractiveFilters during data download

### LINKED ISSUE #142: Published chart - download data do not take into account the current filter settings
On this chart: https://test.visualize.admin.ch/fr/v/JC3mlYyLWQ2e
I can use the "download data" option.
But no matter what I select in the filter, for the "status" for example, the download is always about the "endangered" status.
Is it a normal behavior ?

bprusinowski:
Closes #142.

Currently, interactive filters do not affect the data that is being downloaded, and that might be confusing for users. This PR fixes that, so the `stuff_visible_on_the_chart` is always equal to the `stuff_downloaded_to_csv`.

PS. Not sure if we should change this yet, as @herrstucki didn't have a chance to go through the #142 and the current behavior might be the expected behavior.

### DIFF
--- a/app/charts/shared/chart-helpers.tsx
+++ b/app/charts/shared/chart-helpers.tsx
@@ -16,7 +16,7 @@ export const useQueryFilters = ({
   interactiveFiltersIsActive,
 }: {
   chartConfig: ChartConfig;
-  interactiveFiltersIsActive: boolean;
+  interactiveFiltersIsActive: boolean | undefined;
 }): Filters | FilterValueSingle => {
   const [IFstate] = useInteractiveFilters();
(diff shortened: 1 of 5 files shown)


ptbrowne:
It looks like the default behavior of useQueryFilters if the prop interactiveFiltersIsActive is not passed should be to take the value from inside the chartConfig. And users of useQueryFilters would only have to pass itneracriveFiltersIsActive if they want to override the value from chartConfig. I think it's pretty inconvenient that the users of useQueryFilters have to know about the nested structure of chartConfig.

Anyway, maybe that can be a further improvement since the main PR deals with the original problem.

PR #148, opened 2021-10-19
Properly check for undefined data points in the area chart generator

bprusinowski:
Currently, when we would like to create a stacked area chart, which is based on a dataset that contains some missing values, we could see e.g. this:

Screenshot 2021-10-19 at 11 47 21

Repro: https://dev.visualize.admin.ch/en/v/Jcy9ujD2SmCX

This happens due to the fact, that the D3's stack method generates `NaN` values, when there is no data for some category in a given point of time, and in the area generator we check for `null`, which doesn't work.

This PR fixes that by checking if a value is a number.

Chart after fixing:

Screenshot 2021-10-19 at 11 50 19

### DIFF
--- a/app/charts/area/areas.tsx
+++ b/app/charts/area/areas.tsx
@@ -1,5 +1,6 @@
 import { area } from "d3";
 import { memo } from "react";
+import { isNumber } from "../../configurator/components/ui-helpers";
 import { useTheme } from "../../themes";
 import { useChartState } from "../shared/use-chart-state";
 import { AreasState } from "./areas-state";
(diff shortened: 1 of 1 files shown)


ptbrowne:
I personally think using 0 as a fill value for missing data looks like it is not biaised but it is actually problematic : what I do not like with this method is that when you add a series with different sampling time to the chart, the previous area might look completely different due to time holes being added to the previous series.

- Using the mean of the series looks like it would be more unbiased than using 0
- Using forward filling or back filling looks like it is a standard method for dealing with missing data https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.fillna.html
- Interpolating is already used when using a stacked area chart, I think it would be also valid as a method for filling NAs as the area would not change size

bprusinowski:
Yeah, filling the series with 0s can introduce "spikes" in the data, as seen on the screenshots above, which is suggesting that the values for a given point in time were equal to 0. Any kind of interpolation could also yield values that aren't equal to actual, unknown values, but it seems to be closer to them, and as @ptbrowne indicated, this is a common approach to fill in the missing data, used a lot in machine learning, for example 😀.

The point about a lot of gray stacked areas is valid, I'm also not sure about the appearance of that.

We talked about that with @ptbrowne before our vacations and prepared a rough draft in Observable ([link](https://observablehq.com/d/7c02dc8d9a3160e9)), to illustrate how it might look like (the dots might be some way to indicate the "interpolated" values?).

bprusinowski:
Closed by #206

Issue #87, opened 2021-05-21
Provide a link to pre-select a cube dataset.


l00mi:
To be able to have a direct link to a published cube inside visualze.admin.ch from:
* https://github.com/zazuko/cube-creator
* http://lindas.admin.ch/dataset (see also https://github.com/zazuko/cube-creator/issues/730)
* and also potentially opendata.swiss

It would be necessary to be able to compose a URI based on the Cube Dataset Uri. Ideally in a non-versioned and versioned fashion.

As illustrative proposition:
`https://visualize.admin.ch/en/create/new/https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fudb066`
and
'`https://visualize.admin.ch/en/create/new/https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fudb066%2F1`
 for versioned.

herrstucki:
@l00mi Let's use a `cube` query param, e.g. `https://visualize.admin.ch/en/create/new?cube=https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fudb066`. That way we have the possibility to add more options later

herrstucki:
We need to define if both unversioned and versioned IRIs need to be possible. Some options:

- only use the versioned one (which we use now already as starting point to update to a newer version)
- use different params, e.g. `cube` and `cubeVersioned`, so Visualize can know what to expect
- use an unversioned `cube` param and an optional `version` param (not sure if I like this one, seems more complex to query)

l00mi:
I propose to go step, by step.

Lets define the un-versioned variant first, e.g. as you have proposed above. If there is a need to have a deep link to the versioned later on, we can re-evaluate based on the real need? If you think it's better to touch this only once lets go for the 2nd variant?

herrstucki:
Step by step sounds good to me! This shouldn't be a big deal on our side but maybe we can synchronize it with the changes related to validThrough/expires and #76 so we can complete this in one go.

ptbrowne:
@l00mi On our end, it was much simpler not to introduce the unversioned IRI concept (otherwise we would have to introduce this unversioned iri in all our resolvers), and support only the version iri. This is done here : https://github.com/visualize-admin/visualization-tool/pull/170. This means that the cube creator should use the versioned iri. We also think that it would make sense from the cube creator UI point of view since the links are available from a page that shows a specific version anyway, so linking to a page that uses a specific dataset version makes sense.

[...]

PR #150, opened 2021-10-19
Abiity to create a chart from an existing dataset

ptbrowne:
When a user goes to `/chart/new?dataset=<datasetIri>`, the cube metadata is fetched
and we go directly to the second step. If the cube metadata cannot be fetched, we
stay in first step where the user has to select a dataset.

This could be useful to entice users to create charts from an existing dataset or
for tests.

Closes https://github.com/visualize-admin/visualization-tool/issues/87.

### DIFF
--- a/app/configurator/configurator-state.spec.tsx
+++ b/app/configurator/configurator-state.spec.tsx
@@ -1,11 +1,13 @@
+import { Client } from "@urql/core";
 import {
   getLocalStorageKey,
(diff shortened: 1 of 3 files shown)


herrstucki:
@ptbrowne Currently, the Cube Creator links to `/create/new?cube=<versionHistoryIri>` (as described in #87). So the param and the logic don't quite match yet 😅

PR #159, opened 2021-10-21
Refactor published chart

ptbrowne:
Smallish refactor on the published chart file to prevent code duplication and drift.

Since it can be a bit "invasive" and since I had to had a @ts-ignore, I prefer to keep this commit out of the next release.


ptbrowne:
I like it thanks ! I think here the problem was that since the usage of chartConfig was not explicited, typescript could not understand that the chart config type was the right one for the component used.

herrstucki (approved):

PR #164, opened 2021-10-22
Improve debug panel

ptbrowne:
- Extract debug panel from panel-middle
- Add interactive filters to what's viewable in the debug panel
- Use react-inspector instead of pre + JSON.stringify for better developer experience


ptbrowne:
Thanks a lot, it looks much better :100: I replaced the link to cube viewer by a helper text if no dataset have been selected.

ptbrowne:
I merged this in main through https://github.com/visualize-admin/visualization-tool/commit/40a51c46cc31bfd1b93f4682a4aa67122e3bdb85

PR #169, opened 2021-11-01
fix: On published chart, allow to select none in interactive filters

### LINKED ISSUE #157: Can't select "No filter" in interactive, optional filters with a default value
When creating a chart, users have a possibility to select optional filters, which then can be added to a published chart as interactive filters. In such scenario, if user selected some value in the optional filter when creating a chart, the "No filter" option can't be selected after publishing the chart.

See https://dev.visualize.admin.ch/en/v/_Pnr3KRM3FX0, and try to select "No filter" in Land use.

It works correctly for optional filters with no default value set during a chart creation process, e.g. Remarks.

ptbrowne:
Closes #157

Instead of deleting the value of the interactive filter when selecting none,
we have to keep the none value in the interactive filter state, otherwise the
value taken is from configurator state = the user selects "None" and the
state of the select goes back to the value from configurator state; and the
chart data is still filtered.
Since None is kept inside the state, we need to filter it before sending
the actual query, hence the changes in preview/published/download.


ptbrowne:
As discussed :

- [x] Added tests
- [x] filterNone is now the default behavior and there is no more option to change it

herrstucki (approved):
I checked with the Bathing water quality dataset and it works perfectly!

PR #170, opened 2021-11-02
fix: Use "cube" parameter instead of "dataset" when creating chart from existing dataset

ptbrowne:
https://github.com/visualize-admin/visualization-tool/pull/150#issuecomment-956097226

> @ptbrowne Currently, the Cube Creator links to /create/new?cube=<versionHistoryIri> (as described in #87). So the param and the logic don't quite match yet 😅

Instead of using the "dataset" parameter in the URL, use the "cube" to match what the cube creator is using.


herrstucki (approved):
LGTM, although the Cube Creator still links to the IRI of the version history, not the versioned cubes themselves (maybe something they should change anyway? 🤔)

PR #172, opened 2021-11-02
Fix Column Chart display bugs

### LINKED ISSUE #166: Small display bug - negativ scale
Here is a test [chart](https://www.dev.visualize.admin.ch/en/v/-Pj2EI1eQL8v)
image

Everything's fine !
But with another filter :

image

or this one

image

bprusinowski:
Closes #166, #190.

This PR:
- fixes cropped Y axis ticks for column charts,
- introduces a "start with 0" constraint for Y axes with negative scales, which fixes bars that were overlapping axis titles,
- fixes min and max Y scale ranges for stacked column charts (previously, it was taking min and max of total, which was a sum of all positive and negative values in a given "column stack" – now it takes max(positive values in a stack) and min(negative values in a stack)),
- extracts a Column function (which was the same in 3 files) to a separate file.

### DIFF
--- a/app/charts/column/columns-grouped-state.tsx
+++ b/app/charts/column/columns-grouped-state.tsx
@@ -26,11 +26,10 @@ import {
 } from "../../configurator/components/ui-helpers";
 import { Observation, ObservationValue } from "../../domain/data";
 import { sortByIndex } from "../../lib/array";
-import { estimateTextWidth } from "../../lib/estimate-text-width";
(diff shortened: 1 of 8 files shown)


ptbrowne (commented):
Thanks, looks ok to me 👍 For this kind of logic/arithmetic bugs I think it would be helpful to have tests :)

ptbrowne (approved):

herrstucki (approved):
LGTM!

PR #173, opened 2021-11-02
Consider user-selected color in line and area charts for one-element selections

### LINKED ISSUE #171: Line colour does not match the legend colour
I have noticed several times that for line diagrams, the line colour does not match the legend colour.
This seems to be the case only when there is only one category.
image

bprusinowski:
Closes #171.

This PR makes the line and area charts consider user-selected colors when the selection is one-element long.

@herrstucki I'm not sure why the current behavior wasn't allowing that, as this was explicitly coded in the app – let me know if you think it shouldn't be changed :)

### DIFF
--- a/app/charts/area/areas.tsx
+++ b/app/charts/area/areas.tsx
@@ -20,7 +20,7 @@ export const Areas = () => {
           <Area
             key={`${d.key}-${i}`}
             path={areaGenerator(d) as string}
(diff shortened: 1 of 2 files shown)


ptbrowne (approved):
LGTM 👍 I wonder if there was a specific reason to use the primary color here 🤔

herrstucki:
Also looks good to me. I don't remember why this was a special case for the single segment/category before.

PR #175, opened 2021-11-02
Automerge

ptbrowne:
PRs whose checks pass and where the label automerge is present
should be automerged via this workflow.


herrstucki (approved):
Please automerge this ;)

PR #191, opened 2021-11-03
Fix hidden grouped-by columns in the Table chart

### LINKED ISSUE #188: Table Chart - Grouping section - Unhide column
If a column was hidden and then dragged into the section "Groupings", the checkbox "Hide column" should be selected by default. However, this is not the case and the user can't unhide it without having to ungroup it again first.

Screenshot_2021-02-18_at_15 14 18

bprusinowski:
Closes #188.

This PR:
- keeps the hidden state when using a hidden column as a group (as suggested in the issue raised by @AnninaWalker),
- allows to hide/unhide key dimensions used as group-by columns,
- shows a "Use as group" checkbox for non-key dimensions that are not of a "Measure" type (previously anyway we could use them as grouping variables by drag&drop), but this might be more convenient and consistent with how it's dealt with in case of key dimensions.

@herrstucki previously it was coded to clear the "isHidden" prop, so let me know if you think it should be the way the PR proposes – but it feels more natural to keep the state IMO, as anyway I had to hide a column first and then use it as a group (the other way I see it could be to always clear the "isFiltered" prop when using a key dimension to group the data by and disable the ability to "Hide and filter column").

### DIFF
--- a/app/charts/index.ts
+++ b/app/charts/index.ts
@@ -4,11 +4,11 @@ import {
   GenericFields,
   TableColumn,
 } from "../configurator";
-import { getCategoricalDimensions, getTimeDimensions } from "../domain/data";
 import { mapColorsToComponentValuesIris } from "../configurator/components/ui-helpers";
+import { getCategoricalDimensions, getTimeDimensions } from "../domain/data";
+import { DimensionMetaDataFragment } from "../graphql/query-hooks";
 import { DataCubeMetadata } from "../graphql/types";
 import { unreachableError } from "../lib/unreachable";
-import { DimensionMetaDataFragment } from "../graphql/query-hooks";
 
 export const enabledChartTypes: ChartType[] = [
   // "bar",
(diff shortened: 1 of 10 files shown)


herrstucki (commented):
OK … so apparently this fixes one thing but breaks another 😅 

Let me try to describe the logic:

- You shouldn't be able to group or filter on measure columns – just hide them (in case there's multiple ones) → this seems okay
- On **Key Dimensions** must either be
  - grouped by and (optionally) hidden; the filter list should be a multi-select because you still may want to exclude certain options → in this case, the checkbox now says incorrectly "filtered and hidden"
  - not grouped, not hidden
  - not grouped, hidden **and filtered on a single value**, otherwise we remove the ability to distinguish individual rows → on the main branch, the grouping checkbox is disabled but unfortunately it's possible to drag and drop to group, leading to a bad state (grouped, hidden, **and** filtered). I think that's why the `isHidden` property was reset whenever the grouping status changed but maybe there's a better way?
- **Non-Key Dimensions** should be either
  - not hidden, not grouped
  - grouped but not hidden
  - grouped and hidden
  - hidden but not grouped (no single value filter requirements as on the key dimensions because per definition these are not used to distinguish rows)

herrstucki:
Soo to summarize, I think this PR is good but grouped key dimensions should _not_ use `isFiltered` but `isHidden` instead (or something like that) … and in the transition from grouped - ungrouped we need to convert between these two states

ptbrowne:
Do you think it would be possible to add tests here ? The logic seems quite complex and it would be nice to have a spec somewhere, at least for us to know in 6 months what is supposed to happen. IMHO as a general rule, we should not have a fix commit without a corresponding spec test.

bprusinowski:
I agree that we should have some tests there; will try to add them! Thanks for raising this 👍

ptbrowne (approved):
LGTM, thanks !

PR #194, opened 2021-11-04
Add Git Hash To Version

### LINKED ISSUE #167: Versions number handling - question
I talked yesterday in the meeting about the version number with Jeremy.
One thing looked "wrong", as the tool on dev (version 3.0.1 yesterday) seemed more recent than the tool on test (version 3.1.0).
One improvement that was available on dev, but not yet on test, is the fact that the date filters work on dev (the final user can display the filters, and change the date).
Today I see that both environmnent now have version 3.1.0, but I "guess" that they do not both use the same code, as the date filters are ok on dev (as expected), but still not ok on test.
My question: as I follow a few issues, and I would test if they are solved only if I see that the version has changed, could you handle those versions numbers more precisely ? Or am I wrong to rely on versions numbers ?
For instance, I fillow this issue: https://github.com/visualize-admin/visualization-tool/issues/162, and currently I can only create a new visualization and edit it to see if the fix is has been deployed or not.
Thank you for any feedback

### LINKED ISSUE #192: Include git commit info in version number
To indicate that the deployed versions can differ between different environments, even if the version _number_ can be the same, we should include the git hash in the version number, and even link to the commit in this repo.

To bake it into the app, we need to update the Dockerfile to use an argument like so:

```
docker build --build-arg COMMIT=$(git rev-parse HEAD)
```

In the Dockerfile, we expose it to the app with `NEXT_PUBLIC_COMMIT`, so it gets [inlined during build](https://nextjs.org/docs/basic-features/environment-variables#exposing-environment-variables-to-the-browser)

Then we can display it in the app like so:

> v3.1.1 (https://github.com/visualize-admin/visualization-tool/commit/8e64d532cdac3992d8dd927d3ca0d77805a5d20f)

herrstucki:
Adds the Git hash and repo via `NEXT_PUBLIC_` build-time env vars, so they can be displayed along with the version number.

- [x] Display hash and link to GH repo if env vars are present
- [x] Set `NEXT_PUBLIC_COMMIT` during Docker build

Fixes #192
Fixes #167

### DIFF
--- a/app/components/footer.tsx
+++ b/app/components/footer.tsx
@@ -4,30 +4,28 @@ import NextLink from "next/link";
 import { forwardRef, ReactNode } from "react";
 import contentRoutes from "../content-routes.json";
 import { useLocale } from "../locales/use-locale";
+import { COMMIT, GITHUB_REPO, VERSION } from "../domain/env";
 
 const Version = () => {
   let commitLink = null;
 
-  if (process.env.NEXT_PUBLIC_GITHUB_REPO && process.env.NEXT_PUBLIC_COMMIT) {
+  if (GITHUB_REPO && COMMIT) {
     commitLink = (
       <>
         (
-        <Link
-          variant="primary"
-          href={`${process.env.NEXT_PUBLIC_GITHUB_REPO}/commit/${process.env.NEXT_PUBLIC_COMMIT}`}
-        >
-          {process.env.NEXT_PUBLIC_COMMIT.substr(0, 7)}
+        <Link variant="primary" href={`${GITHUB_REPO}/commit/${COMMIT}`}>
+          {COMMIT.substr(0, 7)}
         </Link>
         )
       </>
     );
-  } else if (process.env.NEXT_PUBLIC_COMMIT) {
-    commitLink = `(${process.env.NEXT_PUBLIC_COMMIT.substr(0, 7)})`;
+  } else if (COMMIT) {
+    commitLink = `(${COMMIT.substr(0, 7)})`;
   }
 
   return (
     <>
-      {process.env.NEXT_PUBLIC_VERSION} {commitLink}
+      {VERSION} {commitLink}
     </>
   );
 };
--- a/app/components/footer.tsx
+++ b/app/components/footer.tsx
@@ -27,7 +27,7 @@ const Version = () => {
 
   return (
     <>
-      {process.env.VERSION} {commitLink}
+      {process.env.NEXT_PUBLIC_VERSION} {commitLink}
     </>
   );
 };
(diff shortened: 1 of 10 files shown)


app/components/footer.tsx
```diff
@@ -5,6 +5,33 @@ import { forwardRef, ReactNode } from "react";
 import contentRoutes from "../content-routes.json";
 import { useLocale } from "../locales/use-locale";
 
+const Version = () => {
+  let commitLink = null;
+
+  if (process.env.NEXT_PUBLIC_GITHUB_REPO && process.env.NEXT_PUBLIC_COMMIT) {
```
ptbrowne, line 11: In which case do we not have have GITHUB_REPO in next public variables ?
 Nit: for readability I think assign the env variables to local variables would be a tidbit better.

```
const githubRepo = process.env.NEXT_PUBLIC_GITHUB_REPO
```

ptbrowne: Ah maybe in preview deployments ?

herrstucki: We always have it, it's actually set in next.config.js and comes from package.json (like `NEXT_PUBLIC_VERSION`)

herrstucki: Regarding the nit: we actually have an `env.ts` file which reads and re-exports env variables. Maybe that would be the place to do this? I'm not super happy with that because currently it mixes runtime and build env vars. Since runtime env vars are only available server-side, this doesn't always work as expected.

What do you think about splitting that up into two files, e.g.

```ts
// env-build.ts
export const VERSION = process.env.NEXT_PUBLIC_VERSION;
// etc.

// env-server.ts
export const PUBLIC_URL = process.env.PUBLIC_URL;
// ...
```

ptbrowne: So it seems like the else if will never be entered right ?

ptbrowne: I like the idea of making the environment explicit.

herrstucki: > So it seems like the else if will never be entered right ?

You mean since GITHUB_REPO is always set we shouldn't have this condition in the first place? Maybe … I think since from the frontend perspective this is still optional, we should handle the case where it's undefined.

herrstucki:
@ptbrowne Contrary to what we discussed before, I opted for a simpler approach to expose the runtime env vars to the browser in https://github.com/visualize-admin/visualization-tool/pull/194/commits/75c7faf51786215bf0dfc730b3a954b623b61968. Since we have a _document.tsx, *all* the pages are always server-rendered anyway, so the env can be safely set on the window there. I don't think there's a need to separate the env.ts anymore, since all values should be available regardless of context. I replaced all uses of process.env in the app with imports from env.ts.

bprusinowski (approved):

herrstucki:
Just for the record, this doesn't work for the Heroku preview builds, as it's not possible to copy env variables just via configuration (I don't think we should add Heroku-specific code to the repo?).

image

(☝️ this doesn't work, `$SOURCE_VERSION` just shows up as a string)

ptbrowne (approved):
Thanks, LGTM !

PR #195, opened 2021-11-04
Replace examples on the home page

### LINKED ISSUE #82: Replace examples on home page
Currently, the home page example charts are not working. We need to replace them with new ones, ideally with static data, so they don't depend on live cubes.

bprusinowski:
Closes #82.

Side question: can I be sure that these examples will work when we publish to PROD? These datasets don't appear on that env and I'm not sure if we can fetch them from there.


herrstucki:
@bprusinowski The charts on test.visualize.admin.ch (i.e. https://test.visualize.admin.ch/de/v/scBBn-4e4o3p) should already come from the PROD Lindas env.

ptbrowne (approved):
LGTM, thanks

PR #198, opened 2021-11-09
Fix non-table filtering issue

bprusinowski:
I have not considered all possible situations in #191 – I assumed that we don't need to differentiate between tables and other charts when dealing with filters, but in fact we do.

This PR fixes an issue currently available on DEV, when using a given dimension as a segment,  which removes currently applied single filter, and this behavior is not desired in this case.

I have also added some comments, to not forget in the future what's going on here (along with another test) 😆

### DIFF
--- a/app/configurator/configurator-state.spec.tsx
+++ b/app/configurator/configurator-state.spec.tsx
@@ -111,13 +111,24 @@ describe("deriveFiltersFromField", () => {
     __typename: "NominalDimension",
   } as DimensionMetaDataFragment;
 
-  const _applyDimensionToFilters = (
-    filters: any,
-    isField: boolean,
-    isHidden: boolean,
-    isGrouped: boolean
-  ) =>
-    applyDimensionToFilters(filters, dimension, isField, isHidden, isGrouped);
+  const _applyDimensionToFilters = ({
+    initialFiltersState,
+    isField,
+    isHidden,
+    isGrouped,
+  }: {
+    initialFiltersState: any;
+    isField: boolean;
+    isHidden: boolean;
+    isGrouped: boolean;
+  }) =>
+    applyDimensionToFilters({
+      filters: initialFiltersState,
+      dimension,
+      isField,
+      isHidden,
+      isGrouped,
+    });
 
   it("non-table: should remove single value filter when a dimension is used as a field", () => {
     const initialFiltersState = {
@@ -128,7 +139,12 @@ describe("deriveFiltersFromField", () => {
     };
     const expectedFiltersState = {};
 
-    _applyDimensionToFilters(initialFiltersState, true, false, false);
+    _applyDimensionToFilters({
+      initialFiltersState,
+      isField: true,
+      isHidden: false,
+      isGrouped: false,
+    });
     expect(initialFiltersState).toEqual(expectedFiltersState);
   });
 
@@ -141,7 +157,12 @@ describe("deriveFiltersFromField", () => {
       },
     };
 
-    _applyDimensionToFilters(initialFiltersState, true, true, false);
+    _applyDimensionToFilters({
+      initialFiltersState,
+      isField: true,
+      isHidden: true,
+      isGrouped: false,
+    });
     expect(initialFiltersState).toEqual(expectedFiltersState);
   });
 
@@ -159,7 +180,12 @@ describe("deriveFiltersFromField", () => {
       },
     };
 
-    _applyDimensionToFilters(initialFiltersState, true, true, false);
+    _applyDimensionToFilters({
+      initialFiltersState,
+      isField: true,
+      isHidden: true,
+      isGrouped: false,
+    });
     expect(initialFiltersState).toEqual(expectedFiltersState);
   });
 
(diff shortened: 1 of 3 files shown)


app/configurator/configurator-state.tsx
```diff
@@ -261,7 +272,7 @@ export const applyDimensionToFilters = (
     } else {
       // Add filter for this dim if it's not one of the selected multi filter fields, but
       // only when the dimension isn't grouped, as only then we need to switch to a single filter.
-      if (isHidden && !isGrouped) {
+      if ((isHidden || !isField) && !isGrouped) {
```
herrstucki, line 275: Nit: I find this condition quite hard to understand 😅 Is there a way to make it more clear or at least update the comment to describe the cases that are covered? But maybe it's just me 👴

herrstucki, line 275: Nit: I find this condition quite hard to understand 😅 Is there a way to make it more clear or at least update the comment to describe the cases that are covered? But maybe it's just me 👴

bprusinowski: I'll try to! 👨‍💻

herrstucki (approved):

app/configurator/configurator-state.spec.tsx
```diff
@@ -111,13 +111,26 @@ describe("deriveFiltersFromField", () => {
...
-    filters: any,
-    isHidden: boolean,
-    isGrouped: boolean
-  ) => applyDimensionToFilters(filters, dimension, isHidden, isGrouped);
+  const _applyDimensionToFilters = ({
+    initialFiltersState,
+    isField,
+    isHidden,
+    isGrouped,
+  }: {
```
herrstucki, line 119: Thank you for naming these options! Much clearer now than before!

bprusinowski:
Based on the fact that both of you mentioned that it's not very easy to understand what's going on there (and it also took me a while to get this) I have refactored the code and added the comments, so hopefully it'll be easier now.

app/configurator/configurator-state.tsx
```diff
@@ -202,82 +202,147 @@ export const getFilterValue = (
...
-        filters,
-        dimension,
-        isField: isField(dimension.iri),
-        isHidden: isHidden(dimension.iri),
-        isGrouped: isGrouped(dimension.iri),
-      })
-    );
+    const { chartType, fields, filters } = chartConfig;
+
+    if (chartType === "table") {
```
herrstucki, line 207: Yes, this makes it much easier to understand 👌

app/configurator/configurator-state.tsx
```diff
@@ -202,82 +202,147 @@ export const getFilterValue = (
...
-      // only when the dimension isn't grouped, as only then we need to switch to a single filter.
-      if ((isHidden || !isField) && !isGrouped) {
-        filters[dimension.iri] = {
-          type: "single",
-          value: dimension.values[0].value,
-        };
-      }
+        break;
+      default:
+        throw new Error();
```
herrstucki, line 280: Is this here to check that this case should in fact never happen? If so, I recommend using TypeScript to do the work for us instead of throwing at runtime: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#exhaustiveness-checking

app/configurator/configurator-state.spec.tsx
```diff
@@ -1,5 +1,8 @@
 import { Client } from "@urql/core";
-import { applyDimensionToFilters } from ".";
+import {
+  applyNonTableDimensionToFilters,
+  applyTableDimensionToFilters,
+} from ".";
```
herrstucki, line 5: Nit: shouldn't this be `from "./configurator-state"?

app/configurator/configurator-state.tsx
```diff
@@ -202,82 +202,147 @@ export const getFilterValue = (
...
+        if (!isField) {
+          // Range-filters are not allowed in the left panel.
+          filters[dimension.iri] = {
+            type: "single",
+            value: currentFilter.from,
+          };
+        }
+        break;
+      default:
+        throw new Error();
```
herrstucki, line 335: See above comment about exhaustiveness checking

herrstucki (approved):
Thanks! Looks much clearer to me now!

ptbrowne (approved):
Thanks, it is indeed much clearer now !

app/configurator/configurator-state.spec.tsx
```diff
@@ -126,11 +128,11 @@ describe("deriveFiltersFromField", () => {
     };
     const expectedFiltersState = {};
 
-    _applyDimensionToFilters(initialFiltersState, false, false);
+    _applyDimensionToFilters(initialFiltersState, true, false, false);
```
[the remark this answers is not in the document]
ptbrowne: Ah I forgot to post these comment yesterday. Nevermind...

app/configurator/configurator-state.spec.tsx
```diff
@@ -194,7 +196,25 @@ describe("deriveFiltersFromField", () => {
       },
     };
 
-    _applyDimensionToFilters(initialFiltersState, false, true);
+    _applyDimensionToFilters(initialFiltersState, true, false, true);
+    expect(initialFiltersState).toEqual(expectedFiltersState);
+  });
+
+  it("non-table: should not modify filters if dimension is not used as a field", () => {
```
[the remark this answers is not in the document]
ptbrowne: Also forgot to post it yesterday, this is absolutely not blocking.

app/configurator/configurator-state.spec.tsx
bprusinowski: Right, auto-imports 😅

app/configurator/configurator-state.tsx
bprusinowski: Yes - never used it previously, thanks!

PR #199, opened 2021-11-10
Browsing categories by organizations (browsability step 2)

ptbrowne:
This PR is fairly big.

- Ability to navigate organisations / themes through a dynamic tree
- Themes / organizations are displayed as tags inside the search results
- Can navigate to
  - /browse/organization/<iri>
  - /browse/organization/<iri>/theme/<iri>
  - /browse/theme/<iri>
  - /browse/theme/<iri>/organizations/<iri>
 - Tags inside cards can be used to navigate to organisations / themes
 - A autocomplete can be used to jump to theme / organisations on the index page
 - Full text search can be used on the browse page


app/rdf/query-cube-metadata.ts
```diff
@@ -42,11 +49,41 @@ export const createThemeLoader =
...
+export const createOrganizationLoader =
+  ({ locale }: { locale: string }) =>
+  async (
+    filterIris?: readonly string[]
+  ): Promise<(DataCubeOrganization | null)[]> => {
+    const query = SELECT.ALL.WHERE`
+    graph   <https://lindas.admin.ch/sfa/opendataswiss> {
+      ?theme a ${schema.Organization} ;
+        ${schema.name} ?name ;
+      filter (langmatches(lang(?name), "${locale}"))
```
herrstucki, line 64: Maybe not relevant to the prototype but have a look at the `loadResourceLabels` function from app/rdf/query-labels.ts here. What that does is it falls back to other languages if the main one is not available.

app/rdf/query-cube-metadata.ts
```diff
@@ -15,6 +15,13 @@ const parseSparqlTheme = (sparqlTheme: any) => {
   } as DataCubeTheme;
 };
 
+const parseSparqlOrganization = (sparqlOrg: any) => {
+  return {
+    label: sparqlOrg.name.value,
+    iri: sparqlOrg.theme.value,
+    __typename: "DataCubeOrganization",
```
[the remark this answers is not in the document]
ptbrowne: I've tried to fix the __typename problems in https://github.com/visualize-admin/visualization-tool/pull/199/commits/8999084a58f15184b39b8948de440d4d9e018d5b, my problem is that I reuse in the resolvers/parsers the types that are automatically created from the graphql schema, this is why I added the __typename there.
In the fixing commit, I Omit the __typename in the return type of resolvers / parsers. 🤞

app/rdf/query-cube-metadata.ts
ptbrowne: Ah yes, I had seen that and moved it to the end of the prototype, I will take a look thanks.

app/graphql/schema.graphql
```diff
@@ -25,6 +25,7 @@ type DataCube {
   version: String
   contactName: String
   contactEmail: String
+  creator: DataCubeOrganization
```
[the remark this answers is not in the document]
ptbrowne: If it makes sense, no problem for me to rename the field.

app/rdf/query-cube-metadata.ts
ptbrowne: Done in https://github.com/visualize-admin/visualization-tool/pull/199/commits/633f2b1c772eafee69dbdbf412cebcb08a97e095

app/graphql/schema.graphql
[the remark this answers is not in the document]
herrstucki: Not sure if it really makes sense but I guess someone decided it for a reason. Maybe a super-detail thing to consider is that IF there are multiple creators attached to a cube, there's no defined order to them, so if we decide to pick the first one, it may be a bit of a random choice

app/rdf/query-cube-metadata.ts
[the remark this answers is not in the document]
ptbrowne: Solved with using resolver types instead of query result types.

app/components/autocomplete.tsx
```diff
@@ -0,0 +1,155 @@
...
+    <Flex
+      as="li"
+      {...boxProps}
+      sx={{
+        alignItems: "center",
+        px: 4,
+        py: 4,
+      }}
+    >
+      {icon ? <Box sx={{ width: 24, height: 24 }}>{icon}</Box> : null}
```
bprusinowski, line 51: nit: icon && <Box...></Box>?

ptbrowne: Ah yes, I always do that, not sure why, I agree.

app/configurator/components/breadcrumbs.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const Breadcrumbs = ({
+  breadcrumbs,
+  Breadcrumb,
+  onClickBreadcrumb,
+  ...boxProps
+}: {
+  breadcrumbs: any;
+  Breadcrumb: React.ElementType<{ breadcrumb: any; onClick: () => void }>;
+  onClickBreadcrumb?: (i: number) => void;
```
herrstucki, line 14: nit: Does the `<Breadcrumbs>` component even need to know about how to handle clicks or is it better to leave that to the individual breadcrumbs (e.g. the `<Links>`) and use composition. So this could simply become `{ children: ReactNode } & BoxProps` …

```jsx
// Instead of
<Breadcrumbs
  breadcrumbs={["one", "two"]}
  Breadcrumb={MyComponent}
  onClickBreadcrumb={() => ...}
/>

// Rather this?

<Breadcrumbs>
  <Link href="...">One</Link>
  <Link href="...">Two</Link>
</Breadcrumb>
```

herrstucki (commented):
Overall LGTM! It really works nicely!

- The autocomplete on the home page doesn't *really* work yet, sometimes I select something and nothing happens
- When selecting a dataset to preview should we go to /create/new?cube=<iri> already? Otherwise we can't link to this view.

app/rdf/query-cube-metadata.ts
```diff
@@ -88,3 +87,63 @@ export const loadThemes = ({ locale }: { locale: string }) => {
...
+}: {
+  theme?: string;
+}) => {
+  const baseQuery = SELECT`(count(?iri) as ?count) ?creator`.WHERE`
+    ?iri ${dcterms.creator} ?creator.
+    ${theme ? `?iri <${dcat.theme.value}> <${theme}>.` : ``}
+    ?iri ${schema.workExample} <https://ld.admin.ch/application/visualize>.
+    ?iri ${
+      schema.creativeWorkStatus
+    } <https://ld.admin.ch/vocabulary/CreativeWorkStatus/Published>
```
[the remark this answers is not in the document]
ptbrowne: Yes. Keeping this in mind.

app/configurator/components/breadcrumbs.tsx
ptbrowne: Yes you're right I prefer the second way also.

app/components/autocomplete.tsx
ptbrowne: I will keep it this way as it is not so different and that I prefer to be able to quickly change the null to something :)

herrstucki:
Another thing to _maybe_ consider: I usually prefer to set up the routes more explicitly like `/browse/organizations/[orgIri]`, `/browse/organizations/[orgIri]/themes/[themeIri]` etc. rather than `/browse/[type]/[iri]/[subtype]/[subiri]`. It's a bit more redundant (needs more pages) _but_ has the advantage that you can name the dynamic params `orgIri` and `themeIri` regardless of how they're ordered, and that impossible types/subtypes (e.g. `/browse/dogs/.../cats/...` don't need to be handled explicitly. Of course, when more types get added it may need a more generic solution … but then maybe using query params would be preferable. Anyway, feel free to ignore this comment :)

PR #205, opened 2021-11-10
Rename env variables for clarity

herrstucki:
Follow-up on #194. Rename env variables to make it more clear which are available in the client, server and which are set during build time.


ptbrowne (approved):
Thanks !

PR #206, opened 2021-11-11
Add missing data imputations for stacked area charts

bprusinowski:
This PR is based on the conversation in #148 and it:
- introduces a concept of imputation of missing data (currently only applicable to stacked area charts with different time ranges for individual series),
- allows imputations with zeros or linearly interpolated values (in this case, if a series has a missing value at the beginning or at the end, then it will still be filled using 0),
- adds a proper section (called "Missing values") in the chart configurator (right panel) when using an area chart,
- the above section is disabled when no missing values are detected and is activated otherwise (I found it distracting when the whole section was disappearing and re-appearing but let me know what do you think about this).

Still TODO if you think the approach is ok:
- [x] add translations (only for published charts as of now)
- [x] add some info on the published chart to let people know about the fact that missing values were filled (?)
- [x] add tests

I will also add some comments to the code itself, as I'm not sure if I placed the some things in correct places :)

Thanks!

### DIFF
--- a/app/charts/area/areas-state.tsx
+++ b/app/charts/area/areas-state.tsx
@@ -2,6 +2,7 @@ import {
   ascending,
   descending,
   extent,
+  group,
   max,
   min,
   rollup,
@@ -32,11 +33,12 @@ import { useLocale } from "../../locales/use-locale";
 import { BRUSH_BOTTOM_SPACE } from "../shared/brush";
 import {
   getLabelWithUnit,
+  getWideData,
   useOptionalNumericVariable,
   usePreparedData,
   useSegment,
+  useStringVariable,
   useTemporalVariable,
-  useWideData,
 } from "../shared/chart-helpers";
 import { TooltipInfo } from "../shared/interaction/tooltip";
 import { ChartContext, ChartProps } from "../shared/use-chart-state";
@@ -95,10 +97,14 @@ const useAreasState = ({
 
   const getX = useTemporalVariable(fields.x.componentIri);
   const getY = useOptionalNumericVariable(fields.y.componentIri);
+  const getGroups = useStringVariable(fields.x.componentIri);
   const getSegment = useSegment(fields.segment?.componentIri);
 
   const hasSegment = fields.segment;
-  const allSegments = [...new Set(data.map((d) => getSegment(d)))];
+  const allSegments = useMemo(
+    () => [...new Set(data.map((d) => getSegment(d)))],
+    [data, getSegment]
+  );
 
   const xKey = fields.x.componentIri;
   const hasInteractiveTimeFilter = useMemo(
@@ -115,13 +121,21 @@ const useAreasState = ({
     [data, getX]
   );
 
-  const allDataWide = useWideData({
-    data: sortedData,
-    segments: allSegments,
-    getSegment,
-    getY,
-    xKey,
-  });
+  const sortedDataGroupedByX = useMemo(
+    () => group(data, getGroups),
+    [data, getGroups]
+  );
+
+  const allDataWide = useMemo(
+    () =>
+      getWideData({
+        dataGroupedByX: sortedDataGroupedByX,
+        xKey,
+        getY,
+        getSegment,
+      }),
+    [sortedDataGroupedByX, xKey, getY, getSegment]
+  );
 
   // Data for chart
   const preparedData = usePreparedData({
@@ -133,14 +147,30 @@ const useAreasState = ({
     getSegment,
   });
 
-  const chartWideData = useWideData({
-    data: preparedData,
-    xKey,
-    segments: allSegments,
-    imputationType: fields.y.imputationType,
-    getSegment,
-    getY,
-  });
+  const preparedDataGroupedByX = useMemo(
+    () => group(preparedData, getGroups),
+    [preparedData, getGroups]
+  );
+
+  const chartWideData = useMemo(
+    () =>
+      getWideData({
+        dataGroupedByX: preparedDataGroupedByX,
+        xKey,
+        getY,
+        allSegments,
+        getSegment,
+        imputationType: fields.y.imputationType,
+      }),
+    [
+      preparedDataGroupedByX,
+      xKey,
+      getY,
+      allSegments,
+      getSegment,
+      fields.y.imputationType,
+    ]
+  );
 
   const yMeasure = measures.find((d) => d.iri === fields.y.componentIri);
 
--- a/app/charts/area/areas-state.tsx
+++ b/app/charts/area/areas-state.tsx
@@ -20,7 +20,7 @@ import {
   sum,
 } from "d3";
 import { ReactNode, useCallback, useMemo } from "react";
-import { AreaFields, ImputationType } from "../../configurator";
+import { AreaFields } from "../../configurator";
 import {
   getPalette,
   parseDate,
@@ -33,9 +33,9 @@ import { estimateTextWidth } from "../../lib/estimate-text-width";
 import { useLocale } from "../../locales/use-locale";
 import { BRUSH_BOTTOM_SPACE } from "../shared/brush";
 import {
-  getBaseWideData,
   getLabelWithUnit,
   getTemporalWideDataWithImputedValues,
+  getWideData,
   usePreparedData,
 } from "../shared/chart-helpers";
 import { TooltipInfo } from "../shared/interaction/tooltip";
@@ -69,14 +69,12 @@ const useAreasState = ({
   fields,
   dimensions,
   measures,
-  imputationType,
   interactiveFiltersConfig,
   aspectRatio,
 }: Pick<
   ChartProps,
   "data" | "dimensions" | "measures" | "interactiveFiltersConfig"
 > & {
-  imputationType: ImputationType;
   fields: AreaFields;
   aspectRatio: number;
 }): AreasState => {
@@ -131,7 +129,7 @@ const useAreasState = ({
     sortedData,
     (d) => d[fields.x.componentIri] as string
   );
-  const allDataWide = getBaseWideData({
+  const allDataWide = getWideData({
     groupedMap: allDataGroupedMap,
     getSegment,
     getY,
@@ -157,7 +155,7 @@ const useAreasState = ({
     groupedMap,
     xKey,
     segments: allSegments,
-    imputationType,
+    imputationType: fields.y.imputationType || "none",
     segmentKey,
     getSegment,
     getY,
@@ -374,14 +372,12 @@ const AreaChartProvider = ({
   measures,
   dimensions,
   interactiveFiltersConfig,
-  imputationType,
   aspectRatio,
   children,
 }: Pick<
   ChartProps,
   "data" | "fields" | "dimensions" | "measures" | "interactiveFiltersConfig"
 > & {
-  imputationType: ImputationType;
   children: ReactNode;
   aspectRatio: number;
 } & { fields: AreaFields }) => {
@@ -390,7 +386,6 @@ const AreaChartProvider = ({
     fields,
     dimensions,
     measures,
-    imputationType,
     interactiveFiltersConfig,
     aspectRatio,
   });
@@ -405,14 +400,12 @@ export const AreaChart = ({
   measures,
   dimensions,
   interactiveFiltersConfig,
-  imputationType,
   aspectRatio,
   children,
 }: Pick<
   ChartProps,
   "data" | "fields" | "dimensions" | "measures" | "interactiveFiltersConfig"
 > & {
-  imputationType: ImputationType;
   children: ReactNode;
   fields: AreaFields;
   aspectRatio: number;
@@ -425,7 +418,6 @@ export const AreaChart = ({
           fields={fields}
           dimensions={dimensions}
           measures={measures}
-          imputationType={imputationType}
           interactiveFiltersConfig={interactiveFiltersConfig}
           aspectRatio={aspectRatio}
         >
--- a/app/charts/area/areas-state.tsx
+++ b/app/charts/area/areas-state.tsx
@@ -13,7 +13,6 @@ import {
   ScaleTime,
   scaleTime,
   stack,
-  stackOffsetDiverging,
   stackOrderAscending,
   stackOrderDescending,
   stackOrderReverse,
@@ -34,6 +33,7 @@ import { BRUSH_BOTTOM_SPACE } from "../shared/brush";
 import {
   getLabelWithUnit,
   getWideData,
+  stackOffsetDivergingPositiveZeros,
   useOptionalNumericVariable,
   usePreparedData,
   useSegment,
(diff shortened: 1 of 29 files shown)


app/configurator/config-types.ts
```diff
@@ -273,13 +281,16 @@ const AreaFields = t.intersection([
...
+const AreaConfig = t.intersection(
+  [
+    t.type({
+      chartType: t.literal("area"),
+      filters: Filters,
+      interactiveFiltersConfig: InteractiveFiltersConfig,
+      fields: AreaFields,
+    }),
+    t.partial({ imputationType: ImputationType }),
+  ],
```
bprusinowski, line 293: Not sure is imputationType should be at this level; maybe it belongs to Y field? Or a segment field, as currently it's only used in conjunction with segments?

app/charts/shared/chart-helpers.tsx
```diff
@@ -121,31 +127,166 @@ export const usePreparedData = ({
...
+export const getTemporalWideDataWithImputedValues = ({
+  xKey,
+  groupedMap,
+  imputationType,
+  segmentKey,
+  segments,
+  getSegment,
+  getX,
+  getY,
+}: {
```
bprusinowski, line 169: From my very simple performance tests, it seems that drawing a chart w/o any imputations is as fast as filling with 0s, while linear interpolation is several times slower. Maybe some caching could be used there to not run this function again and again when we use the same inputs? I' would appreciate your view on that :)

app/configurator/components/chart-options-selector.tsx
```diff
@@ -42,8 +47,27 @@ export const ChartOptionsSelector = ({
...
+      locale,
+      iri: state.dataSet,
+      measures,
+      filters: state.chartConfig.filters,
+    },
+  });
+
+  const imputationNeeded = useImputationNeeded({
+    chartConfig: state.chartConfig,
+    data: data?.dataCubeByIri?.observations.data,
```
bprusinowski, line 70: Also not sure if this should be done there, but I didn't come up with more elegant solution :)

app/components/chart-preview.tsx
```diff
@@ -159,49 +167,49 @@ const Chart = ({
   return (
     <>
       {/* CHARTS */}
-      {chartConfig.chartType === "column" && (
+      {isColumnConfig(chartConfig) && (
```
herrstucki, line 170: Nice!

app/charts/shared/chart-helpers.tsx
```diff
@@ -160,4 +168,43 @@ export const getLabelWithUnit = (
...
+  for (const value of groupedMap.values()) {
+    if (value.length !== segments.length) {
+      return true;
+    }
+  }
+
+  return false;
+};
 
+export const useImputationNeeded = ({
```
herrstucki, line 317: Do I understand correctly … this is only true IF we use segments? No imputation needed when the area chart is not stacked?

app/charts/shared/chart-helpers.tsx
herrstucki: Can we find the actual slow parts (using Chrome dev tools)? I haven't looked at the code too closely yet but it seems that there are quite a few intermediate arrays created and looped over and maybe there's an opportunity to optimize this a bit.

Just by looking at the arguments of this function, memoizing seems not super-trivial, since we need to make sure none of these inputs change between renders. Not the easiest to do with React …

app/configurator/config-types.ts
herrstucki: Yeah … sounds more like a field-level config (could there be a case where x AND y are imputed?)? I don't think it just belongs to the segment part though.

app/charts/shared/chart-helpers.tsx
```diff
@@ -160,4 +168,43 @@ export const getLabelWithUnit = (
     : dimension.label;
 };
 
+export const checkForMissingValuesInSegments = (
+  groupedMap: InternMap<string, Array<Observation>>,
```
ptbrowne, line 305: `observationsByX` would be clearer to me, I had trouble understanding by what the observations were grouped here.

herrstucki:
I think the thing we really need to decide on a design level is which messages to show and when, especially on a published chart. I'm generally not a fan of adding disclaimers all over the place :)

app/charts/shared/chart-helpers.tsx
bprusinowski: Yes, right now this only applies to stacked area charts – in case of a regular area chart, there won't be any missing segments which would require imputation

app/charts/shared/chart-helpers.tsx
```diff
@@ -160,4 +168,43 @@ export const getLabelWithUnit = (
...
+
+  if (isAreaConfig(chartConfig) && data) {
+    return checkForMissingValuesInSegments(
+      group(
+        data,
+        (d: Observation) => d[chartConfig.fields.x.componentIri] as string
+      ),
+      [...new Set(data.map((d) => getSegment(d)))]
+    );
+  }
```
ptbrowne, line 340: I think you should useMemo here so that we do not checkForMIssingValues everytime we pass into this function. I think the getSegment helper could be intiialized inside the useMemo, not sure it it serves that it is useful on its own.

app/configurator/components/chart-options-selector.tsx
```diff
@@ -42,8 +46,27 @@ export const ChartOptionsSelector = ({
...
-  const [{ data }] = useDataCubeMetadataWithComponentValuesQuery({
-    variables: { iri: state.dataSet, locale },
+  const measures =
+    "y" in state.chartConfig.fields
+      ? [state.chartConfig.fields.y.componentIri]
+      : isTableConfig(state.chartConfig)
+      ? Object.values(state.chartConfig.fields).flatMap((f) =>
+          f.componentType === "Measure" && !f.isHidden ? [f.componentIri] : []
+        )
+      : [];
```
ptbrowne, line 57: Nit: useMemo with chartConfig as dep would not hurt.

app/configurator/components/chart-options-selector.tsx
ptbrowne: I think it's OK for now :)

app/configurator/components/chart-options-selector.tsx
```diff
@@ -120,13 +150,15 @@ const EncodingOptionsPanel = ({
...
   metaData,
+  imputationNeeded,
 }: {
   encoding: EncodingSpec;
   state: ConfiguratorStateConfiguringChart;
   field: string;
   chartType: ChartType;
   component: DimensionMetaDataFragment | undefined;
   metaData: DataCubeMetadata;
+  imputationNeeded: boolean;
```
ptbrowne, line 162: Not a fan of this though (prop drilling the imputationNeeded), it seems like it should be part of a more generic context that would contain computed informations on the chart. The generic "stats" computation context would hold imputationNeeded, and then component that need it would just take it from the context.

Let's keep it in mind when we do further improvements, I do not think it is needed at this point.

app/charts/shared/chart-helpers.tsx
```diff
@@ -129,27 +126,36 @@ export const usePreparedData = ({
...
+// Helper to pivot a dataset to a wider format
+export const getBaseWideData = ({
   xKey,
   groupedMap,
   getSegment,
   getY,
 }: {
   xKey: string;
-  groupedMap: Map<string, Record<string, ObservationValue>[]>;
+  groupedMap: Map<string, Array<Observation>>;
```
ptbrowne, line 138: Again I think that `groupedBy<GroupVariable>` would be better.

app/charts/area/areas-state.tsx
```diff
@@ -106,11 +105,13 @@ const useAreasState = ({
     const v = d[fields.y.componentIri];
     return v !== null ? +v : null;
   };
+  const segmentKey = fields.segment ? fields.segment.componentIri : "segment";
   const getSegment = useCallback(
     (d: Observation): string =>
       fields.segment ? (d[fields.segment.componentIri] as string) : "segment",
     [fields.segment]
   );
+  const allSegments = [...new Set(data.map((d) => getSegment(d)))];
```
ptbrowne, line 114: Nit: For the record, I had trouble understand "segment", maybe stackVar would be better ? Not for this PR though.

app/configurator/config-types.ts
bprusinowski: I think that generally, as of now, there isn't a case where x and y are both inputed. We are only imputing missing values that were "artificially" introduced to the data by selecting a particular chart type (stacked area chart) and not all the missing values. So let's say that we have an X temporal axis with a "yearly" resolution; if we are missing some years, we are not offering a way to impute these; only if a missing value appeared due to stacking of series of different timespans.

app/charts/shared/chart-helpers.tsx
bprusinowski: Will look into that!

ptbrowne: It seems that we would only need

- data
- x variable
- y variable
- stack variable ("segment")
- imputation type

and then we can memoize easily. getX and xKey seem redundant and groupedMap seems to be computable through data and segmentKey right ?

I think that memoization should be done regardless of whether the underlying algorithm is performant.

I find the algorithm below hard to read, I am wondering if it could be simplified into:

```
const fillNaLinear = (values) => {
  
}

const fillNaZeros = (values) =. {}

const fillNaNone = (values) => {}
```

Here it seems like the core algorithm is obscured by our particular business logic on how to access x value, y value, segment value.

app/configurator/config-types.ts
ptbrowne: To me, it would make sense to put it in the "y" field config. I do not have a super strong opinion though.

ptbrowne:
I think reworking the particular `getTemporalWideDataWithImputedValues` needs to be simplified a bit. The API should be simpler and memoized. In general though I think the PR is going in the right direction.

bprusinowski:
Thanks for your comments @herrstucki and @ptbrowne! I have introduced some changes that made the linear imputation as fast as filling with zeros; also I hope the code will be more clear now.

app/charts/shared/imputation.tsx
```diff
@@ -0,0 +1,82 @@
...
+  dataSortedByX: Array<TemporalSeriesBeforeImputationEntry>;
+}): Array<TemporalSeriesAfterImputationEntry> => {
+  const indexedData: Array<[TemporalSeriesBeforeImputationEntry, number]> =
+    dataSortedByX.map((d, i) => [d, i]);
+  const presentDataIndexes = indexedData
+    .filter((d) => d[0].value !== null)
+    .map((d) => d[1]);
+  const missingDataIndexes = indexedData
+    .filter((d) => d[0].value === null)
+    .map((d) => d[1]);
```
herrstucki, line 48: Mostly a nit but this still creates some unnecessary Arrays that you don't really need (and have to be garbage-collected).

```ts
const indexedData: Array<[TemporalSeriesBeforeImputationEntry, number]> =
    dataSortedByX.map((d, i) => [d, i]); // -> new Array!
  const presentDataIndexes = indexedData
    .filter((d) => d[0].value !== null) // -> new Array!
    .map((d) => d[1]); // -> new Array!
  const missingDataIndexes = indexedData
    .filter((d) => d[0].value === null) // -> new Array!
    .map((d) => d[1]); // -> new Array!
```

You could combine `.filter().map()` into one loop with `dataSortedByX.flatMap((d, i) => d.value !== null ? [i] : [])` but if you wanted to optimize this a bit more, a good old-fashioned for-loop is the way to go (less "functional" but arguably even more readable?). This way you only need one loop:

```ts
let presentDataIndexes = [];
let missingDataIndexes = [];
for (let i = 0; i < dataSortedByX.length; i++) {
  if (dataSortedByX[i].value !== null) {
    presentDataIndexes.push(i)
  } else {
    missingDataIndexes.push(i)
  }
}
```

app/charts/shared/imputation.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  for (const missingDataIndex of missingDataIndexes) {
+    const nextIndex = presentDataIndexes.findIndex((d) => d > missingDataIndex);
```
herrstucki, line 51: Nit: The name `nextIndex` is bit confusing because this refers to the index of the array containing indexes.

herrstucki (commented):
Nice, I like how you extracted some of the shared logic between charts. To me it's much clearer than before, I only commented on some details in the code.

herrstucki:
The build on Heroku is currently failing … I'll try to check out locally and find out what the issue is

herrstucki:
There still seems to be an issue with the linear interpolation, maybe specific to this value being 0

Bildschirmfoto 2021-11-16 um 10 35 11

app/charts/shared/imputation.tsx
```diff
@@ -0,0 +1,82 @@
+import { interpolate } from "d3-interpolate";
+
+export const interpolateZerosValue = () => {
+  return Number.MIN_VALUE;
```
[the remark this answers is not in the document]
bprusinowski: This is for the same reason why the interpolation seems broken for 0 (the image you provided below)... We are using a [d3.stackOffsetDiverging](https://github.com/d3/d3-shape#stackOffsetDiverging) for specify how the values should be stacked (_positive values are stacked above zero, negative values are stacked below zero, and zero values are stacked at zero_). By using this, 0s are always stacked at zeros (I also encountered problems like this and that's why I opted for the Number.MIN_VALUE when filling with 0s).

I guess we still need this offset type, as there can be negative values in series, but maybe we also could replace 0s with Number.MIN_VALUE in the actual data?

app/charts/shared/imputation.tsx
ptbrowne: 👍  for jeremy approach if this is a tight loop.

Another approach with still intermediate arrays (for the record): 

```
const { true: presentDataIndexes, false: missingDataIndexes } = mapValues(groupBy(indexedData, d => d[0].value === null), x => x[1])
```

herrstucki (approved):
Approved! Some of the German could be improved slightly but let's do that another time.

herrstucki:
What I mean specifically is that "Imputation" is a very technical term but I don't have a better proposal at this point 😉

app/charts/shared/imputation.spec.tsx
```diff
@@ -0,0 +1,56 @@
...
+      dataSortedByX: dataToImpute,
+    });
+
+    expect(Math.round(dataToImpute[1].value! * 1e5) / 1e5).toEqual(
+      Math.round(expectedLinearlyInterpolatedValue * 1e5) / 1e5
+    );
+    expect(dataToImpute[0].value).toEqual(previousValue);
+    expect(dataToImpute[2].value).toEqual(nextValue);
+  });
+});
```
ptbrowne, line 56: 🎉

bprusinowski:
Thanks! Yeah, @ptbrowne also mentioned that – maybe we could go with "filling"? But let's decide that after client's feedback :)

ptbrowne:
👍 🎉  Thanks good job for the tests :)

PR #216, opened 2021-11-22
Dataset browsing feedback

### LINKED ISSUE #212: Number of existing datasets per theme
On the Swiss Open Government Data page (https://test.visualize.admin.ch/en/browse)
the number of datasets indicated per theme doesn’t correspond to the number of datasets in the result list.

image

ptbrowne:
- fix: Can go back to dataset results after creating a dataset
- fix: Translate create dataset button
- fix: Filter out datasets with validThrough in dataset count

Fix #212


app/rdf/query-cube-metadata.ts
```diff
@@ -115,23 +114,23 @@ const makeVisualizeDatasetFilter = () => {
...
+    `;
 };
 
 export const queryDatasetCountByTheme = async ({
   organization,
 }: {
   organization?: string;
 }) => {
-  const baseQuery = SELECT`(count(?iri) as ?count) ?theme`.WHERE`
+  const query = SELECT`(count(?iri) as ?count) ?theme`.WHERE`
```
herrstucki, line 126: Nit: you can also interpolate variables from `rdf-ext` instead of writing `?varname` by hand. Cf. 

https://github.com/visualize-admin/electricity-prices-switzerland/blob/84793e87f8f4d78814a08d0a5ed74e562fc91a93/src/rdf/search-queries.ts#L14-L19

ptbrowne: I can't clearly see the benefit in the queries I see in the elcom project ? It seems you have to write "?varName as ${vars.varName}" = repetition of varName ? I'd be happy to just do "SELECT ${vars.varName}", is is possible ? Or is there a benefit I do not see ?

herrstucki: Yes, that would be the (arguably very minor) point. It would be sth like

```js
const iri = rdf.variable("iri");
const count = rdf.variable("count");
const theme = rdf.variable("theme");

const query = SELECT`(count(${iri}) as ${count}) ${theme}`.WHERE`...`; 
```

I don't think it's really necessary here but may prevent bugs in larger queries because you get support from TS when writing them :)

bprusinowski (approved):

app/rdf/query-cube-metadata.ts
ptbrowne: Thanks for the advice, I will do that when I go again on this file :)

Issue #211, opened 2021-11-19
No "data Load" by Chart Types Lines and Areas as a first step


ortnever:
The data cannot be loaded when the "Lines" or "Areas" chart type is selected. 
To solve the problem, you have to reselect the Bar type or an other Chart Type, then go back to Lines or Areas a second time to display the chart.
Try with “Bathing Water Quality” but the bug cannot always be reproduced.
![image](https://user-images.githubusercontent.com/79834090/142630309-219fdf6e-f0c7-4b1e-bfeb-8bdfd5c4ee57.png)

PR #221, opened 2021-11-24
Fix chart lines loading

ptbrowne:
When creating a new dataset and selecting lines chart, the chart would stay blocked in loading mode. This was due to the `fetching` attribute of the OperationResult staying to true when a query was refetched while it was already in cache. Updating urql fixes the problem 🎉 

Fixes: https://github.com/visualize-admin/visualization-tool/issues/211

Two minor commits that I was doing at the time of the fix:

- docs: Rename dataset search to dataset browse
- refactor: Extract and use set instead of array to check for inclusivity


app/configurator/components/dataset-browse.tsx
```diff
@@ -1,28 +1,30 @@
 import { Maybe } from "@graphql-tools/utils/types";
 import { Plural, t, Trans } from "@lingui/macro";
 import { mapValues, pickBy, sortBy } from "lodash";
+import Link from "next/link";
+import { Router, useRouter } from "next/router";
+import qs from "qs";
```
herrstucki, line 6: nit: These imports got added but I don't see them used?

herrstucki (approved):

herrstucki:
Ah, I don't like automerge 😂 @ptbrowne can you make sure to check the imports on main?

ptbrowne:
They are imported, and we now have an eslint rule for unused imports ;)

herrstucki:
My point was, they should not be imported? Maybe I saw wrong 😅

ptbrowne:
Oups sorry, I meant "they are used" (I checked). The eslint rule would error if they were not used ;)

herrstucki:
Ah nevermind … just a quick note: you don't need to use qs.stringify, as you can just pass the query as object to `router.push()`. Just highlighting this because this would add another package to the bundle that (hopefully) is not used elsewhere and duplicates functionality built into next.js.

PR #227, opened 2021-11-25
Add new dimension types

jstcki:
Adds support for ordinal and geographical dimensions

- Consolidates Nominal/Ordinal Dimensions into CategoricalDimension with `hasOrder` property
- Adds GeoShape and GeoPoint Dimensions
- Adds `position` to dimension values (optional, only used when dimension is marked as ordinal)
- Moves properties to specific dimension types (like unit)

To do:
- [ ] Fix type errors
- [ ] Implement resolvers


herrstucki:
Discussed with @ptbrowne that we're not aiming to merge this PR but treat it as an exploration to be reviewed and maybe be used for reference in a future refactor of the GraphQL schema. Support for ordinal dimension will follow a more simple approach that doesn't impact the code base as extensively.

PR #232, opened 2021-11-26
Hierarchical filters step 1

ptbrowne:
This PR prepares the future work on hierarchical filters. It does not add any new functionality (yay!?).

Since in the future we want to be able to display hierarchical multi checkboxes, and we will possibly separate the color pickers from the data filters, this PR ensures that the UI components can be more freely re-arranged. 

- Checkboxes and color pickers are now separate components
- They know what to dispatch and do not rely on their parents to pass the correct properties
- Since we need to have a shared state (containing what is checked, the selection state and several other values), I used a context to share the information
- I used hooks that contain the budget logic to separate the presentation from the logic. This could be useful in the future when we want to have the color picker directly on the chart. This followed the model that was already used for some fields in config-form.tsx

### DIFF
--- a/app/configurator/components/field.tsx
+++ b/app/configurator/components/field.tsx
@@ -20,6 +20,7 @@ import { IconName } from "../../icons";
 import {
   useChartOptionBooleanField,
   useChartOptionSelectField,
+  useMultiFilterCheckboxes,
   useSingleFilterSelect,
 } from "../config-form";
 import { FIELD_VALUE_NONE } from "../constants";
@@ -386,7 +387,7 @@ export const MultiFilterFieldCheckbox = ({
   disabled,
   allValues,
   checked,
-  onChange,
+  onChange: onChangeProp,
   checkAction,
   colorConfigPath,
 }: {
@@ -400,36 +401,13 @@ export const MultiFilterFieldCheckbox = ({
   checkAction: "ADD" | "SET";
   colorConfigPath?: string;
 }) => {
-  const [state, dispatch] = useConfiguratorState();
-
-  const onFieldChange = useCallback<(e: ChangeEvent<HTMLInputElement>) => void>(
-    (e) => {
-      if (e.currentTarget.checked) {
-        dispatch({
-          type:
-            checkAction === "ADD"
-              ? "CHART_CONFIG_FILTER_ADD_MULTI"
-              : "CHART_CONFIG_FILTER_SET_MULTI",
-          value: {
-            dimensionIri,
-            value,
-            allValues,
-          },
-        });
-      } else {
-        dispatch({
-          type: "CHART_CONFIG_FILTER_REMOVE_MULTI",
-          value: {
-            dimensionIri,
-            value,
-            allValues,
-          },
-        });
-      }
-      // Call onChange prop
-      onChange?.();
-    },
-    [dispatch, dimensionIri, allValues, value, onChange, checkAction]
+  const [state] = useConfiguratorState();
+  const { onChange: onFieldChange } = useMultiFilterCheckboxes(
+    dimensionIri,
+    value,
+    allValues,
+    checkAction,
+    onChangeProp
   );
 
   if (state.state !== "CONFIGURING_CHART") {
@@ -453,6 +431,7 @@ export const MultiFilterFieldCheckbox = ({
     />
   );
 };
+
 export const SingleFilterField = ({
   dimensionIri,
   label,
--- a/app/configurator/components/field.tsx
+++ b/app/configurator/components/field.tsx
@@ -21,7 +21,6 @@ import {
   useChartOptionBooleanField,
   useChartOptionSelectField,
   useMultiFilterCheckboxes,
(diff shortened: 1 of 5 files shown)


app/configurator/components/field.tsx
```diff
@@ -293,122 +296,151 @@ export const MetaInputField = ({
...
+  dimensionIri: string;
+  checked?: boolean;
+}) => {
+  const { color, checked, palette, onChange } = useMultiFilterColorPicker(
+    value,
+    dimensionIri,
+    colorConfigPath
+  );
+
+  return color && checked ? (
```
bprusinowski, line 399: Nit: color && checked && (...)?

app/configurator/components/filters.tsx
```diff
@@ -97,68 +77,41 @@ export const DimensionValuesMultiFilter = ({
...
-                colorConfigPath={colorConfigPath}
-              />
+                sx={{
+                  justifyContent: "space-between",
+                  alignItems: "center",
+                  mb: 1,
+                  height: "2rem",
+                }}
+              >
+                <Box sx={{ maxWidth: "82%" }}>
```
bprusinowski, line 96: Was there a specific reason to use 82%? :D

bprusinowski (approved):

app/configurator/components/filters.tsx
ptbrowne: I'd say "no", but this is not something I change, I mostly rearranged code here ;)

app/configurator/components/field.tsx
ptbrowne: Yes it can be done but I do not think it changes anything though :) I think I prefer the ternary form because it's easier to add the falsy case.

app/configurator/components/filters.tsx
ptbrowne: I will see what can be done about this in a future PR. @herrstucki if you remember the intention ?

herrstucki: Just checked, the max-width is limited to make space for the color picker. Not sure why 82% was used though 😁

app/configurator/config-form.tsx
```diff
@@ -364,14 +366,102 @@ export const useSingleFilterField = ({
...
+  }, [dimensionData, activeFilter]);
+
+  const selectionState: MultiFilterSelectionState = !activeFilter
+    ? "ALL_SELECTED"
+    : isFilterActive.size === 0
+    ? "NONE_SELECTED"
+    : "SOME_SELECTED";
+
+  const checkAction =
+    selectionState === "NONE_SELECTED" ? "SET" : ("ADD" as CheckActionType);
```
herrstucki, line 433: nit: instead of using `"ADD" as CheckActionType`, use `const checkAction: CheckActionType = ...` (if the annotation is even needed?)

app/configurator/config-form.tsx
```diff
@@ -402,7 +492,20 @@ export const useMultiFilterCheckboxes = (
...
+      : false;
+
+  return {
+    onChange,
+    checked:
+      selectionState === "ALL_SELECTED"
+        ? true
+        : selectionState === "SOME_SELECTED"
+        ? !!isFilterActive.has(value)
+        : undefined && checkedState,
```
herrstucki, line 502: This will always be `false`?

herrstucki (commented):
Hard to see from the code alone if everything still works as intended but will check. Code mostly LGTM though!

herrstucki:
The "deselect all" action turns all checkboxes into a fake disabled state (they're still checkable). They should NOT be disabled but unchecked instead.

Current PR:

image

Before:

image

app/configurator/components/filters.tsx
ptbrowne: In the next PR, I changed the code to using flex, with a flex grow item for the checkbox part and a flex shrink for the color picker.

ptbrowne:
Through [`a6d1866` (#232)](https://github.com/visualize-admin/visualization-tool/pull/232/commits/a6d186698eeec41aca5b43f5f57c4e2dd5044146),  the "select none" bug is solved , see https://visualize-ad-refactor-f-w0s42w.herokuapp.com/en/create/i7dh_FqQgB8A I

app/configurator/config-form.tsx
```diff
@@ -487,19 +469,14 @@ export const useMultiFilterCheckboxes = (
     [dispatch, dimensionIri, allValues, value, onChangeProp]
   );
 
-  const checkedState =
+  const isChecked =
     state.state === "CONFIGURING_CHART"
       ? isMultiFilterFieldChecked(state.chartConfig, dimensionIri, value)
       : false;
```
herrstucki, line 475: nit: I think this would be more clear:

```ts
const isChecked = state.state === "CONFIGURING_CHART" && isMultiFilterFieldChecked(state.chartConfig, dimensionIri, value);
```

herrstucki (approved):
Works great! 👍

PR #233, opened 2021-11-26
Hierarchical filters

ptbrowne:
DimensionValuesMultiFilter is now displayed as a tree
The hierarchy is for now hardcoded for the bathing site dataset

Various refactors were necessary on widgets like the selectbox to be able to cleanly insert this feature.

See the result here : https://visualize-ad-feat-hiera-6sn1sd.herokuapp.com/en/create/zU4PAOQ38gfL


app/configurator/configurator-state.tsx
```diff
@@ -770,48 +766,47 @@ const reducer: Reducer<ConfiguratorState, ConfiguratorStateAction> = (
...
-      }
-      return draft;
-
     case "CHART_CONFIG_FILTER_ADD_MULTI":
       if (draft.state === "CONFIGURING_CHART") {
-        const { dimensionIri, value, allValues } = action.value;
+        const { dimensionIri, values, allValues } = action.value;
         const f = draft.chartConfig.filters[dimensionIri];
+        const valuesUpdate = Object.fromEntries(
+          values.map((v: string) => [v, true as true])
```
bprusinowski, line 774: Is this `true as true` done on purpose?

ptbrowne: I am going to check but I think yes since otherwise it is considered as boolean whereas here there can't be `false`.

app/charts/shared/chart-helpers.tsx
```diff
@@ -87,48 +90,31 @@ export const usePreparedData = ({
...
+      getX && from && to && timeFilterActive
+        ? (d: any) =>
+            getX(d).getTime() >= from.getTime() &&
+            getX(d).getTime() <= to.getTime()
+        : null;
+    const legendFilter: ValuePredicate | null =
+      legendFilterActive && getSegment
+        ? (d: any) => !activeInteractiveFilters.includes(getSegment(d))
+        : null;
+    return flowRight.apply(null, [legendFilter, timeFilter].filter(truthy));
```
bprusinowski, line 104: This doesn't really matter, but I think you could do `(d: Observation)` here?

app/charts/column/columns-stacked-state.tsx
```diff
@@ -411,7 +411,7 @@ const useColumnsStackedState = ({
           color: colors(getSegment(datum)) as string,
         },
         values: sortedTooltipValues.map((td) => ({
-          label: getSegment(td),
+          label: `${getSegment(td)}`,
```
bprusinowski, line 414: `getSegment` is supposed to only return strings, so I think there is no need to change that?

bprusinowski (approved):
LGTM! The code is very clear and I like that you improved some of the code related to charts along the way :) 👍

app/configurator/configurator-state.tsx
ptbrowne: This is because of the type of the filters in chartConfig:

```
{
        type: "multi";
        values: {
            [x: string]: true;
        };
    }
```

app/charts/column/columns-stacked-state.tsx
ptbrowne: Yes ! Thanks

app/charts/shared/chart-helpers.tsx
ptbrowne: Yes, done, thanks !

PR #234, opened 2021-11-29
Improve translations on dataset selection step

herrstucki:
The wording on the dataset page was inconsistent, not self-descriptive enough or just wrong (at least in German).

Changes:
- "back to list" (was "back to results" in some languages?) to "back to datasets" (more descriptive and consistent)
- "create visualization from dataset" to "create visualization" (was inconsistent between languages and wrong in German)


herrstucki:
Checked with Tania that the 🇮🇹 changes are correct.

ptbrowne (approved):
Thank you !

PR #240, opened 2021-11-30
Maps integration

bprusinowski:
Draft of integrating maps to the app. You can test the feature by using a "Covid 19 Vaccinated Persons" dataset (and pass e.g. 2021-11-20 to the date filter to actually see some data :) ).

Currently all the shapes are fetched dynamically (so the static files are not used).

TODO:
- [x] handle GeoCoordinates dimensions
- [x] set GEO_SPARQL_ENDPOINT in .env file
- [x] separate tooltip for area layer & symbol layer when selected geo dimensions are different
- [ ] add translations
- [ ] centroids for weird shapes (symbolLayer based on geoShape)
- [ ] handle aggregated rows - partially done, waiting for Zazuko to decide on the final approach
- [ ] refactor geoShapes fetching so they are fetched once (not per every locale, as the shapes are the same)
- [ ] fix jenks color legend bug when number of classes is bigger than number of observations\
- [ ] index by label - add DataProvider
- [ ] move hierarchy creation to dimension-hierarchy + add test

TBD:
- [ ] hide shapes with no data linked to them
- [ ] add more information to a tooltip / allow switching measures in interactive filters
- [ ] match observations to shapes by IRI, not label
- [ ] hierarchies for geoCoordinates

Nice to have:
- [x] highlight areas or symbols on hover
- [x] change the select element for choosing the color palette to a colorPicker
- [x] optimize map legend logic
- [ ] polish the design of the chart options selector
- [ ] add an ability to set mesh color
- [ ] initial zoom to bounding box of visible shapes


app/charts/map/map-state.tsx
```diff
@@ -145,21 +144,13 @@ const useMapState = ({
...
-    measures
-      .find((m) => m.iri === fields["areaLayer"].measureIri)
-      ?.label.split("_")[1] || "";
+    measures.find((m) => m.iri === fields["areaLayer"].measureIri)?.label || "";
   const symbolMeasureLabel =
-    measures
-      .find((m) => m.iri === fields["symbolLayer"].componentIri)
-      ?.label.split("_")[1] || "";
+    measures.find((m) => m.iri === fields["symbolLayer"].componentIri)?.label ||
+    "";
```
ptbrowne, line 154: It looks like those calls could be either memoized or used an index, I am always a bit worried when I see a list scan in a render call.

app/charts/map/map.tsx
```diff
@@ -255,12 +237,40 @@ export const MapComponent = () => {
...
+              if (object) {
+                dispatch({
+                  type: "INTERACTION_UPDATE",
+                  value: {
+                    interaction: {
+                      visible: true,
+                      mouse: { x, y },
+                      d: data.find(
+                        (d) => getLabel(d) === object.properties.label
+                      ),
```
ptbrowne, line 264: Mmh, again an index seems like it could be good, I am worried for performance here since we need to find the object in a list with each movement of the mouse. Since we have the object being hovered  in the callback, couldn't we "save" the datum inside this object when rendering so that we can just do `interaction: { object.d }` ?

app/charts/map/map-legend.tsx
```diff
@@ -128,7 +129,7 @@ const CircleLegend = () => {
 
           return (
             <>
-              {d !== 0 && (
+              {d !== 0 && ( // FIXME: 0 also should be shown
```
ptbrowne, line 132: Do not forget this :)

app/charts/map/map.tsx
```diff
@@ -231,7 +231,7 @@ export const MapComponent = () => {
               getLineColor={[255, 255, 255]}
             />
           </>
-        )}
+        )} */}
```
ptbrowne, line 234: Should this be removed ?

app/charts/map/map.tsx
```diff
@@ -275,7 +275,7 @@ export const MapComponent = () => {
             updateTriggers={{ getFillColor: getColor }}
             getFillColor={(d: GeoShapeFeature) => {
               const entry = data.find(
-                (o) => getLabel(o) === d.properties.label
+                (o) => getLabel(o) === d.properties.label // FIXME?
```
ptbrowne, line 278: Do not forget the FIXME :)

app/charts/map/map-state.tsx
```diff
@@ -280,7 +293,7 @@ export const MapChart = ({
           fields={fields}
           dimensions={dimensions}
           measures={measures}
-          // interactiveFiltersConfig={interactiveFiltersConfig}
+          interactiveFiltersConfig={interactiveFiltersConfig}
```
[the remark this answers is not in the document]
bprusinowski: Yes, that's another thing to change! Thanks for catching :)

app/configurator/configurator-state.tsx
```diff
@@ -547,7 +547,10 @@ const reducer: Reducer<ConfiguratorState, ConfiguratorStateAction> = (
         ];
         if (!f) {
           // The field was not defined before
-          if (action.value.field === "segment") {
+          if (
+            action.value.field === "segment" &&
+            !isMapConfig(draft.chartConfig) // maps don't use segments
```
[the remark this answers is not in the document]
bprusinowski: Yes, that makes more sense in that case 👍

app/rdf/query-geoshapes.ts
```diff
@@ -0,0 +1,62 @@
...
+          iri: d.geoShapeIri.value,
+          wktString: d.WKT.value,
+        }));
+      })
+    );
+
+    return results.flat();
+  } else {
+    return [];
+  }
```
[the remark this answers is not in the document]
bprusinowski: Will try that! Added to the TODO list.

app/charts/map/map.tsx
bprusinowski: Yes, this kind of pattern will be changed, I marked some of it with FIXME label 😄 Probably it makes much more sense to save the data inside the object to avoid doing a search.

app/charts/map/map.tsx
bprusinowski: Generally yes, I left it for the moment because maybe it also makes sense to draw the "mesh" of the shapes, but maybe a regular border would be enough – honestly I didn't think about this yet

ptbrowne:
Apart from the small stuff I commented, the code is clear, thanks for having the smallish commits ;)

app/domain/env.ts
```diff
@@ -20,10 +20,8 @@ export const PUBLIC_URL = (
   ""
 ).replace(/\/$/, "");
 
-export const SPARQL_ENDPOINT =
-  clientEnv?.SPARQL_ENDPOINT ??
-  process.env.SPARQL_ENDPOINT ??
-  "https://int.lindas.admin.ch/query";
+export const SPARQL_ENDPOINT = "https://int.lindas.admin.ch/query";
+export const SPARQL_GEO_ENDPOINT = "https://ld.geo.admin.ch/query";
```
herrstucki: This should properly be read from the clientEnv/process.env like the SPARQL_ENDPOINT above.

app/domain/env.ts
```diff
@@ -20,10 +20,8 @@ export const PUBLIC_URL = (
   ""
 ).replace(/\/$/, "");
 
-export const SPARQL_ENDPOINT =
-  clientEnv?.SPARQL_ENDPOINT ??
-  process.env.SPARQL_ENDPOINT ??
-  "https://int.lindas.admin.ch/query";
+export const SPARQL_ENDPOINT = "https://int.lindas.admin.ch/query";
```
herrstucki, line 23: Please revert this change :) If you want to configure this differently, set it in `.env.development` or `.env.local`

app/configurator/map/map-chart-options.tsx
```diff
@@ -148,6 +150,49 @@ export const AreaLayerSettings = memo(
...
+              path="paletteType"
+              value={"jenks"}
+              disabled={!chartConfig.fields.areaLayer.show}
+            ></ChartOptionRadioField>
+            <ChartOptionSelectField
+              id="areaLayer.nbClass"
+              label="Number of classes"
+              field={activeField}
+              path="nbClass"
+              options={Array.from({ length: 7 }, (_, i) => 3 + i).map((d) => ({
```
herrstucki, line 197: nit: Move the creation of this (constant) options array into a const outside of the component function. E.g.

```ts
const CLASSES_OPTIONS = Array.from(/* ... */)

export const AreaLayerSettings = () => {
/*...*/
<ChartOptionSelectField
  id="areaLayer.nbClass"
  options={CLASSES_OPTIONS}
/>
}
```

app/domain/data.ts
```diff
@@ -19,6 +19,12 @@ export type GeoShape = {
   geometry: GeoJSONGeometry;
 };
 
+export type GeoPoint = {
+  iri: string;
+  label: string;
+  coordinates: [number, number];
```
herrstucki, line 25: Nit: better to do `{ iri:string, label:string, longitude:number, latitude:number }`?

herrstucki (commented):
Nice! Added a few comments but overall seems to work nicely 👏

app/domain/env.ts
bprusinowski: This is only for the local dev purposes, I have a point in TODO to revert this 😆 I just needed to change the endpoint there to access proper datasets, but will revert after the feature is ready.

app/domain/data.ts
bprusinowski: Ah, yes! Anyway I think this will be needed to change when merging with your branch :)

app/domain/env.ts
herrstucki: I know but seems like it's easy to forget :) If you want to change it for testing, I recommend using the `.env` files https://nextjs.org/docs/basic-features/environment-variables

app/charts/map/map.tsx
herrstucki: Just a quick note: Drawing the mesh is better because it's just 1 line for shared borders. If each shape draws its outline, this can look a bit messy.

ptbrowne:
Feedback from michael, the dataset cannot be loaded due to an Internal Server Error

image

https://visualize-ad-feat-integ-hqsuze.herokuapp.com/en/browse/dataset/https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fnfi%2F49-90-None-None%2Fcube%2F1?previous=%7B%7D

Will have to check this next year.

app/charts/map/chart-map-prototype.tsx
```diff
@@ -65,74 +55,78 @@ export const ChartMapVisualization = ({
...
-        d.properties.observation = observations?.find(
-          (o) => o[areaLayerComponentIri] === d.properties.label
-        );
-      });
-
-      return {
-        shapes: topojsonShapes,
-        mesh: topojsonMesh(rawShapes, rawShapes.objects.shapes as any),
-      };
+    if (dimensions) {
```
ptbrowne: Nit: to prevent a big diff while only the indentation level has changed, I think the early return pattern can be used.

if (!dimensions) {
  return
}

app/charts/map/chart-map-prototype.tsx
```diff
@@ -65,74 +55,78 @@ export const ChartMapVisualization = ({
...
-              } as GeoPoint)
-          );
+          ).map((d) => ({
+            coordinates: [d.longitude, d.latitude] as [number, number],
+            properties: {
+              iri: d.iri,
+              label: d.label,
+              observation: observations?.find(
+                (o) => o[symbolGeoDimensionIri] === d.label
+              ),
```
ptbrowne: A observations by label could be useful. Since it is needed in a lot of places, I wonder if we should have a DataProvider somewhere in the tree that does the indexing so that children do not have to do this repeatedly, what do you think ? (this is not the scope of the PR).

app/rdf/query-geo-shapes.ts
```diff
@@ -39,28 +45,139 @@ export const createGeoShapesLoader =
...
+
+  return go({
+    hierarchy: [],
+    currentLevelIris: narrowerValues
+      .filter((d) => !d.narrower)
+      .map((d) => d.iri),
+    allIris: dimensionIris,
+    narrowerValues,
+  });
+};
```
ptbrowne, line 175: Nit: this could be moved into the dimension-hierarchy file, and a test could be good :)

app/rdf/query-geo-shapes.ts
```diff
@@ -39,28 +45,139 @@ export const createGeoShapesLoader =
...
+    const nextLevelValues: HierarchyLevel[] = [];
+
+    for (const iri of currentLevelIris) {
+      const nextInHierarchy = narrowerValues.find((d) => d.narrower === iri);
+      const nextLevelIris = nextLevelValues.map((d) => d.iri);
+
+      if (nextInHierarchy && !nextLevelIris.includes(nextInHierarchy.iri)) {
+        nextLevelValues.push({ iri: nextInHierarchy.iri, level: level + 1 });
+      }
+    }
```
ptbrowne, line 161: Nit: It seems like grouping first by narrower would help. nextLevelValues also could be an object by iri so that the check for inclusion is faster.

app/charts/map/map-state.tsx
```diff
@@ -148,6 +152,61 @@ const useMapState = ({
   const getAreaValue = useOptionalNumericVariable(areaLayer.measureIri);
   const getSymbolValue = useOptionalNumericVariable(symbolLayer.measureIri);
 
+  const getDataByHierarchy = useCallback(
```
ptbrowne, line 151: getDataByHierarchyLevel would be a bit clearer IMHO

app/charts/map/map-state.tsx
```diff
@@ -148,6 +152,61 @@ const useMapState = ({
...
+
+      if (isGeoShapesDimension(dimension)) {
+        const hierarchyLabels = dimension.geoShapes.hierarchy
+          .filter((d: any) => d.level === hierarchyLevel)
+          .map((d: any) => d.label);
+
+        return data.filter((d) => hierarchyLabels.includes(getLabel(d)));
+      }
+
+      return data;
```
ptbrowne, line 171: Instead of having a callback, I think we could directly index the shapes by level so that we can use them directly afterwards without recomputation.

const dataByLevel = useMemo(() => { groupBy(data, x=> getLevelByLabel(x.label)) }

ptbrowne (approved):
I do not see any blocker for here. LGTM :) Great job !

A pattern I have noticed though in the code was the use of `find` and `filter` where indexing could be used and be clearer / faster.

app/charts/map/map-legend.tsx
```diff
@@ -122,55 +122,55 @@ const CircleLegend = () => {
...
+                            fontSize={legendFontSize}
+                          >
+                            {formatNumber(d)} ({thisFeatureLabel})
+                          </text>
+                        </>
+                      )}
+                    </g>
+                  }
+                </>
+              );
```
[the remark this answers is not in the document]
bprusinowski: After re-thinking this I don't think there is a good reason for that - I will change it :)

app/charts/map/map-legend.tsx
```diff
@@ -440,92 +380,78 @@ const QuantizeColorLegend = () => {
       .attr("text-anchor", "start");
   };
 
-  useEffect(() => {
-    const g = select(legendAxisRef.current);
-    mkAxis(g as Selection<SVGGElement, unknown, null, undefined>);
-  });
+  useEffect(() => (select(legendAxisRef.current) as any).call(mkAxis));
```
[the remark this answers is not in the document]
bprusinowski: Yeah, but then we would need useCallback for `mkAxis` and useMemo for `thresholds` and the result would be the same - right now the useEffect hook is anyway only called when the data, width or color palette changes and this also would be the case if we used useCallback + useMemo + `mkAxis` dependency in useEffect, so I'm not sure if there would be any benefit from doing this?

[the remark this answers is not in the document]
ptbrowne: The benefit I tsee would be clarity and resiliency to future code changes. Here it is a "coincidence" that useEffect works to our advantage.

app/charts/map/map-legend.tsx
```diff
@@ -74,96 +76,76 @@ export const MapLegend = () => {
...
-  const margins = {
-    top: 6,
-    right: 4,
-    bottom: 4,
-    left: 4,
-  };
-  const height = 60;
+  const value =
+    interaction.d && typeof getValue(interaction.d) === "number"
+      ? getValue(interaction.d)
```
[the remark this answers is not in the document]
bprusinowski: Sure, done! :)

app/charts/map/map-legend.tsx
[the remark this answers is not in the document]
bprusinowski: I extracted the `mkAxis` outside of the functions, I think it's better now :)

app/charts/map/map-legend.tsx
```diff
@@ -74,96 +76,76 @@ export const MapLegend = () => {
...
         })`}
       >
-        {symbolDataDomain.map((d, i) => {
+        {dataDomain.map((d, i) => {
           // FIXME: Potentially a performance problem if a lot of data
-          const observation = data.find((x) => getSymbolValue(x) === d);
-          const thisFeatureLabel = observation
-            ? getSymbolLabel(observation)
-            : "";
+          const observation = data.find((x) => getValue(x) === d);
```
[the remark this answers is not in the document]
bprusinowski: Done :)

app/charts/map/map-legend.tsx
```diff
@@ -123,49 +122,50 @@ const CircleLegend = () => {
...
+                          fontSize={legendFontSize}
+                        >
+                          {formatNumber(d)} ({thisFeatureLabel})
+                        </text>
+                      </>
+                    )}
+                  </g>
+                }
+              </>
+            )
```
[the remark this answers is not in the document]
bprusinowski: Good point, changed it!

app/configurator/map/map-chart-options.tsx
```diff
@@ -222,6 +223,11 @@ export const SymbolLayerSettings = memo(
     chartConfig: MapConfig;
     metaData: DataCubeMetadata;
   }) => {
+    const isDisabled = useMemo(
+      () => !chartConfig.fields.symbolLayer.show,
+      [chartConfig.fields.symbolLayer.show]
+    );
```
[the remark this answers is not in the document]
bprusinowski: Right, I wasn't sure if this isn't an overkill :D I will change that.

app/configurator/map/map-chart-options.tsx
```diff
@@ -95,42 +96,55 @@ export const AreaLayerSettings = memo(
...
-              path="fields.areaLayer.componentIri"
-              options={metaData.dimensions
-                .filter((d) => d.__typename === "GeoShapesDimension")
-                .map((d) => ({ value: d.iri, label: d.label }))}
-              disabled={!chartConfig.fields.areaLayer.show}
+              path="componentIri"
+              options={geoShapesDimensions.map((d) => ({
+                value: d.iri,
+                label: d.label,
+              }))}
```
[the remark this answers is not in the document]
bprusinowski: Changed also for following options, thanks!

app/rdf/query-geo-coordinates.ts
```diff
@@ -11,43 +13,49 @@ export interface RawGeoCoordinates {
...
  */
+const loadGeoCoordinates = async (dimension: ResolvedDimension) => {
+  const isVersioned = dimensionIsVersioned(dimension.dimension);
+
+  const query = SELECT.DISTINCT`?dimension ?label ?latitude ?longitude`.WHERE`
+    ${dimension.cube.term} ${ns.cube.observationSet} ?observationSet .
+    ?observationSet ${ns.cube.observation} ?observation .
+    ?observation ${dimension.dimension.path} ${
+    isVersioned
+      ? `?dimension_versioned . ?dimension_versioned schema:sameAs ?dimension ;`
```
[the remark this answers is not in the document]
bprusinowski: Right! I don't know why I left it like that, thanks :)

app/charts/map/map.tsx
```diff
@@ -226,7 +226,11 @@ export const MapComponent = () => {
               getLineWidth={100}
               updateTriggers={{ getFillColor: getColor }}
               getFillColor={(d: GeoShapeFeature) =>
-                getColor(getValue(d.properties.observation))
+                getColor(
+                  d.properties.observation
+                    ? getValue(d.properties.observation)
```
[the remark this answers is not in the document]
bprusinowski: Previously observation could not be undefined, but after connecting to a real data I believe it can; for example when there is no value for a given shape in a particular filter set. Not sure if this *should* happen, but it's already there in at least one of the datasets created by Fabian (OFPP POC Carte couverture cantons).

app/rdf/query-geo-coordinates.ts
```diff
@@ -0,0 +1,59 @@
...
+          ?geoCoordinateIri ${ns.schema.name} ?label
+        }
+
+        OPTIONAL {
+          ?geoCoordinateIri ${ns.schema.latitude} ?latitude
+        }
+
+        OPTIONAL {
+          ?geoCoordinateIri ${ns.schema.longitude} ?longitude
+        }
```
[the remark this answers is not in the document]
bprusinowski: Resolved in further commits :)

app/charts/map/map.tsx
```diff
@@ -100,7 +101,8 @@ export const MapComponent = () => {
       getRadius,
     },
   } = useChartState() as MapState;
-  const [, dispatch] = useInteraction();
+  const [, dispatchInteraction] = useInteraction();
+  const [mapTooltipState, dispatchMapTooltip] = useMapTooltip();
```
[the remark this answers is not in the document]
bprusinowski: Good point, wasn't sure about that either!

app/charts/map/map.tsx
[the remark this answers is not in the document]
ptbrowne: Are our types aware of this fact ?

app/domain/data.ts
```diff
@@ -110,6 +110,19 @@ export const getCategoricalDimensions = (
...
+) => dimensions.filter((d) => d.__typename === "GeoCoordinatesDimension");
+
+export const getGeoShapesDimensions = (
+  dimensions: DimensionMetaDataFragment[]
+) => dimensions.filter((d) => d.__typename === "GeoShapesDimension");
+
+export const getGeoDimensions = (dimensions: DimensionMetaDataFragment[]) =>
+  dimensions.filter((d) =>
+    ["GeoCoordinatesDimension", "GeoShapesDimension"].includes(d.__typename)
+  );
```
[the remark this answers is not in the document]
bprusinowski: I have improved the types there so it returns (GeoCoordinatesDimenions | GeoShapesDimensions)[], but it might make sense to use regular type guards in this case.

PR #242, opened 2021-11-30
Add datasets title

### LINKED ISSUE #241: TItle of root browsing page is not translated
Should be changed and translated to "All visualize.admin datasets"

herrstucki:
Fixes #241


herrstucki:
I left out the "visualize.admin" part because it's already in the header of the page and we never mention "visualize.admin" as a name elsewhere (if anything it should be "visualize.admin.ch"). So it's just "All datasets".

herrstucki:
I noticed there are some other translations missing. Will open another issue for that.

ptbrowne (approved):

PR #245, opened 2021-12-01
feat: Remove select dataset step from stepper

### LINKED ISSUE #244: Separate browsing mode from creation mode
@ptbrowne Together with BAFU we have agreed, that we won't add the Stepper on the Dataset-Page, but what we still have to implement is, that as soon as the user has clicked on "Create visualization" that we only show 3 steps. Step 1 = Chart type Step 2 = Visualize and Step 3 = Annotate. The first step "Dataset" needs to be removed

Screenshot 2021-12-01 at 09 08 01
.

_Originally posted by @AnninaWalker in https://github.com/visualize-admin/visualization-tool/issues/214#issuecomment-983393083_

ptbrowne:
Since the browsing mode is now separated from the chart creation, we chose to remove the step 1 "Selecting dataset" of the creation stepper.

fix #244 

- [x] Provide a mean to go back to the dataset selection

### DIFF
--- a/app/configurator/components/action-bar.tsx
+++ b/app/configurator/components/action-bar.tsx
@@ -1,5 +1,4 @@
 import { Trans } from "@lingui/macro";
-import NextLink from "next/link";
 import { ReactNode, useCallback } from "react";
 import { Button, Flex } from "theme-ui";
 import {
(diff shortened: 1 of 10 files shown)


ptbrowne:
If we remove the step 1, we do not have any mean to go back to dataset browsing (apart from going back with the browser). IMHO it was already a problem that the dataset title was not present in the panels before, I think we could add it to the first step, and it would also be a mean to go back to dataset browsing. Quick mockup

image

@AnninaWalker

AnninaWalker:
I like your proposal @ptbrowne

ortnever:
In your Mockup I see a link to go back to dataset Browser (choose another dataset). But don't you mind we should have the possibility to go back to Dataset preview at first and then to Dataset Browser?
image

ortnever:
it's also good like on the mockup. The user can always reselect the same dataset if he wants to review the preview. Two links would be too much.

ptbrowne:
> it's also good like on the mockup. The user can always reselect the same dataset if he wants to review the preview.

:+1:

I think here that you hint to the need of seeing the data in tabular form *while* being in the editing mode, I think it is completely fair to want to see the data in tabular form to better understand it, but then it should not the preview (but it will look like it), because the preview is during "browsing" mode. With the new browsing layout, there is a clear distinction between the "browsing" mode and the "editing" mode, and being able to go back to preview would break a bit the distinction here.

AnninaWalker:
Thank you for your feedback @ortnever . I would also suggest to currently keep it as it is in the mockup. 
As the user just came from previewing the dataset in order to get to the next step, most probably, instead of wanting to get back to the dataset preview, the user would probably prefer to change the dataset.

ptbrowne:
@ortnever I merged since the release is near and part of your feedback is "it also looks good in the mockup" :-) Nonetheless, I think the point you raised is good; we have other possible enhancement related to navigation in the backlog, and I think there will definitely be a time in the future to further discuss about it. Thanks again for your feedback.

ptbrowne:
I guess https://github.com/visualize-admin/visualization-tool/issues/249 will be the best issue to continue the discussion.

PR #246, opened 2021-12-01
fix: Allowing for empty key dimension time filter

### LINKED ISSUE #209: Mandatory filter not filled in
I can display a visualisation of bathing water quality data by leaving the "Date" filter empty, although it is a Key Dimension.
https://test.visualize.admin.ch/fr/create/owjFaUuj9OVq
image

bprusinowski:
Closes #209.

This PR allows to have an empty optional time filter (non-key dimension) and forces it to have some value for key dimensions.

### DIFF
--- a/app/configurator/components/field.spec.tsx
+++ b/app/configurator/components/field.spec.tsx
@@ -1,0 +1,71 @@
+import { fireEvent, render } from "@testing-library/react";
+import { getD3TimeFormatLocale } from "../../locales/locales";
+import { TimeInput } from "./field";
+
+describe("TimeInput", () => {
+  const expectedValue = "2020-05-24";
+  const setup = ({ isOptional, id }: { isOptional: boolean; id: string }) => {
+    let currentValue = "";
+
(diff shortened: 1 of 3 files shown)


app/configurator/components/field.spec.tsx
```diff
@@ -0,0 +1,71 @@
...
+        label={id}
+        value={currentValue}
+        timeFormat="%Y-%m-%d"
+        formatLocale={getD3TimeFormatLocale("en")}
+        isOptional={isOptional}
+        onChange={(e) => (currentValue = e.currentTarget.value)}
+      />
+    );
+
+    return { node: root.getByLabelText(id), getValue: () => currentValue };
```
ptbrowne, line 22: Nit: I think it is a little bit convoluted to use the functionality of the component to set a local variable. I think a spy on onChange would be a bit clearer, in the tests you would directly see that the onChange behavior is tested.

ptbrowne (approved):
Thanks !

PR #250, opened 2021-12-06
Homepage example change & time brush domain fix

bprusinowski:
This PR:
- replaces an example of the homepage (which stopped working due to some changes in the cube),
- fixes a bug where a chart time brush wasn't being updated when using an interactive filter.

I kept the current brush's behavior – so when you change an interactive filter option, the new time domain tries to fit into the old domain. It can be easily changed to reset the domain after each change, but I'm not sure which feels more natural.

### DIFF
--- a/app/charts/shared/brush.tsx
+++ b/app/charts/shared/brush.tsx
@@ -56,6 +56,11 @@ export const BrushTime = () => {
 
   brushWidthScale.range([0, brushWidth]);
 
+  const [minBrushDomainValue, maxBrushDomainValue] = useMemo(
+    () => brushWidthScale.domain().map((d) => d.getTime()),
+    [brushWidthScale]
+  );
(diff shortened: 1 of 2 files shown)


herrstucki (approved):
Example works, thanks! 👍 

The time brush now correctly resets the extent but keeps the selection when going from "Bundesarchiv" to "Bundesreisezentrale" and back. I think the comparison needs work a bit differently? E.g. tracking if the full extent is selected as opposed to a specific date range … until the user has actually selected something?

image

But maybe this is too much to fix now, from my perspective it's fine to release and fix later!

bprusinowski:
Yeah, I was also thinking about that – I will add the issue to not forget about this and introduce a fix afterwards (if I won't manage to fix this tomorrow before the release).

PR #255, opened 2021-12-08
fix: Expose client env via API endpoint

### LINKED ISSUE #254: window.__clientEnv__ is always empty
This is not a huge issue at the moment but can be problematic for debugging, since the debug panel accesses the env client-side.

herrstucki:
Because the process.env.* variables for some reason are only set in next.js dev mode inside of _document.tsx, the only way to expose them reliably is through an API route.

To ensure it's loaded before other scripts, it's imported as a synchronous script that sets `window.__clientEnv__` in the document `<head>`.

Fixes #254

### DIFF
--- a/app/pages/_document.tsx
+++ b/app/pages/_document.tsx
@@ -5,15 +5,8 @@ import Document, {
   NextScript,
   DocumentContext,
 } from "next/document";
+import { GA_TRACKING_ID } from "../domain/env";
 import { parseLocaleString } from "../locales/locales";
-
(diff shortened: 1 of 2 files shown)


herrstucki:
For the record, I tried using the [publicRuntimeConfig](https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration) approach but it broke the build for statically rendered pages like /docs and a few others.

ptbrowne (approved):

Issue #180, opened 2021-11-02
Search in Step 1 doesn't really filter out the typed in word as least for the word "Anzahl"


AnninaWalker:
If I type in the word "Anzahl" it doesn't really filter the dataset according to that: It also highlights single letters 🤷🏻‍♀️. It seems to work with other words though ;-)

https://user-images.githubusercontent.com/59613691/139867306-1085d365-6edb-48fb-a577-ee1898c19a54.mov

herrstucki:
Also reported in #105 and #96

PR #256, opened 2021-12-08
Search improvements

### LINKED ISSUE #96: Search datasets - Character String
Improvement :
The "Search datasets" field should only consider the character string and not each character.
Or to be discussed.
image

### LINKED ISSUE #105: Search full words
The actual feature searches by individual characters, not by words, with unexpected results:
image
image

### LINKED ISSUE #258: Improvements to search
Meta issue:

- Order of results: https://github.com/visualize-admin/visualization-tool/issues/251
- Full word search: https://github.com/visualize-admin/visualization-tool/issues/105

Search terms:

"Anzahl": https://github.com/visualize-admin/visualization-tool/issues/180
"Bade": https://github.com/visualize-admin/visualization-tool/issues/96
"OFPP"
"PRTR": https://github.com/visualize-admin/visualization-tool/issues/105
"MFM-U"

ptbrowne:
Still using server side search, but improving the search by using lunr.

- [x] Better support for acronyms (OFPP returns correct results, MFM-U returns correct results), fix #105
- [x] Less fuzzy search fix #96 #180
- [x] Ability to search inside themes / creator
- [x] Stemming

Fix #258


ptbrowne:
@fabiancretton, @pdumas, @ortnever I'd be interested on your feedback concerning the search on this deployment. I think it's much better than before. I want to know if I include this in the release of Monday. Thanks !

You can test here: https://visualize-ad-search-lk1hpzdeez.herokuapp.com/fr/browse

cc @sosiology

ortnever:
Thanks for this improvment.
It works great but unfortunately only if the keyword is in the title. As the title is often very short, it requires very precise keywords. Would it be possible to do exactly the same thing by also looking for the word in the description?

ptbrowne:
Can you give me your search @ortnever ? The search should works in the description:

image

Additionally, it searches in the themes, in the creator, and in the keywords (that are not displayed yet on visualize).

ortnever:
It was "coli" but I tried again and it works. 
It is very good. 
To be deployed on Monday if you are sure there are no side effects elsewhere but there is no reason why the research should influence anything else.

FabianCretton:
i did only a small test, but yes,  looks much better, thanks!

app/utils/search.ts
```diff
@@ -19,3 +25,130 @@ export const wrap = (
...
+lunr.Pipeline.registerFunction(hyphenator, "hyphenator");
+
+var customHiphenatorPipeline = function (builder: lunr.Builder) {
+  builder.pipeline.before(lunr.stemmer, hyphenator);
+  builder.searchPipeline.before(lunr.stemmer, hyphenator);
+};
+
+const highlightMatch = (match?: Fuse.FuseResultMatch) => {
+  return match?.value
+    ? wrap(match.value, match.indices, {
```
bprusinowski, line 78: Is it a correct type for the `match` – you removed the Fuse in a previous commit, so I am not sure?

app/utils/search.ts
```diff
@@ -19,3 +25,130 @@ export const wrap = (
...
+  var idx = lunr(function () {
+    const self = this;
+    self.use(customHiphenatorPipeline);
+    self.ref("iri");
+    self.field("title", { boost: 2 });
+    self.field("description");
+    self.field("publisher");
+    self.field("themes");
+    self.field("keywords");
+    self.field("creator");
```
bprusinowski, line 99: I wonder if `index` shouldn't be created just once (or once per locale) and not on every search? I guess it's not a problem ATM when we don't have hundreds of cubes, I am just curious :)

app/utils/search.ts
```diff
@@ -19,3 +25,130 @@ export const wrap = (
...
+      const themes = doc.themes?.map((t) => t.label || "").join(" ") ?? "";
+      const keywords = doc.keywords?.map((t) => t || "").join(" ") ?? "";
+      self.add({
+        ...doc,
+        themes,
+        keywords,
+        creator: doc.creator?.label,
+      });
+      console.log({ themes, keywords, creator: doc.creator?.label });
+    }, this);
```
bprusinowski, line 112: 👀

bprusinowski (approved):
LGTM! Works much better than before 👌

app/utils/search.ts
ptbrowne: No longer there :)

app/utils/search.ts
ptbrowne: Good point ! Was able to remove the type, then the import, then the package ;)

app/utils/search.ts
ptbrowne: Good point, https://github.com/visualize-admin/visualization-tool/pull/256/commits/d1f579bd0e4cc1901bc078ca5d7c611778c6b508

bprusinowski: Great, thanks! 👌

PR #262, opened 2021-12-09
Include Position for Ordinal Dimensions

herrstucki:
A less disruptive approach than #227

### To Do
- [x] Implement query and resolver for `position`

### DIFF
--- a/app/charts/shared/use-chart-state.tsx
+++ b/app/charts/shared/use-chart-state.tsx
@@ -36,6 +36,11 @@ export type ChartState =
   | MapState
   | undefined;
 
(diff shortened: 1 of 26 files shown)


app/charts/shared/use-chart-state.tsx
```diff
@@ -36,6 +36,11 @@ export type ChartState =
   | MapState
   | undefined;
 
+type Has<T extends unknown, R extends string> = T extends { [k in R]: any }
+  ? T
+  : never;
+
+export type ColorsChartState = Has<ChartState, "colors">;
```
bprusinowski, line 43: 👌

bprusinowski:
LGTM! 👍

PR #263, opened 2021-12-09
docs: Add cube update rules

ptbrowne:
Add documentation on how to safely update a cube so that charts still work.

### DIFF
--- a/app/docs/cube-update-constraints.mdx
+++ b/app/docs/cube-update-constraints.mdx
@@ -1,0 +1,26 @@
+## Constraints
+
+When updating a cube, you have to follow a number of rules to make sure that
+the charts that rely on those cubes will continue to work.
+
+- Structure of a dimension should not be changed, otherwise, if a filter relies on the dimension, it might
+  not work anymore. Here are the properties of a dimension that should not change:
+
+  - `a`: Dimension type
+  - `qudt:unit`: Unit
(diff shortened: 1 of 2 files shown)


herrstucki (approved):

l00mi:
Noooo, I was writing a review ((-..

Minor thing, can we remove the elipsis .. and just state what can not change of the structure. Its these three things, datatype, datakind, scale of measure and dimension type. What about Unit?

ptbrowne:
Thanks good idea to remove the ellipsis and be explicit. I went through the cube schema doc to list the properties.

l00mi:
Thank you, very good. It add a suprise for me, why would the `sh:order` invalidate an update?

ptbrowne:
I added it as a precaution. I struggled a bit to find an example but for example for maps, the order could be used for the z-order of layers and changing the order could make one shape go below another one. It might be over-zealous I concede. We could re add it if necessary.

l00mi:
> I added it as a precaution. I struggled a bit to find an example but for example for maps, the order could be used for the z-order of layers and changing the order could make one shape go below another one. It might be over-zealous I concede. We could re add it if necessary.

No example yet, we need to implement it still in the cube-creator. The cubes from WSL, not created by Cube-Creator might already provide it.

PR #264, opened 2022-01-11
Try to fetch dimension values when no values are there initially

bprusinowski:
Right now the logic brings back the previous functionality – tries to fetch all of the dimension values if no values are present initially. We still need to wait for autocomplete API from Zazuko and I'm not quite sure how this should look like in UI (and the code), so I didn't try to come up with new ideas, it probably would make sense to also consult this with Annina.

Let me know what do you think (and if this should be already merged to main – if not, I can merge this to maps PR to continue working on it if the code looks fine).

### DIFF
--- a/app/rdf/queries.ts
+++ b/app/rdf/queries.ts
@@ -1,6 +1,6 @@
 // import { sparql } from "@tpluscode/rdf-string";
 // import { descending } from "d3";
-import { descending, index, rollup } from "d3";
+import { descending, group, index, rollup } from "d3";
 import {
   Cube,
   CubeDimension,
@@ -312,14 +312,6 @@ const groupLabelsPerValue = ({
 const dimensionIsVersioned = (dimension: CubeDimension) =>
   dimension.out(ns.schema.version)?.value ? true : false;
 
-const filterCubeDimensionValuesByTermType = ({
-  values,
-  termType,
-}: {
-  values: (Literal | NamedNode)[] | undefined;
-  termType: "Literal" | "NamedNode";
-}) => values?.filter((d) => d.termType === termType) ?? [];
-
 const getCubeDimensionValuesWithLabels = async ({
   dimension,
   cube,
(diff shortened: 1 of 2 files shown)


app/rdf/queries.ts
```diff
@@ -323,13 +329,14 @@ const getCubeDimensionValuesWithLabels = async ({
...
-    (v) => v.termType === "Literal"
-  ) ?? []) as Literal[];
+  let dimensionValueNamedNodes = filterCubeDimensionValuesByTermType({
+    values: dimension.in,
+    termType: "NamedNode",
+  }) as NamedNode[];
+  let dimensionValueLiterals = filterCubeDimensionValuesByTermType({
+    values: dimension.in,
+    termType: "Literal",
+  }) as Literal[];
```
ptbrowne, line 339: I think it would be a bit simple not to have the function `filterCubeDimensionValuesByTermType`. It not very composable and hides the `filter` function.

Maybe we could use groupBy here ?

```
let { dimensionValueLiterals = [], dimensionValueNamedNodes = [] } = groupBy(dimension.in, x => x.termType)
```

Downside is that to add the right types is a bit more difficult.

At first I had thought that only having type guards functions would be also OK.

```
const isLiteral = (x: NamedNode | Literal): x is Literal => x.termType === 'Literal'
const isNamedNode = (x: NamedNode | Literal): x is NamedNode => x.termType === 'NamedNode'

const dimensionValueLiterals = (dimension.in || []).filter(isLiteral) // dimensionValueLiterals should be Literal[]
```

app/rdf/queries.ts
```diff
@@ -349,9 +356,29 @@ const getCubeDimensionValuesWithLabels = async ({
...
+      );
+
+      dimensionValueNamedNodes = filterCubeDimensionValuesByTermType({
+        values: unpackedDimensionValues,
+        termType: "NamedNode",
+      }) as NamedNode[];
+      dimensionValueLiterals = filterCubeDimensionValuesByTermType({
+        values: unpackedDimensionValues,
+        termType: "Literal",
+      }) as Literal[];
```
ptbrowne, line 376: Nit: Seems like the logic is repeated from above no ? Maybe it could be expressed in the way "there is two places to take the dimensionValues, either in dimension.in or through loadDimensionValues ?

```
const loadDimensionValues = async () => {
	  const dimensionLoaders = [() => dimension.in, () => loadDimensionValues({ })]
	  
	  for (let dimensionLoader of dimensionLoaders) {
	    const dimensionValues = await dimensionLoader()
	    // then groupBy named nodes and literals
	    const { NamedNode: namedNodes = [], Literal: literals = [] } = groupBy(dimensionValues, x => x.termType)
	    if (namedNodes.length || literals.length) {
	      return { namedNodes, literals }
	    }  
	  }
          console.warn('no values...')
}
```

(untested)

ptbrowne (approved):
Two nits, LGTM :)

bprusinowski:
Thanks @ptbrowne! I know it was NIT but I really liked your approach :)

ptbrowne (approved):
Cool ;)

PR #265, opened 2022-01-11
fix: Make column groups draggable on all the column row

ptbrowne:
Buttons are not draggable with react-dnd. We have to put `as='div'` onto theme ui Button.


bprusinowski (approved):
LGTM!

PR #269, opened 2022-01-20
fix: Geo data loaders locale switching bug

bprusinowski:
Basing on information from https://nextjs.org/docs/advanced-features/i18n-routing:

- locale is accessible through document context, there is no need to retrieve it in such complex way,
- lang is automatically appended to HTML, so there is no need to set it explicitly.

The way of setting the locale was **always** returning a default locale (de) initially, probably because Next.js introduced some changes since the project was created and then it was changed through _app.tsx.

Additionally, retrieving the lang property by using querySelector in urql client was resulting in an undefined locale for the initial request, which introduced some bugs in fetching of the shapes, which isn't the case anymore.

### DIFF
--- a/app/graphql/context.tsx
+++ b/app/graphql/context.tsx
@@ -1,23 +1,16 @@
 import { ReactNode } from "react";
 import { createClient, Provider } from "urql";
 import { GRAPHQL_ENDPOINT } from "../domain/env";
-import { defaultLocale } from "../src";
+import { useLocale } from "../src";
 
-const client = createClient({
-  url: GRAPHQL_ENDPOINT,
-  fetchOptions: () => {
-    const lang =
-      typeof document !== "undefined"
-        ? document.querySelector("html")?.getAttribute("lang")
-        : undefined;
-    return {
-      headers: {
-        "Accept-Language": lang ? lang : defaultLocale,
-      },
-    };
-  },
-});
+const client = createClient({ url: GRAPHQL_ENDPOINT });
 
 export const GraphqlProvider = ({ children }: { children: ReactNode }) => {
+  const locale = useLocale();
+
+  client.fetchOptions = () => ({
+    headers: { "Accept-Language": locale },
+  });
+
   return <Provider value={client}>{children}</Provider>;
 };
--- a/app/graphql/context.tsx
+++ b/app/graphql/context.tsx
@@ -1,4 +1,4 @@
-import { ReactNode } from "react";
+import { ReactNode, useCallback } from "react";
 import { createClient, Provider } from "urql";
 import { GRAPHQL_ENDPOINT } from "../domain/env";
 import { useLocale } from "../src";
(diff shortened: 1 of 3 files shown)


app/pages/_app.tsx
```diff
@@ -37,10 +36,6 @@ export default function App({ Component, pageProps }: AppProps) {
     i18n.activate(locale);
   }
 
-  useEffect(() => {
-    document.querySelector("html")?.setAttribute("lang", locale);
-  }, [locale]);
```
ptbrowne, line 42: Is it really necessary to remove this ? I am not sure this should be removed as it could be used somewhere else (maybe browser extensions also rely on this to propose translation to the user).

app/graphql/context.tsx
```diff
@@ -1,23 +1,16 @@
...
-  },
-});
+const client = createClient({ url: GRAPHQL_ENDPOINT });
 
 export const GraphqlProvider = ({ children }: { children: ReactNode }) => {
+  const locale = useLocale();
+
+  client.fetchOptions = () => ({
+    headers: { "Accept-Language": locale },
+  });
```
ptbrowne, line 13: I think this should be wrapped into a useEffect with locale as depedency, for clarity (we are effecting something).

app/pages/_document.tsx
```diff
@@ -1,31 +1,7 @@
...
-     * Parse locale from query OR pathname
-     * - so we can have dynamic locale query params like /[locale]/create/...
-     * - and static localized pages like /en/index.mdx
-     */
-    const locale = /^\/\[locale\]/.test(pathname)
-      ? parseLocaleString(query.locale?.toString() ?? "")
-      : parseLocaleString(pathname.slice(1));
-
-    return { ...initialProps, locale };
-  }
```
ptbrowne, line 27: I am not sure why this is necessary.

app/pages/_document.tsx
```diff
@@ -4,10 +4,7 @@ import { GA_TRACKING_ID } from "../domain/env";
 class MyDocument extends Document {
   render() {
     return (
-      <Html
-        data-app-version={`${process.env.NEXT_PUBLIC_VERSION}`}
-        lang={this.props.locale}
-      >
+      <Html data-app-version={`${process.env.NEXT_PUBLIC_VERSION}`}>
```
ptbrowne, line 7: Since in later commits the setting of the lang is removed, why don't we keep it here ?

ptbrowne: Understood, this is done automatically by next.js now, settings it directly on document is deprecated.

app/pages/_document.tsx
ptbrowne: Understood, this is done automatically by next.js now, settings it directly on document is deprecated.

app/graphql/context.tsx
ptbrowne: Cannot use useEffect since the setting of fetchOptions will be too late (useEffects are run asynchronously), and maybe network calls (with wrong locale) will already have been fired by children.

ptbrowne: Nit: Maybe useMemo could be used here to have the clarity that fetchoptions only changes when locale changes.

bprusinowski: I added useCallback there, thanks for the tip!

app/pages/_app.tsx
bprusinowski: It's being set automatically by Next.js + i18n

bprusinowski:
@ptbrowne I set the PR to automerge, if you think it's ready you can approve 👀

ptbrowne (approved):
👍  Thanks !

PR #271, opened 2022-01-21
Cascading facets

ptbrowne:
This is a work in progress to rework the left filters into cascading facets.

- [x] Ability to fetch unique values for a dimension taking into account filters
- [x] Order of the filters is used so that filters below have values corresponding to filters above
- [x] Ability to reorder the filters (UX is to be improved)
- [ ] See how to manage adding new filters


app/rdf/query-dimension-values.ts
```diff
@@ -1,29 +1,133 @@
...
+  },
+  filters?: Filters
+): Promise<Array<Literal | NamedNode>> {
+  const dimensionIri = dimension.path;
+
+  let filterList = filters ? Object.entries(filters) : [];
+  filterList = filterList.slice(
+    0,
+    filterList.findIndex(([iri]) => iri == dimensionIri?.value) + 1
+  );
```
bprusinowski, line 91: Why are we slicing here?

app/rdf/query-dimension-values.ts
```diff
@@ -1,29 +1,133 @@
...
+                dimensionIri?.value === iri
+              ) {
+                return "";
+              }
+              const versioned = filterDimension
+                ? dimensionIsVersioned(filterDimension)
+                : false;
+              return `${
+                versioned
+                  ? `?dimension${idx} <http://schema.org/sameAs> ?dimension_unversioned${idx}.`
```
bprusinowski, line 116: I think you also had an issue with using ${ns.schema.sameAs} here? I faced it in geoCoordinates query, you can take a look: https://github.com/visualize-admin/visualization-tool/blob/main/app/rdf/query-geo-coordinates.ts#L31. Not sure if this is an elegant solution, but it does a trick :)

app/configurator/components/chart-configurator.tsx
```diff
@@ -61,20 +102,78 @@ export const ChartConfigurator = ({
   state: ConfiguratorStateConfiguringChart;
 }) => {
   const locale = useLocale();
-  const [{ data }] = useDataCubeMetadataWithComponentValuesQuery({
-    variables: {
+  const [, dispatch] = useConfiguratorState();
+  const variables = React.useMemo(
+    () => ({
+      date: new Date(),
```
bprusinowski, line 132: Why do you have a `date` here?

app/components/form.tsx
```diff
@@ -292,6 +297,90 @@ export const Input = ({
...
+        minHeight: "1.5rem",
+        display: "block",
+        borderRadius: "0.25rem",
+        width: "100%",
+        border: "1px solid var(--theme-ui-colors-monochrome500)",
+
+        // @ts-ignore
+        ...props.inputProps?.style,
+      },
+      ...props.inputProps,
```
bprusinowski, line 347: Not sure if `style` wouldn't be overwritten if `props.inputProps` contains them?

bprusinowski (approved):
LGTM 👌

app/rdf/query-dimension-values.ts
ptbrowne: To apply above filters onto below filters

Ex: 3 columns, ordered as follow : date, monitoring program, bathing site
In this case, 

* for date, no filter apply,
* for monitoring program, date applies
* for bathing site, date and monitoring program apply

bprusinowski: Ah, so they are ordered by their hierarchy on the left panel? It makes sense now 👍

ptbrowne: No they are not yet ordered by hierarchy, the user has to do it. But when we have hierarchies, the default order will be by hierarchy.

app/rdf/query-dimension-values.ts
ptbrowne: Mmh yes but my problem is that if I use the ns.schema.sameAs.toString(), it will correctly expand it to `<https://...>` but my problem is that the prefix will not be added to the query since the usage of the sparql bit will not be detected (since it is nested and already resolved).

```
const { SELECT, sparql } = require("@tpluscode/sparql-builder");
const { schema } = require("@tpluscode/rdf-ns-builders");

const filters = ["a"];
const query = SELECT.DISTINCT`?value`.WHERE`
    ${filters.map(
      (f) =>
        `FILTER (( ${sparql`${schema.sameAs}`.toString({
          prologue: true,
        })}) = 1 )`
    )}
  `;

console.log(query.build());
```

```
SELECT DISTINCT ?value

WHERE {

    FILTER (( schema:sameAs) = 1 )
  
}
```

Here the prefix is not added.

I will post an issue on sparql builder

bprusinowski: Hmmm, right, I see that I successfully included` ${ns.schema.[smth]}` in geo coords query, so it was already appended in that case.  Good 👍

ptbrowne: https://github.com/tpluscode/sparql-builder/issues/62 ;)

bprusinowski: Nice :D

ptbrowne:
@bprusinowski I rebased and pushed new commits, I think you finished your last review at 
af52a0d .

app/configurator/components/chart-configurator.tsx
ptbrowne: Added some docs about this in a new commit, this is to force urql to fetch with the new filters. Otherwise urql does not check the order of the filters.

app/components/form.tsx
ptbrowne: Oups, you're right ;) Moved inputprops above.

app/configurator/components/chart-configurator.tsx
bprusinowski: Great, thanks!

bprusinowski:
Yes, I will review the changes tomorrow morning :) Thanks for letting me know!

app/configurator/components/chart-configurator.tsx
```diff
@@ -120,6 +118,33 @@ const DataFilterSelectGeneric = ({
...
+  className,
+}: {
+  onClickUp: () => void;
+  onClickDown: () => void;
+  className?: string;
+}) => {
+  return (
+    <>
+      <Button className={className} variant="arrow" onClick={onClickUp}>
+        ▲
```
bprusinowski: This probably doesn't matter at all, but shouldn't we use <Icon name="caretUp" />?

app/icons/components/IcDragndrop2.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from "react";
+
+function SvgIcDragndrop2(props: React.SVGProps<SVGSVGElement>) {
```
bprusinowski, line 3: `function SvgIcDragndropCompact`?

app/configurator/components/chart-configurator.tsx
```diff
@@ -154,10 +154,14 @@ export const ChartConfigurator = ({
...
     () => ({
-      date: new Date(),
       iri: state.dataSet,
       locale,
       filters: state.chartConfig.filters,
+      // This is important for urql not to think that filters
+      // are the same  while the order of the keys has changed.
+      // If this is not present, we'll have outdated dimension
+      // values after we change the filter order
+      date: new Date(),
```
bprusinowski, line 136: 👌

app/configurator/components/chart-configurator.tsx
bprusinowski: Nevermind, I see that you changed it in following commits :)

bprusinowski:
Apart from one comment about a drag n drop icon name, everything looks good!

app/icons/components/IcDragndrop2.tsx
ptbrowne: Thanks, I removed the older the older dnd icon.

ptbrowne:
Thanks good catch, this was due to styling for the panel left being applied a bit too broadly, I moved the styles only to the PanelLeft in the chart configurator. Those styles are there for the content of the panel left to take all the space available.

PR #273, opened 2022-01-24
fix: Sort by relevance by default

ptbrowne:
The score by default was "title", by mistake, which meant the results were
alphabetically ordered instead of being ordered by relevance.﻿


bprusinowski (approved):
LGTM :)

PR #289, opened 2022-01-25
fix: Weird shapes colors

### LINKED ISSUE #281: Maps: Weird Drawing of shapes
Used the following dataset: LFI Data Cube [Version 1]: Totholzvolumen / Produktionsregion
When I change the filter "unit of evaluation" in the left side panel to "accessible forest without shrub forest NF13/NFI4" it create some odd shades (or modern art ;-) in the map.
See Screenshot

Screenshot 2022-01-25 at 09 11 00

bprusinowski:
Closes #281.

Previously, if no observation was mapped to a shape (or its measure value was equal to `null`), a `null` was used as a shape fill color. This led to weird bugs - changing this to a light gray color fixes the issue.

### DIFF
--- a/app/charts/map/map.tsx
+++ b/app/charts/map/map.tsx
@@ -242,7 +242,7 @@ export const MapComponent = () => {
                   ? areaLayer.getColor(
                       areaLayer.getValue(d.properties.observation)
                     )
-                  : null
+                  : [33, 33, 33, 33]
               }
             />
             <GeoJsonLayer


ptbrowne (commented):
it means that if there a lot of observations with null measures, we will have a lot of gray shapes on the map ? Why do we draw the shape at all ? Couldn't we just not draw the shape ?

bprusinowski:
Yes, I think it's a common approach to show shapes with no data as "greyed out" - let's say that somebody configured a map with no relief, and that all cantons are visible at first. But there are also interactive filters enabled and after doing some filtering only one canton has some data, then it would be alone on the map without any context. I just thought that it makes sense, do you think it should be done differently?

ptbrowne:
I see, I had not the interactive filter with no data in mind. Shouldn't it be managed by the base layer then ?

AnninaWalker:
We have also forseen to add a legend for either "No value" or "No data" in the Design Mockup. In the Covid Dashboard it is dealt with it in exactly the same way.

ptbrowne:
> We have also forseen to add a legend for either "No value" or "No data" in the Design Mockup. 

If you filter out some cantons, is it not because you do not want to see them on the map ? Anyway, I think it's good that we have no problems for the moment, the fix is good to me ;) LGTM

AnninaWalker:
Screenshot 2022-01-25 at 15 18 05

bprusinowski:
> If you filter out some cantons, is it not because you do not want to see them on the map ?

Generally yes, but I think that we aren't able to filter by geo dimensions in interactive filters, so this should only happen when we filter by some other filter and encounter missing data there :)

I will have to check the logic on including dimensions in interactive filters, but this is the case for current testing datasets.

ptbrowne:
I see, thanks for the explanations 💯

bprusinowski:
@ptbrowne if you think everything looks good, you can approve so the bug shouldn't happen anymore (maybe BAFU has already started testing the maps). I'm going to handle "no data" on legend in a different PR :)

ptbrowne:
Oups sorry I had forgotten to review again.

PR #294, opened 2022-01-25
feat: New design for color scale palette picking for maps

### LINKED ISSUE #275: Maps: Area Layer - Design of Color Selection
@bprusinowski Please adapt the design of the Color Selection according to the Design Mockups.

- First the user can select if the Color Palette should be Continuous or in Steps. 
- Then he selects the type of Interpolation. The selection should be in a dropdown instead of checkboxes. 
- Colors should be displayed as Color Palettes and not in text (oranges, purples etc.

Current implementation:
Screenshot 2022-01-24 at 16 32 57

Design Screens:
3 1 3_l_map_chart-maker_step-33 1 2_l_map_chart-maker_step-3
t-maker_step-3.png…]()
3 1 4_l_map_chart-maker_step-3
3 1 5_l_map_chart-maker_step-3

bprusinowski:
Closes #275.

Besides updating design:
- added ColorRampField (color picker which uses color ramps)
- improved colors naming conventions (previously e.g. divergent scales were named as sequential)
- added types for divergent and sequential color scales for maps & tables.

If you think this approach is good, then I can create an issue to also do this for other chart types and categorical color scales in the future.


bprusinowski:
@AnninaWalker I still need to replace a select element for choosing a number of breaks and use a text input; but if you could take a look and let me know if it looks good now, it would be helpful :)

https://visualize-ad-feat-new-d-mkjyqc.herokuapp.com/

AnninaWalker:
@bprusinowski It looks great! And honestly, I think the dropdown for the steps works even better, than an inputfield. I think I would leave it as it is!

bprusinowski:
@AnninaWalker it's great to hear that! I'll leave the dropdown then :)

app/charts/index.ts
```diff
@@ -278,9 +278,10 @@ export const getInitialConfig = ({
             componentIri: geoShapes[0]?.iri || "",
             measureIri: measures[0].iri,
             hierarchyLevel: 1,
+            colorScaleType: "continuous",
+            colorScaleInterpolationType: "linear",
             palette: "oranges",
             nbClass: 5,
-            paletteType: "continuous",
```
ptbrowne, line 283: I guess that if we open a chart with an old config with this new change, we will not be able to open the new chart right ? I wonder how to manage breaking changes within the config.
The best would be to migrate the config before ingestion into io-ts I guess.

app/configurator/map/map-chart-options.tsx
```diff
@@ -221,47 +226,65 @@ export const AreaLayerSettings = memo(
...
+                        label: "Quantize (equal intervals)",
+                        value: "quantize",
+                      },
+                      {
+                        label: "Quantiles (equal distribution of values)",
+                        value: "quantile",
+                      },
+                      {
+                        label: "Jenks (natural breaks)",
+                        value: "jenks",
```
ptbrowne, line 273: These strings do not seem translated.

app/configurator/components/ui-helpers.ts
```diff
@@ -726,60 +767,66 @@ export const getColorInterpolator = (
...
+
+export const sequentialPalettes = sequentialPaletteKeys.map((d) => ({
+  label: d,
+  value: d,
+  interpolator: getColorInterpolator(d),
+})) as {
+  label: string;
+  value: SequentialPaletteType;
+  interpolator: (t: number) => string;
+}[];
```
ptbrowne, line 819: Nit: Why do we have to explicitly type here ? I think it would be a bit clearer and useful to have a named SequentialPalette type and then you do the `as SequentialPalette` inside the map.

app/configurator/components/ui-helpers.ts
```diff
@@ -726,60 +767,66 @@ export const getColorInterpolator = (
...
+
+export const sequentialSteppedPalettes = sequentialPaletteKeys.map((d) => ({
```
ptbrowne, line 821: Aren't they diverging palettes now ?

app/configurator/components/chart-controls/color-ramp.tsx
```diff
@@ -1,34 +1,209 @@
...
+};
+
+export const ColorRamp = ({
+  colorInterpolator,
+  nbClass = 512,
+  width = 148,
+  height = 28,
+}: ColorRampProps) => {
   const canvasRef = useRef<HTMLCanvasElement>(null);
-  // Clear by returning a function from usEEFFECT?
```
ptbrowne, line 16: 😅

app/configurator/components/chart-controls/color-ramp.tsx
```diff
@@ -1,34 +1,209 @@
...
+
+  const palettes = useMemo(
+    () => [...divergingPalettes, ...sequentialPalettes],
+    []
+  );
+
+  const currentPaletteName = get(
+    state,
+    `chartConfig.fields.${field}.${path}`
+  ) as DivergingPaletteType | SequentialPaletteType;
```
ptbrowne, line 78: Nit: I am not a fan of using get here since we lose type information and have to readd it manually, can't we use optional chaining here ?

app/configurator/components/chart-controls/color-ramp.tsx
```diff
@@ -1,34 +1,209 @@
...
+  const {
+    isOpen,
+    getToggleButtonProps,
+    getLabelProps,
+    getMenuProps,
+    highlightedIndex,
+    getItemProps,
+  } = useSelect({
+    items: palettes,
+    defaultSelectedItem: palettes.find((d) => d.value === "oranges"),
```
ptbrowne, line 92: This could be extracted as a const right ? (I am a bit triggered by `find`s in render 😅 )

app/configurator/components/chart-controls/color-ramp.tsx
```diff
@@ -1,34 +1,209 @@
...
+  field,
+  path,
+  nbClass,
+}: ColorRampFieldProps) => {
+  const [state, dispatch] = useConfiguratorState();
+
+  const palettes = useMemo(
+    () => [...divergingPalettes, ...sequentialPalettes],
+    []
+  );
```
ptbrowne, line 73: This is a const right ?

app/configurator/components/chart-controls/color-ramp.tsx
```diff
@@ -1,34 +1,209 @@
...
+      if (selectedItem) {
+        dispatch({
+          type: "CHART_OPTION_CHANGED",
+          value: {
+            field,
+            path,
+            value: selectedItem.value,
+          },
+        });
+      }
```
ptbrowne, line 103: Nit: Use callback could be used.

app/configurator/components/chart-controls/color-ramp.tsx
```diff
@@ -1,34 +1,209 @@
...
+          borderColor: "monochrome500",
+          ":hover": {
+            bg: "monochrome100",
+          },
+          ":active": {
+            bg: "monochrome100",
+          },
+          ":disabled": {
+            cursor: "initial",
+            bg: "muted",
```
ptbrowne, line 133: Could we add it to the button variants in the theme or is it really specific to this place ?

app/configurator/components/chart-controls/color-ramp.tsx
```diff
@@ -1,34 +1,209 @@
...
+            </Text>
+            {sequentialPalettes.map((d, i) => (
+              <PaletteRamp
+                key={`sequential-${i}`}
+                palette={d}
+                backgroundColor={
+                  i + divergingPalettes.length === highlightedIndex
+                    ? "monochrome200"
+                    : "monochrome100"
+                }
```
ptbrowne, line 176: Nit: Seems like the PaletteRamp could handle a `highlighted` prop.

app/configurator/components/chart-controls/color-ramp.tsx
```diff
@@ -1,34 +1,209 @@
...
+    <Box sx={{ p: 1, cursor: "pointer", backgroundColor }}>
+      <Box sx={{ backgroundColor }} {...itemProps}>
+        <ColorRamp
+          key={`option-${palette.value}`}
+          colorInterpolator={palette.interpolator}
+          nbClass={nbClass}
+        />
+      </Box>
+    </Box>
+  );
```
ptbrowne, line 213: Wondering if the PaletteRamp be added to the docs ?

app/configurator/config-types.ts
```diff
@@ -376,7 +412,7 @@ const ColumnStyleCategory = t.type({
 const ColumnStyleHeatmap = t.type({
   type: t.literal("heatmap"),
   textStyle: ColumnTextStyle,
-  palette: t.string,
+  palette: DivergingPaletteType,
```
ptbrowne, line 415: Nice 👍

app/charts/map/map-state.tsx
```diff
@@ -96,22 +96,24 @@ const getColorScale = ({
...
-  const paletteDomain = getSingleHueSequentialPalette({
-    palette,
-    nbClass: 9,
-  });
+  const interpolator = getColorInterpolator(palette);
+  const getDiscreteRange = () => {
+    return Array.from({ length: nbClass }, (_, i) =>
+      interpolator(i / (nbClass - 1))
+    );
+  };
```
ptbrowne, line 104: TIL that array from can take a function map as 2nd argument 👍 I would always map in the past to achieve the same effect.

app/configurator/map/map-chart-options.tsx
```diff
@@ -133,8 +133,11 @@ export const AreaLayerSettings = memo(
       [metaData.measures]
     );
 
-    const numberOfGeoShapes = (dimension.geoShapes as any).topology.objects
-      .shapes.geometries.length as number;
+    const numberOfGeoShapes = (
+      dimension
+        ? (dimension.geoShapes as any).topology.objects.shapes.geometries.length
```
ptbrowne, line 138: I wonder why we have to resort to any here ?

ptbrowne (approved):
Thanks, looking good, my comments are nits, LGTM !

app/configurator/components/ui-helpers.ts
bprusinowski: Nope, `sequentialPaletteKeys` contain "blues", "greens", etc. Diverging are the ones with two color shortcuts in name, e.g. "PuOr".

ptbrowne: Ah yes, my bad, thanks.

app/configurator/components/chart-controls/color-ramp.tsx
bprusinowski: Hmmm, `field` and `path` are raw strings, so there are some TS conflicts. I could change this so the `field` and `path` would only take pre-defined values (e.g. "areaLayer" and "palette"), but right now we only use this in maps, so I think there also would need to be some kind of check for the chart type inside a `ColorRampField` to make TS happy and then it would not be so generic. I am not sure how to handle this properly - do you have any idea?

app/configurator/components/chart-controls/color-ramp.tsx
bprusinowski: Yeah, but I needed an access to both `divergingPalettes` and `sequentialPalettes` so I didn't merge them into one array beforehand (e.g. in `ui-helpers`). Maybe it would make sense to have a {"diverging": [...], "sequential": [...]}, but still I would need to construct one array out of these two arrays.

app/configurator/components/chart-controls/color-ramp.tsx
bprusinowski: Hmmm, I based this on styles from `ColorPalette`, I guess it's common for custom select elements like these in color palettes. You mean extracting the whole CSS as a button variant?

app/charts/map/map-state.tsx
bprusinowski: Yeah, I'm not sure if it's more performant, but for sure it's nice :)

app/configurator/map/map-chart-options.tsx
bprusinowski: Good point, we don't have to :)

app/configurator/components/chart-controls/color-ramp.tsx
ptbrowne: Yes. It is not particularly problematic if it is really needed, but I think that if we have the concept of "custom buttons for color/palette pickers" maybe it should be a button variant, it would be clearer.

app/configurator/components/chart-controls/color-ramp.tsx
ptbrowne: yes I think having const allColorPalettes = [...divergingPalettes, ...sequentialPalettes] in ui-helpers would be good. Not super critical...

app/configurator/components/chart-controls/color-ramp.tsx
ptbrowne: I think having more narrow types in COlorRampField props and having typeguards in the renderer of ColorRampField so that we make sure that a ColorRampField is not rendered with the wrong types would be good.

ptbrowne: const ColorRampFieldProps = {
field: MapField,
path: MapPath
}

const Panel = () => {
  return <>
    { isMapField(field) && isMapPath(path) && <ColorFieldRamp field={field} path={path} />  
</>
}
```

something like that ?

app/configurator/components/chart-controls/color-ramp.tsx
bprusinowski: Done!

app/configurator/components/chart-controls/color-ramp.tsx
bprusinowski: Yes, but this would limit the ramp picker only to the maps, and maybe it could be also used in other places? But for now I can limit it to maps and we can change this later if needed 👍 Anyway, color types for other charts aren't there yet, so there is still some work to do related to colors :D

app/configurator/components/chart-controls/color-ramp.tsx
bprusinowski: I changed the logic a little bit in following commits; I think it would make sense to refactor the colors further in the future and handle types better anyway 😄

PR #297, opened 2022-01-27
feat: Remove homepage banner and icons

### LINKED ISSUE #295: Remove Test Banner and Icons in the background on the landing page
- Remove the red banner and the little symbols on the home page. (Do not remove the red banner for drafts).
(source: Email from Véronique on 26.01.2022) 
- Replace the "no data" message when the filter combination does not give any results with a message making clear that the user has to select another Filter Combination. But I think this will be solved anyway (or in most of cases?) with the new functionality on filter selection; the message should not be "no data", but rather "for this filter no data available" or similar:
(source: Email from Véronique on 26.01.2022 + Email from Sabine on 26.01.2022)

bprusinowski:
Closes #295.


ptbrowne (approved):
LGTM

PR #300, opened 2022-01-27
feat: Disable diverging palettes for maps

bprusinowski:
As discussed [here](https://github.com/visualize-admin/visualization-tool/issues/275#issuecomment-1023038300), I think it makes sense to temporarily disable divergent color palettes for maps for the time being, until clarified with Zazuko.

### DIFF
--- a/app/configurator/components/chart-controls/color-ramp.tsx
+++ b/app/configurator/components/chart-controls/color-ramp.tsx
@@ -10,7 +10,8 @@ import {
 } from "../..";
 import { Label } from "../../../components/form";
 import { Icon } from "../../../icons";
-import { divergingPalettes, sequentialPalettes } from "../ui-helpers";
+import { sequentialPalettes } from "../ui-helpers";
+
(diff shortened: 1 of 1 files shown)


ptbrowne:
LGTM !

@l00mi Do you have any plans to put steps inside a dimension metadata (for example a median of the values) so that we can add diverging palettes ? For example for a column showing money earned/lossed, the step would be 0 and we would be able to use a diverging palette centered around 0.

ptbrowne (approved):

l00mi:
> @l00mi Do you have any plans to put steps inside a dimension metadata (for example a median of the values) so that we can add diverging palettes ? For example for a column showing money earned/lossed, the step would be 0 and we would be able to use a diverging palette centered around 0.

That would be an annotation per MeasurmentDimension, right? I added this remark here: https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/286

PR #302, opened 2022-02-01
Polishing the maps

bprusinowski:
- changed base layer active tab switch to `OnOffControlTabField`
- updated translations for maps
- added `urql` devtools for non-prod environments
- fixed duplicated fetching of observations due to mismatch in input variables across different files for maps
- added separate queries to fetch `geoCoordinates` and `geoShapes`, so they aren't fetched with each filter change
- made the dimension values fetching only applicable for actual filter reorder so results are cached


app/configurator/components/chart-configurator.tsx
```diff
@@ -135,7 +135,7 @@ export const ChartConfigurator = ({
       // are the same  while the order of the keys has changed.
       // If this is not present, we'll have outdated dimension
       // values after we change the filter order
-      date: new Date(),
+      filterKeys: Object.keys(state.chartConfig.filters).join(", "),
```
ptbrowne, line 138: Nice 🎉

app/charts/map/chart-map.tsx
```diff
@@ -55,7 +55,10 @@ export const ChartMapVisualization = ({
     variables: {
       locale,
       iri: dataSetIri,
-      measures: [areaDimensionIri, symbolDimensionIri],
+      measures: [
+        chartConfig.fields.areaLayer.measureIri,
+        chartConfig.fields.symbolLayer.measureIri,
+      ],
```
ptbrowne, line 62: What was the problem here ?

app/configurator/interactive-filters/interactive-filters-configurator.tsx
```diff
@@ -134,22 +130,13 @@ const InteractiveFilterTabField = ({
...
-      </ControlTabButton>
-    </Box>
+    <OnOffControlTab
+      value={`${value}`}
+      label={label}
+      icon={icon}
+      checked={checked}
+      active={active}
+      onClick={onClick}
+    />
```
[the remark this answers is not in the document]
bprusinowski: Yes, good point. I am also planning on updating our internal docs on maps in general today.

app/charts/map/chart-map.tsx
bprusinowski: Previously there were component iris (geo dimensions) and not measure dimensions in this array. Probably `measures` could be blank, as they are not used in this place, but in order to make advantage of `urql` caching, I modified this to match variables from `data-download` and `chart-options-selector` to reduce the number of requests.

bprusinowski: Hmm... After going through resolvers, it seems that `measures` prop is not even used :D

https://github.com/visualize-admin/visualization-tool/blob/main/app/graphql/resolvers.ts

Maybe it could be good to remove them altogether and clean up the code?

ptbrowne: Mmh it seems used here no https://github.com/visualize-admin/visualization-tool/blob/main/app/graphql/resolvers.ts#L184 ?

bprusinowski: Ah, I mean the `measures` prop in the observations query:

https://github.com/visualize-admin/visualization-tool/blob/main/app/graphql/queries/data-cubes.graphql

bprusinowski: It's not used in the resolver

bprusinowski: Sorry, I see that references to code lines weren't pasted in the links I sent...

bprusinowski: (updated the links) 😆

app/configurator/components/field.tsx
```diff
@@ -229,14 +229,14 @@ export const DataFilterSelectDay = ({
...
       value={dateValue}
+      isDayDisabled={isDisabled}
       inputProps={{
         id,
         disabled,
       }}
       dayPickerProps={{
         fromMonth,
         toMonth,
-        disabledDays: isDisabled,
```
ptbrowne, line 208: I do not even understand why the day picker allows clicking on disabled days :/

app/configurator/components/chart-options-selector.tsx
```diff
@@ -178,19 +178,12 @@ const EncodingOptionsPanel = ({
...
-    symbolLayer: t({
-      id: "controls.select.measure",
-      message: "Select a measure",
-    }),
-    segment: t({
-      id: "controls.select.dimension",
-      message: "Select a dimension",
-    }),
+    // FIXME: encoding types, so we don't need these there (chart options are
+    // handled in a separate file)
```
ptbrowne, line 182: Should the FIXME be dealt with in this PR ?

app/locales/fr/messages.po
```diff
@@ -302,6 +302,10 @@ msgstr "Réinitialiser la palette de couleurs"
 msgid "controls.color.palette.sequential"
 msgstr "Séquentielle"
 
+#: app/configurator/map/map-chart-options.tsx:437
+msgid "controls.color.select"
+msgstr "Couleur sélectionner"
```
ptbrowne, line 307: Sélectionner une couleur

app/docs/controls.docs.tsx
```diff
@@ -57,6 +58,22 @@ const ControlsDoc = () => {
...
+  ${(
+    <ReactSpecimen span={2}>
+      <OnOffControlTab
+        value="Test"
+        label={<span>Test</span>}
+        icon="settings"
+        onClick={() => {}}
+      />
+    </ReactSpecimen>
+  )}
```
ptbrowne, line 75: Thanks !

ptbrowne (approved):
LGTM, there is only the comment about the french translation for Select a color. Good job on the perf improvements !

app/charts/map/chart-map.tsx
ptbrowne: Good point ! I think we can remove it.

app/locales/fr/messages.po
bprusinowski: 🇫🇷

app/configurator/components/field.tsx
ptbrowne: Thanks for the fix.

app/locales/it/messages.po
```diff
@@ -302,6 +302,10 @@ msgstr "Ripristina la tavolozza dei colori"
 msgid "controls.color.palette.sequential"
 msgstr "Sequenziale"
 
+#: app/configurator/map/map-chart-options.tsx:437
+msgid "controls.color.select"
+msgstr ""
```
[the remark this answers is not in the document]
bprusinowski: Whoops, good catch, thanks!

app/charts/map/chart-map.tsx
ptbrowne: Not in this PR.

app/docs/controls.docs.tsx
bprusinowski: I think that we should refactor the docs to make them more consistent... Also `Dataset Browsing` docs do not work, but I think it would make sense to tackle this + adding more docs & improving current ones as a future ~ 1-day task? :)

app/configurator/components/field.tsx
bprusinowski: Np!

app/docs/controls.docs.tsx
ptbrowne: Good point: https://github.com/visualize-admin/visualization-tool/issues/306

app/configurator/components/chart-options-selector.tsx
bprusinowski: I would prefer to handle this in another PR by the end of this week 🤞

Issue #288, opened 2022-01-25
Link to unversioned entry of Cube in Visualize does not work anymore.


l00mi:
**Describe the bug**
At some point in time we had the agreement that we can link to a unversioned cube like here https://opendata.swiss/de/dataset/staatsrechnungen-funktion

e.g. https://visualize.admin.ch/en/browse/dataset/https%3A%2F%2Fculture.ld.admin.ch%2Fsfa%2FStateAccounts_Function

This worked at some point, currently it seems to needs the additiona `/3/` in the link above.

**To Reproduce**
Click 2nd link above.

**Environment (please complete the following information):**
 - Visualize environment and version: PROD

l00mi:
Any updates on this. That is a bug in production. The opendata.swiss entries do currently not work.

ptbrowne:
I will work on this this afternoon. Can you provide more context on the opendata.swiss entries ? I thought it was only a bug between cube view and visualisation.admin ?

ptbrowne:
OK I saw the link on opendata.swiss, I did not know at all that there was this contract between opendata.swiss and visualize.admin, thanks for the report.

Rdataflow:
@ptbrowne could you investigate:
- https://test.visualize.admin.ch/en/browse/dataset/https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Ftest_ordinal_507 doesn't work 
- https://test.visualize.admin.ch/en/browse/dataset/https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Ftest_ordinal_507%2F5 OTOH does.

Expected would be a similar behaviour like like https://visualize.admin.ch/en/browse/dataset/https%3A%2F%2Fculture.ld.admin.ch%2Fsfa%2FStateAccounts_Function where URL is forwarded to the versioned cube, right?

Rdataflow:
@ptbrowne could you find the reason alreday? vero is also concerned by this issue...

cc @ortnever

Rdataflow:
here the case for vero https://test.visualize.admin.ch/en/browse/dataset/https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fubd000501

Rdataflow:
@ptbrowne LGTM perfect 👍

PR #303, opened 2022-02-01
feat: Redirect to latest cube iri when going to on browse/dataset/<unversionedIri>

ptbrowne:
Fixes https://github.com/visualize-admin/visualization-tool/issues/288

When we land on /en/browse/dataset/<unversionedIri>, we fetch the latest published cube to get the latest unversioned iri and redirect server side to /en/browse/dataset/<versionedIri>.
To detect unversionedIri, a heuristic is used, it should end with `/<number>/`.

### DIFF
--- a/app/rdf/query-cube-metadata.ts
+++ b/app/rdf/query-cube-metadata.ts
@@ -5,6 +5,7 @@ import { schema, dcat, dcterms } from "../../app/rdf/namespace";
 import { DataCubeOrganization, DataCubeTheme } from "../graphql/query-hooks";
 import { keyBy } from "lodash";
 import { makeLocalesFilter } from "./query-labels";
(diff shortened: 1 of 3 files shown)


app/rdf/query-cube-metadata.ts
```diff
@@ -172,6 +173,27 @@ export const queryDatasetCountBySubTheme = async ({
     .filter((r) => r.iri);
 };
 
+export const queryLatestPublishedCubeFromUnversionedIri = async (
+  unversionedIri: string
+) => {
+  const iri = RDF.variable("iri");
```
bprusinowski, line 179: +1

bprusinowski (approved):
LGTM 👍

PR #309, opened 2022-02-01
fix: Discrete color scales with < 3 values + display lakes and rivers

bprusinowski:
- _Fixes_ discrete color scales with < 3 observations for maps
- Adds new MVTLayer which displays lakes and rivers
- Fixes shape fill color when value === 0


bprusinowski:
Discrete color scales for maps still need a proper fix, but at least they will now work when number of observations is lower than 3.

PR #314, opened 2022-02-02
fix: Ending slash is optional for iris (versioned or unversioned)

### LINKED ISSUE #313: selecting lfi data cubes jumps back to home page
**Describe the bug**
Selecting one of the lfi datasets jumps back to https://test.visualize.admin.ch/

**To Reproduce**
1. Go to https://test.visualize.admin.ch/browse?includeDrafts=true&order=SCORE&search=lfi
2. Click on a LFI dataset

**Expected behavior**
Jump to visualize this chose dataset

Visualize environment and version: TEST v3.2.7 (ba4b7c5)

ptbrowne:
Iris for dataset can contain an optional ending slash.

Since the regex would match only URLs with ending slash, we would consider some URLs as unversioned when in fact they are versioned, and then we would try to find the versioned dataset and it would fail. Now we correctly consider http://<iri>/1 as versioned (and also http://<iri>/1/).

fix #313

### DIFF
--- a/app/pages/browse/[type]/[iri].tsx
+++ b/app/pages/browse/[type]/[iri].tsx
@@ -9,7 +9,7 @@ export default GenericBrowse;
  * Versioned iris look like https://blabla/<number/
  */
 const isDatasetIriVersioned = (iri: string) => {
-  return iri.match(/\/\d+\/$/) !== null;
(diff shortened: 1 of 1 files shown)


bprusinowski (approved):
LGTM :)

Issue #316, opened 2022-02-02
Date filter: slider blocked


MichelaBurzio:
**Describe the bug**
Once the color segmentation is selected, the slider to change the date range (x-axis) is locked.

**To Reproduce**
1. Create a visualization starting from 'Bathing water quality' dataset
2. Select bar chart 
3. Click Next button
4. Click on 'Color' segmentation and select 'Bathing site'
5. Click on 'Date' Horizontal axis
6. See the blocked slider 
<img width="1728" alt="Screenshot 2022-02-02 at 17 01 42" src="https://user-images.githubusercontent.com/98882778/152190627-81bd2fca-40de-4d40-83b6-5c7c64c97357.png">

**Environment (please complete the following information):**
 - Visualize environment and version: TEST v3.2.7
 - Browser and version: Chrome 97

MichelaBurzio:
Thank you @ptbrowne. I tested and it is working well now!

PR #321, opened 2022-02-03
fix: Ability to use date filter for time dimension in days

ptbrowne:
Previously, only the min/max were returned in the values. Now that
values are returned (necessary for date picker), we need to use the
last values and not the 2nd one for the max value.

fix https://github.com/visualize-admin/visualization-tool/issues/316


bprusinowski (approved):
LGTM 👍

PR #322, opened 2022-02-03
fix: Show dimension values even when column datatype is rdf:langString

ptbrowne:
Some datasets have been created with rdf:langString in some columns. This means that strings are tagged with their language. Since we do not keep the language information inside the filter, we have to filter only based on the string representation of the value.


bprusinowski (approved):
LGTM :)

PR #339, opened 2022-02-10
Update messages.po

AnninaWalker:
Change some button labels


app/locales/en/messages.po
```diff
@@ -59,7 +59,7 @@ msgstr "Categories"
 
 #: app/configurator/components/dataset-preview.tsx:60
 msgid "browse.dataset.create-visualization"
-msgstr "Create visualization"
+msgstr "Start a visualization"
```
ptbrowne, line 62: I guess other languages should be change right ?

"Démarrer une visualisation" would be the French translation

ptbrowne: Saw the other PRs.

PR #343, opened 2022-02-11
Improve filter value selection

### LINKED ISSUE #327: Wrong initial filter selection
**Description**
For at least one dataset, there is a "No data available for this current filter selection" displayed during a chart type selection step; however when going to a next step, the data is correctly fetched.

**To Reproduce**
1. Go to https://test.visualize.admin.ch/en/browse/dataset/https%3A%2F%2Fenergy.ld.admin.ch%2Fsfoe%2Fbfe_ogd56_energieperspektiven2050%2F4?previous=%7B%22includeDrafts%22%3Afalse%7D.
2. Click on "Create visualization".
3. Wait for the error to appear.
4. Click "Next".
5. See the data being loaded.

**Expected Behavior**
The data should be loaded in a first step in way it does in the second step.

**Environment**
 - Visualize: TEST v3.2.7 (d1d94f5)
 - Browser: Safari 14.1.1 (16611.2.7.1.4)

### LINKED ISSUE #328: Cascading Filter does not work with "Swiss PRTR (Test03)"
@ptbrowne : il s'agit du bug communiqué par Email le 28.01.

Filtre en cascade ne fonctionne pas pour PRTR (Test03)"

1.	Rechercher PRTR via test.visualize, y compris Draft
2.	Ouvrir « Registre suisse des rejets et transferts de polluants Swiss PRTR (Test03) »
3.	Choisir Bar chart (pas de données, pour l’ordre des filtres affichés par défaut)
4.	Choisir Etablissement comme segmentation (couleurs) et Modifier l’ordre des filtres comme printscreen ci-dessous
5.	Un graphique s’affiche par défaut
6.	Modifier le polluant et choisir "Chloride"
7.	Pas de données, alors qu’il en existe si on choisit comme activité « nicht aufgeschlüsselt ».

Expected behavior
la valeur "nicht aufgeschlüsselt" devrait être sélectionnée automatiquement dans le filtre "Activité principale" et des données seraient ainsi disponibles

image

image

image

ptbrowne:
Previously when changing a filter in a cascading facets. we would take the first value of each dimension value that was present and this could lead to dead-end, and was slow since each filter could be changed after the filter just before it has been changed. Now the approach is different: when a filter is changed/moved, we try to find an observation with the filter, and if we do not find an observation, we relax the filters 1 by 1 starting from the bottom until we have an observations. This ensures that data is shown after filter have been loaded, and it also tries to keep as much as possible what the user has chosen (= change the bottomost filters first).

fix #327
fix #328

### DIFF
--- a/app/configurator/components/chart-configurator.tsx
+++ b/app/configurator/components/chart-configurator.tsx
@@ -119,6 +119,52 @@ const DataFilterSelectGeneric = ({
   );
 };
 
+const useEnsurePossibleFilters = ({
+  state,
+}: {
+  state: ConfiguratorStateConfiguringChart;
+}) => {
+  const [, dispatch] = useConfiguratorState();
+
+  const client = useClient();
+
+  useEffect(() => {
+    const run = async () => {
+      const {
+        data,
+        error,
+      }: { data?: PossibleFiltersQuery; error?: CombinedError } = await client
+        .query(PossibleFiltersDocument, {
+          iri: state.dataSet,
+          filters: state.chartConfig.filters,
+        })
+        .toPromise();
+      if (error || !data) {
+        console.warn("Could not fetch possible filters", error);
+        return;
+      }
+
+      const filters = Object.fromEntries(
+        data.possibleFilters.map((x) => [
+          x.iri,
+          { type: x.type, value: x.value },
+        ])
+      ) as ChartConfig["filters"];
+
+      if (!isEqual(filters, state.chartConfig.filters)) {
+        dispatch({
+          type: "CHART_CONFIG_FILTERS_UPDATE",
+          value: {
+            filters,
+          },
+        });
+      }
+    };
+
+    run();
+  }, [client, dispatch, state.chartConfig, state.dataSet]);
+};
+
 export const ChartConfigurator = ({
   state,
 }: {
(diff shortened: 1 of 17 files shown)


app/configurator/components/chart-configurator.tsx
```diff
@@ -177,23 +177,52 @@ export const ChartConfigurator = ({
...
+    const run = async () => {
+      const {
+        data,
+        error,
+      }: { data?: PossibleFiltersQuery; error?: CombinedError } = await client
+        .query(PossibleFiltersDocument, {
+          iri: state.dataSet,
+          filters: state.chartConfig.filters,
+        })
+        .toPromise();
```
bprusinowski, line 196: I think that we use a `client` directly only in RDF queries and useXYZQuery elsewhere (`usePossibleFiltersQuery` hook in this case). Not sure why this pattern has changed here? Would the results still be cached?

app/rdf/queries.ts
```diff
@@ -395,11 +395,16 @@ export const getCubeObservations = async ({
...
   limit,
+  raw,
 }: {
   cube: Cube;
   locale: string;
+  /** Observations filters that should be considered */
   filters?: Filters;
+  /** Limit on the number of observations returned */
   limit?: number;
+  /** Returns URIs instead of labels for NamedNodes  */
```
bprusinowski, line 406: IRIs?

bprusinowski (approved):
LGTM, had some minor comments 👍

app/configurator/components/chart-configurator.tsx
ptbrowne: Mmh I think using useQuery here is not the clearer thing since what we want is to query the possible filters on filters change, and then possibly change the configuration. I tried just now to use usePossibleFiltersQuery + useEffect and it did not work well as the configuration was reset to what was there before as soon as I make a change (since the query is not yet changed, but configuration has changed, the effect is run, and the filters are reset according to the *previous* filters that have been chosen.

I think the imperative logic here is easier to follow.

ptbrowne: And for the caching, it works according to documentation

> By default, urql uses a concept called Document Caching. It will avoid sending the same requests to a GraphQL API repeatedly by caching the result of each query.

bprusinowski: 👍

app/graphql/query-hooks.ts
```diff
@@ -16,6 +16,7 @@ export type Scalars = {
   Filters: any;
   GeoShapes: any;
   Observation: any;
+  ObservationValue: any;
```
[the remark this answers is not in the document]
ptbrowne: You're right, I think I made several attempts to have good types, and then I settled with a scalar, and now it is the same as DImensionValue. I've removed ObservationValue, in favor of DimensionValue.

PR #352, opened 2022-02-16
Error bars 1

ptbrowne:
In this PR, standard errors support is added to visualize.

- [x] Display error whiskers automatically on simple column chart

Other PRs will follow to refine the functionality (tooltip support, activate / deactivate, support in horizontal chart).

Related meta issue : https://github.com/visualize-admin/visualization-tool/issues/353

The dataset where the standard error dimension is available is "LFI Data Cube [Version 1]: Waldfläche / Produktionsregion"

### DIFF
--- a/app/charts/column/columns-state.tsx
+++ b/app/charts/column/columns-state.tsx
@@ -59,6 +59,7 @@ export interface ColumnsState {
   xEntireScale: ScaleTime<number, number>;
   xScaleInteraction: ScaleBand<string>;
   getY: (d: Observation) => number | null;
+  getYError: null | ((d: Observation) => [number, number]);
   yScale: ScaleLinear<number, number>;
   getSegment: (d: Observation) => string;
   segments: string[];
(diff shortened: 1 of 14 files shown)


app/charts/column/columns-state.tsx
```diff
@@ -102,7 +103,22 @@ const useColumnsState = ({
...
+    ? (d: Observation) => {
+        const y = getY(d) as number;
+        const error =
+          d[errorIri] !== null ? parseFloat(d[errorIri] as string) : null;
+        return (error === null ? [y, y] : [y - error, y + error]) as [
+          number,
+          number
+        ];
+      }
+    : null;
```
bprusinowski, line 128: I would expect to get just a yError when I call `getYError`, not yValue +- yError; maybe the function could be called `getYAdjustedByError`, or similarly?

bprusinowski (approved):
LGTM, good work! I know this is just a first part of error bars, but there is one thing I would also consider adding – making the scales aware of whiskers, so they won't be able to go outside the SVG.

app/charts/column/columns-state.tsx
ptbrowne: getYErrorRange ?

Good point about the scales, I will do it !

bprusinowski: I think `getYErrorRange` it's also ok :)

ptbrowne: OK thanks, I'll do that in the new PR.

PR #354, opened 2022-02-16
Initial zoom to bounding box of visible shapes / symbols for maps

### LINKED ISSUE #349: Zoom to a bounding box of visible shapes
The initial zoom of the map should be dynamic and adjust to currently displayed shapes on initial load.

bprusinowski:
Closes #349.

This PR introduces an initial zoom to a bounding box of visible shapes / symbols. It only happens `onLoad` event of DeckGL (so e.g. when a user opens a published chart or editor goes to chart options / changes locale, etc.).


app/charts/map/map.tsx
```diff
@@ -116,6 +117,38 @@ export const MapComponent = () => {
...
+  }, [viewState, features.symbolLayer?.points]);
+
+  const initialZoom = useCallback(() => {
+    if (areaLayer.show) {
+      zoomToShapes();
+    } else if (symbolLayer.show) {
+      zoomToSymbols();
+    }
+  }, [areaLayer.show, symbolLayer.show, zoomToShapes, zoomToSymbols]);
+
```
bprusinowski, line 151: I am not proud of this solution, as this function only gets called once, when DeckGL instance is loaded; but there was a need to access `ViewState`, as the first event that is fired is `onResize` event, which updates the state so we can't use `INITIAL_VIEW_STATE`...

`onLoad` doesn't provide an access to viewState either – let me know if you have an idea to improve 🙏

ptbrowne: Both area layers and symbol layers can be shown right ? So in this case, we should include in the bounding box both types right ? Maybe onLoad can be used just to track loading status of the map and then zoom can be managed in an effect ? To ensure calling only once, we can use a ref.

```
const hasZoomedRef = useRef()
const setInitialZoom = useCallback(() => {
  if (hasZoomedRef.current) { return }
  const bb = makeBoundingBox(showSymbolLayer ? symbolLayer : undefined, showAreaLayer ? areaLayer : undefined)
 zoomToBoundingBox(bb)
 hasZoomedRef.current = true
}, [mapLayer, symbolLayer, areaLayer])

useEffect(() => {
  if (!mapLoaded || !symbolLayerLoaded || !areaLayerLoaded) {
    return
  }
  setInitialZoom()
}, [mapLoaded, symbolLayerLoaded, areaLayerLoaded])

...
<StaticMap
  onLoad={() => setMapLoaded(true)}
/>
```

ptbrowne: Ah I forgot view state in my code but it could easily be integrated in the effect.

app/charts/map/map.tsx
```diff
@@ -117,37 +128,87 @@ export const MapComponent = () => {
...
+          return bbox;
+        } else {
+          return shapesBbox;
+        }
+      } else {
+        return symbolsBbox;
+      }
+    },
+    []
+  );
```
ptbrowne, line 180: Can be extracted out of the component if there are no dependencies.

ptbrowne (approved):
LGTM 👍  Only the extraction of the bounding box out of the component ;)

app/charts/map/map.tsx
bprusinowski: Right, will do!

PR #356, opened 2022-02-16
Error bars 2

ptbrowne:
- [x] Show error range in tooltip
- [x] Show error whiskers in grouped columns 

Preview here : https://visualize-ad-error-bars-xqzebf.herokuapp.com/en/create/new?from=kbIrSDxJKLqL


bprusinowski (approved):
LGTM 👍

PR #359, opened 2022-02-17
fix: Color segmentation filtering

### LINKED ISSUE #355: Weird behaviour when using filters in the right panel
**Describe the bug**
Something weird is happening when trying to use "Select none" in filtering options in the right panel. A lot of data is showing on the chart (probably with some duplicates) and options are being "re-selected", so it's not actually possible to select none options. Also, checkboxes are not being "checked" when trying to select a single value.

**To Reproduce**
1. Load Bathing water quality dataset.
2. Select a bar chart.
3. Add coloring by Parameter.
4. Click on "Select none" down below in the Filter panel.
5. Click on "E.coli".

**Expected behavior**
I think it should be possible to select none and not display anything; selecting particular, single options should also work.

**Screenshots**

In the below screenshot, I firstly clicked on "Select none" (then checkboxes were re-selected) and then on "E.coli" – both checkboxes were de-selected, but the data filtered by "Enterokokken". It also seems that the data gets duplicated (see tooltip).

Screenshot 2022-02-16 at 13 12 44

 - Visualize environment and version: INT v3.3.0
 - Browser and version: Safari 14.1.1 (16611.2.7.1.4)

ptbrowne:
With the change of loading possible filters, a mistake was made since
the "right" filters (= filters that are on an IRI that is mapped)
should not be considered in the query that tries to find possible values for left filters.

- [x] Remove right filters from the possible filters query
- [x] Do not re-run possible filters query when changing right filters
- [x] Keep right filters when possible filters query result is set as filters 

Tested on multiple datasets:

- [x] Red list
- [x] Bathing water quality
- [x] Heavy metal

Also, a change was made so that GeoShapesDimension and GeoCoordinatesDimension can be filtered correctly in table view.

fix #355

### DIFF
--- a/app/charts/shared/use-chart-state.tsx
+++ b/app/charts/shared/use-chart-state.tsx
@@ -2,6 +2,7 @@ import { createContext, useContext } from "react";
 import { ChartFields, InteractiveFiltersConfig } from "../../configurator";
 import { Observation } from "../../domain/data";
 import { DimensionMetaDataFragment } from "../../graphql/query-hooks";
+import { Has } from "../../lib/has";
 import { AreasState } from "../area/areas-state";
 import { GroupedBarsState } from "../bar/bars-grouped-state";
 import { BarsState } from "../bar/bars-state";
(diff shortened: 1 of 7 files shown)


bprusinowski (approved):
LGTM!

PR #361, opened 2022-02-17
Map style updates

### LINKED ISSUE #360: Remove (i) Icon that appears when editing / creating map visualizations
**Describe the bug**
When creating visualizations with the chart type: maps,  an unexpected (i) icon appears in the bottom right-hand corner below the zoom toggles.

**To Reproduce**
1. Go to test.visualize.admin.ch
2. Select "Bathing Water Quality Dataset"
3. Select Chart Type: Map
4. While editing the map an unexpected (i) icon appears in the bottom right-hand corner below the zoom toggles
Screenshot 2022-02-17 at 10 44 56

**Expected behavior**
The (i) should be removed if no additional information is provided to the user. Alternatively, "Swisstopo" should be shown, analog the published chart. See "Additional Context" below.

**Links**
Example of a published chart: https://test.visualize.admin.ch/en/v/VWCtv10OBxBz

 - Visualize environment and version: Test v3.3.0
 - Chrome

**Additional context**
On the published chart the (i) is not there, but rather "Swisstopo" is shown - which I believe is correct as they are the source of the information / data. However, I find the black arrow to be confusing. Maybe we can remove it?
Screenshot 2022-02-17 at 11 24 16

bprusinowski:
Closes #360.

- [x] change country label color from pink to black
- [x] remove swisstopo label at the bottom of the chart

### DIFF
--- a/app/charts/map/map.tsx
+++ b/app/charts/map/map.tsx
@@ -175,7 +175,9 @@ export const MapComponent = () => {
         controller={{ type: MapController }}
         getCursor={() => "default"}
       >
-        {showBaseLayer && <StaticMap mapStyle={baseLayerStyle} />}
+        {showBaseLayer && (
+          <StaticMap mapStyle={baseLayerStyle} attributionControl={false} />
+        )}
 
         {areaLayer.show && (
           <>
--- a/app/charts/map/get-base-layer-style.ts
+++ b/app/charts/map/get-base-layer-style.ts
@@ -4,6 +4,7 @@ const BORDER_COLOR = {
   lowZoom: "hsl(0, 0%, 50%)",
   highZoom: "hsl(0, 0%, 75%)",
 };
+const COUNTRY_LABEL_COLOR = "rgba(0, 0, 0, 0.9)";
 
 interface Props {
   locale: Locale;
@@ -5883,7 +5884,7 @@ export const getBaseLayerStyle = (props: Props) => {
           visibility: "visible",
         },
         paint: {
-          "text-color": "rgba(145, 70, 145, 0.9)",
+          "text-color": COUNTRY_LABEL_COLOR,
           "text-halo-blur": 1,
           "text-halo-color": "rgba(255, 255, 255, 0.8)",
           "text-halo-width": 1,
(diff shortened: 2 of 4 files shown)


app/charts/map/map.tsx
```diff
@@ -219,7 +219,12 @@ export const MapComponent = () => {
               controller={{ type: MapController }}
               getCursor={() => "default"}
             >
-              {showBaseLayer && <StaticMap mapStyle={baseLayerStyle} />}
+              {showBaseLayer && (
+                <StaticMap
+                  mapStyle={baseLayerStyle}
+                  attributionControl={false}
```
bprusinowski, line 225: I made an assumption that we don't have to display Swisstopo's logo at the bottom of a map, at least this is how it was in a prototype. I think it makes sense to ask this question during a meeting with Swisstopo, just to be sure.

ptbrowne: > Terms of use: the map combines different data sets with different terms of use. The map and the geoservice that provides it may be used in accordance with the [General Terms of Use and Operating Conditions of the Federal Spatial Data Infrastructure FSDI](http://www.geo.admin.ch/terms-of-use). In the case of digital or analog representations and publications by the administrative units of the Federal Administration, the following source reference must be provided in all cases: (with appropriate links) [© swisstopo](https://www.swisstopo.ch/),[© MapTiler](https://www.maptiler.com/copyright/) [© OpenStreetMap contributors](https://www.openstreetmap.org/copyright)

https://www.geo.admin.ch/en/geo-services/geo-services/portrayal-services-web-mapping/vector_tiles_service.html

I think here the problem is that the attribution control is not displayed properly. I think we are lacking CSS classes for attribution elements like `.mapboxgl-ctrl mapboxgl-ctrl-attrib .mapboxgl-compact` as I see that in the above documentation, the "i" is displayed properly (and on hover it shows swisstopo). I think we may have to copy some CSS classes into our global CSS.

ptbrowne (approved):

app/charts/map/map.tsx
ptbrowne: Let's tackle this in another issue right now, disabling attribution for now, issue created #369

PR #364, opened 2022-02-17
Use booleans in map rendering condition

bprusinowski:
Fixes #363.

Changes the logic so the booleans are used for map rendering condition.

PR #367, opened 2022-02-17
Add shapes & points filtering to maps

### LINKED ISSUE #342: Maps: Choose which cantons can be displayed
**Is your feature request related to a problem? **
Input from Michael. Just by chance I discovered one missing feature which I would expect based on the other types of visualzation:
For the Areas Layer and also Symbols Layer (e.g. Canton) I would expect that we can choose which of the cantons are shown. As it is possible in all other visualization types. This will be very usefull once we have more datasets on municipality level, combined with the zoom feature.

Being able to choose which cantons should be displayed by adding a filter section (as we have it in all other charts) to the Area and the Symbol Layer.

bprusinowski:
Closes #342.

This PR adds an ability to filter geoDimension in the right panel. There was a need for some custom logic for shapes, as they are fetched separately from dimension's values.

- [ ] should symbol layer based on the area layer be filterable separately? In such case filtering applies to both areas and symbols


app/charts/map/map-state.tsx
```diff
@@ -173,13 +174,20 @@ const useMapState = ({
...
         const hierarchyLabels = (
           (geoShapes as any).topology.objects.shapes.geometries as GeoFeature[]
         )
           .filter((d) => d.properties.hierarchyLevel === hierarchyLevel)
           .map((d) => d.properties.label);
 
-        return data.filter((d) => hierarchyLabels.includes(getLabel(d)));
+        return data
+          .filter((d) => dimensionLabels.includes(getLabel(d)))
+          .filter((d) => hierarchyLabels.includes(getLabel(d)));
```
ptbrowne: It would be better to have an index to check for inclusion here as otherwise we are looping in N^2.

```
const index = Object.fromEntries([...dimensionLabel, ...hierarchyLabels].map(l => [l, true]))
return data.filter(d => !!index[getLabel(d)])
```

ptbrowne (commented):
I think the filtering could be improved so that we are not in N^2. Otherwise LGTM, very cool 👍

app/charts/map/map-state.tsx
bprusinowski: I have modified the code, so it's no longer needed 🥳

ptbrowne:
When I go there : https://visualize-ad-feat-add-s-8usxwv.herokuapp.com/en/browse/dataset/https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2FCOVID19VaccPersons_sex_w%2F1%2F?previous=%7B%22includeDrafts%22%3Atrue%2C%22order%22%3A%22SCORE%22%2C%22search%22%3A%22covid%22%7D and start a map viz, I have an unexpected error 🤔 Is this because of the dataset ?

bprusinowski:
It doesn't work because #364 is not merged yet 😅

bprusinowski:
I had made a quick fix locally to test it on my machine, but I would have to merge #364 to this PR to enable a correct deployment preview 👀

ptbrowne (approved):
LGTM ! Thanks :)

PR #384, opened 2022-02-18
Do not set automatically empty filters

### LINKED ISSUE #385: Filters are removed when a filter is changed and there is no data for the configuration
**Describe the bug**
The default filters ensure that data is loaded and a visualization is shown. However if the filters are changed, and there is no data for a specific configuration, the filters are removed. This is confusing because it is unclear if the error is caused by the application or the user. 

**To Reproduce**
1. Go to https://int.visualize.admin.ch/en/browse/dataset/https%3A%2F%2Fculture.ld.admin.ch%2Fsfa%2FStateAccounts_Category%2F5%2F?previous=%7B%22includeDrafts%22%3Afalse%2C%22order%22%3A%22SCORE%22%2C%22search%22%3A%22peaks%22%7D
2. Click on 'Start a visualization" -> Select areas chart type and then "next"
3. in the left sidebar change the filter from "Allgemeine Verwaltung" to "Ertrag der Liegenschaften und Kapitalien"
4. All filters are removed from the left side bar and the user has to add them again. 

**Expected behavior**
In this case I would expect to see an error message informing me that no data for the selected filter configuration is available

**Screenshots**
https://user-images.githubusercontent.com/29289679/154710882-b456f96d-95e9-4f60-a08c-ec11a4e760d5.mov

 - Visualize environment and version: INT
 - Browser and version Firefox, Edge
 
**Additional context**

ptbrowne:
This removes a patological case where filters are completely removed. In most cases we do not want filters to be completely removed automatically, it's better if "no data" is shown.
In the case where filters can be changed to show data they will be changed (starting from the bottom).

fix #385

### DIFF
--- a/app/configurator/components/chart-configurator.tsx
+++ b/app/configurator/components/chart-configurator.tsx
@@ -1,6 +1,6 @@
 import { t, Trans } from "@lingui/macro";
 import { Menu, MenuButton, MenuItem, MenuList } from "@reach/menu-button";
-import { isEqual, sortBy } from "lodash";
+import { isEmpty, isEqual, sortBy } from "lodash";
 import React, { useCallback, useEffect, useRef, useState } from "react";
 import {
   DragDropContext,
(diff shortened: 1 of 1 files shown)


bprusinowski (approved):
LGTM!

PR #387, opened 2022-02-21
Hide all elements related to labels, not only text

bprusinowski:
Changed visibility of text layers, so there are no "alone" points and rectangles without labels.


ptbrowne (approved):
Nice 🎉

PR #396, opened 2022-02-21
Fix search and drafts count

### LINKED ISSUE #391: Search options combinations seems buggy
**Describe the bug**
I report here a few things regarding the dataset search feature, the combination of the search options between theme/organisations and keywords:
1/ the number of dataset is updated under theme/organisations when a keyword is entered -> this seems to work
But this number should also be updated when just clicking on the "Include draft datasets" checkbox.
Clarification: when just clicking on the "include draft" checkbox, then the number of datasets is not updated, but it should

2/ when a keyword is entered, and then a theme/organisations is selected -> what is happening ?
Because the keyword seems automatically deleted, but it seems that it is still used to filter the search results.
The correct behavior should be: keep the keyword in the field, show only the datasets that correspond to the keyword+selected theme/organisation.
Complement: when doing it the other way round it seems to work properly (first select a theme/organisation, and then enter a keyword)

3/ Another special case where I do not understand what is happening:
1. Go to 'https://test.visualize.admin.ch/en'
2. Enter "list" in the search field, and select "include drafts"
3. The categories show "1" for "National Economy"
4. Click on "National economy" and then, the results are empty:
image

- Visualize environment and version: TEST v3.4.1 + https://visualize-ad-testingv3--dl6d2r.herokuapp.com/en
- Browser and version: Firefox 97.0.1

ptbrowne:
fix #391

- fix: Consider include drafts checkbox when computing search counts
- feat: Keep search and includeDrafts when clicking inside nav panel

### DIFF
--- a/app/components/search-autocomplete.tsx
+++ b/app/components/search-autocomplete.tsx
@@ -2,6 +2,7 @@ import React, { useMemo } from "react";
 import {
   BrowseFilter,
   DataCubeAbout,
+  useBrowseContext,
 } from "../configurator/components/dataset-browse";
 import { useOrganizationsQuery, useThemesQuery } from "../graphql/query-hooks";
 import SvgIcCategories from "../icons/components/IcCategories";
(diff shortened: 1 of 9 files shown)


bprusinowski (approved):
LGTM! 👍

PR #397, opened 2022-02-21
Map improvements

bprusinowski:
- [ ] add attribution to the map
- [ ] disable right panel filter on the map is layer is not active

### DIFF
--- a/app/charts/map/map-attribution.tsx
+++ b/app/charts/map/map-attribution.tsx
@@ -1,0 +1,57 @@
+import React, { useState } from "react";
+import { Box, Grid } from "theme-ui";
+import { Icon } from "../../icons";
+
+const ATTRIBUTION = "© Data: MapTiler, OpenStreetMap contributors, Swisstopo";
+const ICON_SIZE = 14;
+
+interface Props {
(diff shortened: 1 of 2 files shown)


app/charts/map/map-attribution.tsx
```diff
@@ -0,0 +1,57 @@
...
+      onMouseEnter={() => setIsExpanded(true)}
+      onMouseLeave={() => setIsExpanded(false)}
+    >
+      <Grid
+        sx={{
+          gridTemplateColumns: `1fr ${ICON_SIZE}px`,
+          gridGap: 1,
+          alignItems: "flex-end",
+        }}
+      >
```
ptbrowne, line 29: Interesting, I would have used Stack for this.

```
<Stack direction='row' spacing={1}
</Stack>
```

I think it is a bit simpler and you do not need the ICON_SIZE.

ptbrowne (approved):
Small nit on the implementation of the attribution box.

LGTM 👍  Thanks

ptbrowne (approved):

ptbrowne (approved):

Issue #380, opened 2022-02-18
Visualisation layer under labels / lakes


ptbrowne:
We have disabled labels on the map as they were under the viz layer. We should find a way to insert the viz layer just under the viz layer, see https://deck.gl/docs/api-reference/mapbox/overview#inserting-a-2d-deck-layer-before-an-existing-mapbox-layer.

- Labels over viz layer
- Lakes over viz layer

This could be done via different layer styles.

sosiology:
Notes from Weekly Call (06.04.2022)
- The shapes are stacked and you cannot see everything (e.g. in Tessin the shapes are over the lakes, whereas int he West of CH it is ok) - this is due to the lake layer being under the shapes.

ptbrowne:
Let's also check https://github.com/visualize-admin/visualization-tool/issues/479

ptbrowne:
Fixed by #514

PR #398, opened 2022-02-21
Labels on top of shapes (maps)

bprusinowski:
`_map.tsx` is the current map, saved for reference, so I could copy & paste the code easily.

There is a problem with hovering, as the circles below other circles are highlighted in the first order. I tried with a custom `Controller` (https://deck.gl/docs/api-reference/core/controller) that would use Mapbox GL's `queryRenderedFeatures`, but I wasn't sure about the performance and we would need to change the way of highlighting the circles (no autoHighlight, probably some `const activeCircle = useState()`).

Also, there is another problem with disabling a base layer, probably related with one of the contexts (map or deck). After it's disabled, it can't be enabled again using the current approach.

Related to #380.

### DIFF
--- a/app/charts/map/_map.tsx
+++ b/app/charts/map/_map.tsx
@@ -1,7 +1,6 @@
 import { MapController, WebMercatorViewport } from "@deck.gl/core";
-import { ScatterplotLayer } from "@deck.gl/layers";
-import { MapboxLayer } from "@deck.gl/mapbox";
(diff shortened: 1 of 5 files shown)


AnninaWalker:
Even if I check off the checkbox for Baselayer, the labels are still visible

Screenshot 2022-04-21 at 16 08 17

PR #401, opened 2022-02-22
Publish actions spacing

### LINKED ISSUE #382: Mobile UI alinements
**Describe the bug**
On the mobile device, the share and embed buttons are too close together. Spread them 16 px apart. 
The link [Created with visualize.admin.ch](https://int.visualize.admin.ch/en/v/Fz1D2p8CtLlj) should be on one line if possible.

**To Reproduce**
1. Go to [link](https://int.visualize.admin.ch/en/v/Fz1D2p8CtLlj ) on mobile

**Screenshots**
Image from iOS

bprusinowski:
Fixes #382.

I don't think we need a `direction=['column', 'row']` there, as "Share" and "Embed" buttons IMO should be next to each other if the space allows. Let me know what you think, I can also switch back to `Stack`, but it wasn't working as expected (open https://int.visualize.admin.ch/en/v/Fz1D2p8CtLlj and resize the window; the margin between items is removed at a point where they should be stacked, but they are not).

### DIFF
--- a/app/components/publish-actions.tsx
+++ b/app/components/publish-actions.tsx
@@ -1,4 +1,12 @@
 import { t, Trans } from "@lingui/macro";
+import * as clipboard from "clipboard-polyfill/text";
+import Downshift, { DownshiftState, StateChangeOptions } from "downshift";
+import {
+  MouseEvent as ReactMouseEvent,
+  ReactNode,
+  useEffect,
+  useState,
+} from "react";
 import {
   Box,
   Button,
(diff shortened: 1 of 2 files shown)


ptbrowne:
Weird, it seems Stack is missing here.

```
        display: "flex",
        flexDirection: directions,
 ```

I am fine with the flex gap, I checked for support recent version of Edge and it seems OK.

ptbrowne (dismissed):

ptbrowne:
Stack should use gap and flexbox I think. It'd be the best of both worlds what do you think ? I think it's a good component to have in the toolbox.

bprusinowski:
Hmmm, I think it makes sense. But then there would be no need for this custom CSS in favor of flexGap, right?

ptbrowne:
Right.

ptbrowne (approved):

PR #402, opened 2022-02-22
fix: Not changing filters when initial filters length = 2

### LINKED ISSUE #400: Graph does not reflect auto-selected filters on the left
When the value of filter 1 is changed, the content of filter 2 is automatically adjusted according to the availability of data, but not the graph on the right. The graph on the right-hand side retains the last value of filter 2. A refresh does not refresh the graph either, you have to select another value in filter 2 to refresh the graph. If there is only one value in filter 2, it cannot be displayed.

**To Reproduce**
1. open Bathing Stites in INT
2. choose bar chart
3. add color : parameter
4. add Filter Monitoring programm
5. Deplace the filter monitoring programm in position 1
6. Select BAQUA_SO in Filter Monitoring Programm
-> Impossible to display the graph for the bathing site "Solothurn"

**Screenshots**
image

INT v3.4.2 ([3e72a7d]
 - Edge

ptbrowne:
Possible filters query would make a too big slice of the filters resulting
in possible filters trying twice with 2 filters if filters initial length
is 2. This would result in empty results and the filters not being changed
in the UI.

fix #400

### DIFF
--- a/app/configurator/components/chart-configurator.tsx
+++ b/app/configurator/components/chart-configurator.tsx
@@ -170,6 +170,7 @@ export const useEnsurePossibleFilters = ({
         .query(PossibleFiltersDocument, {
           iri: state.dataSet,
           filters: unmappedFilters,
+          filterKey: Object.keys(unmappedFilters).join(", "),
         })
         .toPromise();
       if (error || !data) {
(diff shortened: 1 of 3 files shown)


bprusinowski (approved):
LGTM 👍

PR #415, opened 2022-02-22
Generic approach to rendering errors

bprusinowski:


app/components/common-chart.tsx
```diff
@@ -27,6 +30,12 @@ const GenericChart = ({
     queryFilters,
   };
 
+  // Remove any chart errors on chart switch
+  useEffect(
+    () => setChartError("none"),
+    [chartConfig.chartType, setChartError]
+  );
```
ptbrowne, line 37: This effect could be done directly in the hook if chartType is given to it.

app/charts/area/chart-area.tsx
```diff
@@ -41,44 +37,59 @@ export const ChartAreasVisualization = ({
...
+    });
 
   const observations = data?.dataCubeByIri?.observations.data;
+  const observationsPresent = useMemo(
+    () =>
+      observations
+        ? observations.length > 0 && observations.map((d) => d.y).some(isNumber)
+        : false,
+    [observations]
+  );
```
ptbrowne, line 58: I think it is a good as a first step.

For future improvement: I wonder if the data cubes observations query should not be handled by the hook directly so the whole data fetching / checking is done in a single hook, so that every chart can then use it easily.

const { data, error } = useChartData({
  dimensions,
  iri,
filters, 
  checkDatum: (d) => isNumber(d)
})

app/charts/shared/errors.tsx
```diff
@@ -0,0 +1,41 @@
...
+      {children}
+    </ChartErrorContext.Provider>
+  );
+};
+
+export const useChartError = () => {
+  return useContext(ChartErrorContext);
+};
+
+export const ChartErrorWrapper = ({ children }: { children: ReactNode }) => {
```
ptbrowne, line 31: I guess following my previous comment that it could be transformed into a useChartData / ChartDataWrapper hook / context.

app/components/chart-preview.tsx
```diff
@@ -140,7 +141,9 @@ const ChartWithInteractiveFilters = ({
           <ChartFiltersList dataSetIri={dataSet} chartConfig={chartConfig} />
         </Flex>
       )}
-      <Chart dataSet={dataSet} chartConfig={chartConfig} />
+      <ChartErrorProvider>
+        <Chart dataSet={dataSet} chartConfig={chartConfig} />
```
ptbrowne, line 145: Maybe the Chart could manage the ChartErrorWrapper wrapping ?

app/charts/area/chart-area.tsx
bprusinowski: Awesome idea, I have made an initial implementation 👌

bprusinowski: Btw. `observations.map((d) => d.y).some(isNumber)` seems long outdated, as `Observation` is` [dimensionIri: dimensionValue]`, there is no arbitrary "x" or "y" fields.

app/configurator/components/chart-type-selector.tsx
```diff
@@ -101,6 +103,10 @@ const ChartTypeSelectorField = ({
       disabled={disabled}
       label={label}
       {...field}
+      onClick={(e) => {
+        setChartError("none");
+        field.onClick(e);
+      }}
```
bprusinowski, line 109: I guess it's temporary, I will think about the way to handle that properly. This probably should happen inside configuratorState, on "CHART_TYPE_CHANGED" action type, but here I can't use a useChartError hook directly in reducer, so I think a bigger refactor would be needed?

app/configurator/configurator-state.tsx
```diff
@@ -1223,14 +1224,16 @@ export const ConfiguratorStateProvider = ({
...
+    <ChartErrorProvider>
+      <ConfiguratorStateProviderInternal
+        key={chartId}
+        chartId={chartId}
+        initialState={initialState}
+        allowDefaultRedirect={allowDefaultRedirect}
+      >
+        {children}
+      </ConfiguratorStateProviderInternal>
+    </ChartErrorProvider>
```
bprusinowski, line 1236: Needed to access `useChartError` from both chart-type-selector and charts themselves. I can also move this to some other part of the app, I am not sure where is the best place to do this :)

app/charts/shared/use-chart-data.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  if (!fetching) {
+    if (dataLoadingError) {
+      error = "dataLoading";
+    } else if (!observationsPresent) {
+      error = "noData";
+    }
+  }
+
+  useEffect(() => setChartError(error));
```
bprusinowski, line 35: The question is: should we clear chart errors on chart type switch and try to fetch the data again? I am not sure if it's a good approach.

Maybe it could be better to have
`useEffect(() => setChartError(error), [error, setChartError]);`
and remove `setChartError('none')` in the other place of the app, so if the data couldn't be loaded, we won't try to do this again, unless there is a page refresh. It would make the logic simpler I think, especially that other part with clearing the errors.

What do you think?

app/components/common-chart.tsx
bprusinowski: I changed the approach, so I think this is not relevant atm

app/charts/shared/errors.tsx
bprusinowski: Done!

PR #424, opened 2022-02-24
fix: Map tooltip text color

bprusinowski:
+ a small refactor for better readability


ptbrowne (approved):
👍

app/charts/map/map-tooltip.tsx
```diff
@@ -183,3 +152,36 @@ export const MapTooltip = () => {
...
   );
 };
+
+interface TooltipRowProps {
+  title: string;
+  background: string;
+  color: string;
+  value: string;
+  error: string | null;
+}
```
ptbrowne, line 162: Nit: Since there is no particular advantage from using `interface` over `type`, and we have a bit more type for props, let's settle on using `type`.

bprusinowski: Ok, I will change that :)

PR #429, opened 2022-02-24
feat: Improve tooltip placement

### LINKED ISSUE #315: Tooltip: Improvements to location
**Describe the bug**
The tooltip that shows the data at a specific point in the graph is not always visible in full on the screen. 

**To Reproduce**
1. Open the [link](https://int.visualize.admin.ch/en/v/Gw8Xb959gsMv) 
2. Hover the bar chart at the date 2020-09-21 
3. Look at the info tooltip
4. See the tooltip cut off at the top
Screenshot 2022-02-02 at 11 51 20

**Expected behavior**
If the tooltip is long, align its position to the side (it is currently aligned at the top). Reposition the triangle in the center of the tooltip. 
The correct behavior is visible if you go over the date 2020-07-06. See image. 

Screenshot 2022-02-02 at 11 52 16

 - Visualize environment and version: INT v3.2.7 
 - Browser and version: Chrome Version 97.0.4692.99

ptbrowne:
- Only allow top placement, this way we do not hide chart data (future work could include having the same behavior that the tooltip stays at the same place vertically). Checked with Annina
- Improve x placement by considering right edge of columns
- Tooltip is now absolutely positionned and portalled so that it can escape the chart container, this removes problems with tooltip being cut

Fix #315

### DIFF
--- a/app/charts/shared/interaction/tooltip-box.tsx
+++ b/app/charts/shared/interaction/tooltip-box.tsx
@@ -1,4 +1,10 @@
-import { ReactNode } from "react";
+import React, {
+  ReactNode,
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from "react";
 import { Box } from "theme-ui";
 import { Margins } from "../use-width";
 import {
@@ -9,6 +15,8 @@ import {
   Yplacement,
 } from "./tooltip";
 import { useTheme } from "../../../themes";
+import ReactDOM from "react-dom";
+import { throttle } from "lodash";
 
 export interface TooltipBoxProps {
   x: number | undefined;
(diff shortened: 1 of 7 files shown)


app/charts/shared/interaction/tooltip-box.tsx
```diff
@@ -27,51 +74,60 @@ export const TooltipBox = ({
...
+            }}
+          >
+            <Box
+              sx={{
+                width: "fit-content",
+                padding: 3,
+                whiteSpace: "nowrap",
+                pointerEvents: "none",
+                backgroundColor: "monochrome100",
+                borderRadius: "default",
```
bprusinowski, line 101: Is this needed?

ptbrowne: (This is not code I changed here)

What is "this" in your question ?

ptbrowne:
Default is the default border radius from the theme and it is not 0 ;) It is 3. Maybe something to rename in the future.

bprusinowski (approved):
LGTM, I like this approach!

app/charts/shared/interaction/tooltip-box.tsx
```diff
@@ -27,51 +74,60 @@ export const TooltipBox = ({
...
+    <>
+      <div ref={posRef} />
+      {pos ? (
+        <Portal>
+          <Box
+            style={{
+              zIndex: 10,
+              position: "absolute",
+              left: x! + margins.left + pos?.left,
+              top: mxYOffset(y!, placement) + margins.top + pos?.top,
```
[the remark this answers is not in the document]
ptbrowne: Right, I removed the condition to display the portal and the "?", thanks

PR #448, opened 2022-03-10
Wrap exported components

ptbrowne:
- v3.4.10-beta.2
- feat: Update preconstruct
- feat: Wrap exported components with the theme and i18n so that host app does not have to do it


ptbrowne:
Closing as we won't use embedded visualize inside BLW.

PR #449, opened 2022-03-11
Search UX improvements

ptbrowne:
Various small improvements that make the search more agreeable.

- Move include drafts box next to sort options so that it does not appear on input focus
- Add small animations when searching for smoother UX
- Do not fire the search at each keystroke as this pattern is better suited for autocomplete and we are not in an autocomplete


app/components/form.tsx
```diff
@@ -56,7 +55,7 @@ export const Label = ({
             textAlign: "left",
             fontFamily: "body",
             pr: 1,
-            fontSize: [3],
+            fontSize: smaller ? [2, 2, 2] : [3, 3, 3],
```
bprusinowski, line 59: Won't `smaller ? 2 : 3` work in the same way?

bprusinowski (approved):
LGTM 👍

app/components/form.tsx
ptbrowne: Yes, thanks, don't know why I left this... FIxed ;)

PR #453, opened 2022-03-17
Migrate to MUI

ptbrowne:
Migrate from theme UI to Material UI v5.

Upgrading to material UI will let us use more complex components like date pickers, autocompletes and more. The federal theme should be shared across projects going forward.

The migration was done through

- Search and replace using `fastmod`
- Custom codemods using `jscodeshift`
- Manual pass


bprusinowski (approved):
💯

PR #455, opened 2022-03-21
Add an ability to switch between chart and table view

bprusinowski:
- added an ability to switch between chart and data preview in published & preview charts

### DIFF
--- a/app/components/chart-footnotes.tsx
+++ b/app/components/chart-footnotes.tsx
@@ -1,11 +1,12 @@
 import { Trans } from "@lingui/macro";
+import { Box, Button, Link, Stack, Typography } from "@mui/material";
 import { useEffect, useState } from "react";
-import { Box, Link, Typography } from "@mui/material";
 import { ChartConfig } from "../configurator";
 import { useDataCubeMetadataWithComponentValuesQuery } from "../graphql/query-hooks";
+import { getChartIcon, Icon } from "../icons";
 import { useLocale } from "../locales/use-locale";
+import { useChartTablePreview } from "./chart-table-preview";
 import { DataDownload } from "./data-download";
-import { Stack } from "@mui/material";
 
 export const ChartFootnotes = ({
   dataSetIri,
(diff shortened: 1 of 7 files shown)


ptbrowne:
I see this 

image

on https://visualize-ad-feat-add-r-aoyb9f.herokuapp.com/en/v/3p3klQ06RoMF

It seems like the design has some problems here ? 

Also, to smooth out going back and forth maybe a `<motion.div layout>` could be used here ? so that the height is properly animated ?

bprusinowski:
Thanks @ptbrowne, I was focusing on the preview chart and I missed that 🫢 I also added Motion layout as suggested, I think it only makes sense in published chart, as for preview everything gets animated, including transitions between chart editing steps (and this was looking kinda distracting).

ptbrowne (approved):

sosiology:
@bprusinowski i really like the animination when switching between table and chart view. While testing i notice it changes the width of the component https://test.visualize.admin.ch/en/v/4linf1VNEHqx could this eventually be an issue for someone trying to embedd the chart? Or will it be possibile to not offer this option when creating the chart, similar to for example the interactive filters?

AnninaWalker:
@bprusinowski As discussed, please add a label next to the table icon with: Switch to table view or Switch to chart view
Screenshot 2022-03-28 at 16 52 39

AnninaWalker:
Referring to the comment of @sosiology . I've noticed that I see a scrollbar between the table but I'm not able to scroll

Screenshot 2022-03-28 at 17 00 02

AnninaWalker:
Minor detail: But I would change the order of the elements to:
Download, Switch to table view and then the SPARQL query
Screenshot 2022-03-29 at 12 08 44

AnninaWalker:
@bprusinowski I see that this issue is in the "Done" Lane but the lastest feedback hasn't been implemented. See above:

- Adding a label next to the Switch Icon
- Change the order of the elements
- Scrollbar in the tables

bprusinowski:
Hi @AnninaWalker, I think it changed automatically when the PR was merged – but you're right, this shouldn't be in the done lane, I started to introduce modifications, but didn't finish yet. Same for the second PR, I will adjust the labels!

AnninaWalker:
Perfect, thank you!

AnninaWalker:
Something I just noticed. The fontstyle for the Download Button, Run SPARQL should be "caption" and not Body2. It should be 12px instead of 14px.
Screenshot 2022-04-05 at 12 02 15

bprusinowski:
Thanks @sosiology and @AnninaWalker for your inputs! I addressed them in #475.

PR #459, opened 2022-03-22
Upgrade data download UI

bprusinowski:
This PR adds new data download UI (popover).

### DIFF
--- a/app/components/data-download.tsx
+++ b/app/components/data-download.tsx
@@ -1,5 +1,11 @@
 import { Trans } from "@lingui/macro";
-import { Button, Link, MenuItem, Typography } from "@mui/material";
+import {
+  Button,
+  CircularProgress,
+  Link,
+  MenuItem,
+  Typography,
+} from "@mui/material";
 import { saveAs } from "file-saver";
 import { keyBy } from "lodash";
 import {
@@ -8,12 +14,12 @@ import {
   usePopupState,
 } from "material-ui-popup-state/hooks";
 import HoverMenu from "material-ui-popup-state/HoverMenu";
-import React, { memo, ReactNode, useCallback, useMemo } from "react";
+import React, { memo, ReactNode, useCallback, useMemo, useState } from "react";
 import { useQueryFilters } from "../charts/shared/chart-helpers";
 import { ChartConfig } from "../configurator";
 import { Observation } from "../domain/data";
 import { useDataCubeObservationsQuery } from "../graphql/query-hooks";
-import { Icon, IconName } from "../icons";
+import { Icon } from "../icons";
 import { useLocale } from "../locales/use-locale";
 
 const EXTENTS = ["visible", "all"] as const;
@@ -99,7 +105,10 @@ export const DataDownloadMenu = memo(
 
     return (
       <>
-        <DownloadButton iconName="download" {...bindHover(popupState)}>
+        <DownloadButton
+          icon={<Icon name="download" size={16} />}
+          {...bindHover(popupState)}
+        >
           <Trans id="button.download">Download</Trans>
         </DownloadButton>
         <HoverMenu
@@ -122,7 +131,7 @@ export const DataDownloadMenu = memo(
               title={title}
               extent={d.extent}
               fileFormat={d.fileFormat}
-              closeMenu={popupState.close}
+              onDownloaded={popupState.close}
               {...(d.extent === "all" ? allPrepared : visiblePrepared)}
             />
           ))}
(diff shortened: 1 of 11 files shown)


app/components/data-download.tsx
```diff
@@ -98,43 +62,162 @@ const DataDownloadInner = memo(
...
+        data,
+        fileFormat,
+      }),
+    }).then((res) =>
+      res.blob().then((blob) => saveAs(blob, `${title}.${fileFormat}`))
+    );
+  }, [columnKeys, data, fileFormat, title]);
+
+  return (
+    <DownloadMenuItemInner onClick={() => download().then(() => closeMenu())}>
```
ptbrowne, line 167: Nit: I think it would be cleaner if the closeMenu function was kept inside the menu and if the DownloadMenuItemInner did not have the closeMenu props. I think it would be a bit better to have a `onDownloadedProp` on the MenuItem. This way menu things are kept at menu level and responsibilities are not spread out.

app/components/data-download.tsx
```diff
@@ -98,43 +62,162 @@ const DataDownloadInner = memo(
...
+      {extent === "visible" ? (
+        <>
+          <Trans id="button.download.data.visible">Overlooking dataset</Trans> (
+          {fileFormat.toUpperCase()})
+        </>
+      ) : (
+        <>
+          <Trans id="button.download.data.all">Full dataset</Trans> (
+          {fileFormat.toUpperCase()})
+        </>
```
ptbrowne, line 177: While download is happening, it seems the user has no visual feedback ? Maybe a spinner could be added while download() runs ?

app/components/data-download.tsx
```diff
@@ -98,43 +62,162 @@ const DataDownloadInner = memo(
...
+  children,
+  onClick,
+}: {
+  children: ReactNode;
+  onClick: () => void;
 }) => {
+  return (
+    <MenuItem onClick={onClick} sx={{ ...PADDING_PROPS }}>
+      <DownloadButton iconName="table">{children}</DownloadButton>
+    </MenuItem>
```
ptbrowne, line 193: Nit: Am not sure this component is very useful here since it is used only one time.

app/components/data-download.tsx
```diff
@@ -98,43 +62,162 @@ const DataDownloadInner = memo(
...
+}) => {
+  return (
+    <Button
+      component="a"
+      variant="text"
+      color="primary"
+      size="small"
+      startIcon={<Icon name={iconName} size={16} />}
+      {...props}
+      sx={{ fontWeight: "regular", ":hover": { color: "primary.main" } }}
```
ptbrowne, line 213: Why do we need to pass sx with hover here, is the style really that different from "variant=text" or should we align variant='text" theme ? Better to add a typography variant here if necessary here ? We shuld not have in general to pass sx for customisations (exception would be for padding/margin override).

app/components/data-download.tsx
```diff
@@ -98,43 +62,162 @@ const DataDownloadInner = memo(
...
+          <Trans id="button.download">Download</Trans>
+        </DownloadButton>
+        <HoverMenu
+          {...bindMenu(popupState)}
+          anchorOrigin={{ vertical: "top", horizontal: "center" }}
+          transformOrigin={{ vertical: "bottom", horizontal: "center" }}
+        >
+          <Typography
+            sx={{
+              ...PADDING_PROPS,
```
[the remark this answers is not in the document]
bprusinowski: This is just a placeholder text (menu title), so I am not sure it should have MenuItem appearance (mostly onHover styling I think) - it looks like it can be selected then. I needed the same padding for both download options & the title, so that's why there is this prop. Do you see a better approach here?

app/components/data-download.tsx
bprusinowski: Yes, this is on my TODO list 😅

app/components/data-download.tsx
bprusinowski: This is due to the fact that MenuItem has its own onHover styling; if I left the default styling for this button, then there would be two onHover styles, one in general for MenuItem, and second one if you hover over the text (color would change).  I think I can also replace this button with just typography + icon, I used it because it's the same as the button to open the menu, but it also will be possible to have the same styling with other elements.

app/components/data-download.tsx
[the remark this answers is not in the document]
ptbrowne: Either have a MenuTitle component, or maybe use the subheader prop on a MenuList component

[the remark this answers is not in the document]
ptbrowne: I'd use the menuListProps={{ subheader: 'Download' }} to be able to style the subheader inside the theme. Weirdly by the subheader is not themed.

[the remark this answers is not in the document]
ptbrowne: If not using the theme, I think it would be clearer with

```
const MenuTitle = styled(Typography)({ ...PADDING_PROPS, borderBottom: '1px solid', borderBottomColor: 'grey.500' })
```

The disadvantage here would be that styling would not be in the theme.

app/components/data-download.tsx
ptbrowne: Yes, I think it is unnecessary to reuse DownloadButton inside the MenuItem, since it complicates things = we have to cancel out the style. Better to only use `<><ListItemIcon><Icon name={iconName} size={16} /></ListItemIcon><ListItemText>Download</ListItemText></>` IMHO.

ptbrowne (approved):
👍 Thanks !

PR #463, opened 2022-03-23
feat: Remove async loading of theme so that hot reload works

ptbrowne:
Previously, we had a dark theme and a federal theme and the dark theme could
be asynchronously loaded via a query param. Since we removed the dark theme,
we are only left with one theme so async loading does not make sense.

Also, with async loading, hot reload was not working with the theme, we had
to reload the page after changes to the theme.﻿


bprusinowski (approved):
👌

PR #467, opened 2022-03-25
Loading observations improvements

ptbrowne:
- feat: Disable distinct when there is no filter to improve performance
- feat: Add way to limit the observations query


ptbrowne:
Merging now to unlock BLW datasets visualization on int.

PR #475, opened 2022-03-29
Minor layout improvements (data download & switch to table preview)

bprusinowski:
This PR:
- fixes scrolling in published chart,
- fixes sizing & display of published chart (consulted with @AnninaWalker),
- updates DataDownload UI,
- changes the order of footnotes items,
- adds a label to switch to table icon,
- introduces new sorting arrow icons to Table Chart.


app/components/data-download.tsx
```diff
@@ -296,8 +296,8 @@ const DownloadMenuItem = ({
       >
         {extent === "visible" ? (
           <>
-            <Trans id="button.download.data.visible">Overlooking dataset</Trans>{" "}
-            ({fileFormat.toUpperCase()})
+            <Trans id="button.download.data.visible">Visible dataset</Trans> (
```
ptbrowne: 💯

app/components/chart-footnotes.tsx
```diff
@@ -67,17 +67,30 @@ export const ChartFootnotes = ({
...
+                      isChartTablePreview
+                        ? getChartIcon(chartConfig.chartType)
+                        : "table"
+                    }
+                    size={16}
+                  />
+                }
                 onClick={() => setIsChartTablePreview(!isChartTablePreview)}
-                sx={{ width: "20px", minWidth: 0, padding: 0 }}
+                sx={{ padding: 0, fontWeight: "regular" }}
```
ptbrowne, line 96: Nit: Am wondering why we need to put a fontWeight here, maybe we would need to have a new button variant ? Or have the text variant of the button as fontWeight: regular ?

app/components/data-download.tsx
```diff
@@ -28,16 +27,16 @@ import {
...
 const FILE_FORMATS = ["csv", "xlsx"] as const;
 export type FileFormat = typeof FILE_FORMATS[number];
 
-const OPTIONS_TO_RENDER = EXTENTS.flatMap((extent) =>
-  FILE_FORMATS.map((fileFormat) => ({ extent, fileFormat }))
-);
+type PreparedData = {
+  columnKeys: string[];
+  data: Observation[];
+  sparqlEditorUrl: Maybe<string> | undefined;
```
ptbrowne: Maybe add a "?" `sparqlEditorUrl?: Maybe` ? This way we do not have to add `sparqlEditorUrl: undefined`

app/components/data-download.tsx
```diff
@@ -134,62 +137,44 @@ export const AllAndVisibleDataDownloadMenu = memo(
...
-              : visiblePrepared.columnKeys,
-          data: d.extent === "all" ? allPrepared.data : visiblePrepared.data,
-        })),
-      [allPrepared, visiblePrepared]
-    );
+    const preparedAllData = usePreparedAllData({ dataSetIri });
+    const preparedVisibleData = usePreparedVisibleData({
+      dataSetIri,
+      chartConfig,
+    });
```
ptbrowne: Does this mean that we download all the data, and all the visible data in every case ? For performance reasons, maybe we should only download the data when the user has opened the menu ?

app/components/data-download.tsx
```diff
@@ -246,67 +237,91 @@ const DataDownloadInnerMenu = ({
...
         setIsDownloading(true);
         await download();
         onDownloaded();
         setIsDownloading(false);
       }}
-      sx={{ paddingY: 3 }}
+      sx={{
+        padding: 0,
+        minWidth: 0,
+        pointerEvents: isDownloading ? "none" : "auto",
```
ptbrowne, line 244: Instead of disabling pointerEvents, maybe the button itself should be disabled ?

Nit: For this usecase, I like to have a LoadingButton that receives a fetching prop (here it would receive "downloading" and adds the icon. I think since it is a generic usecase, it is good to have a generic component for this.

Nit: There is also the useDisclosure hook that gives you a way to easily use a boolean.

app/components/data-download.tsx
```diff
@@ -246,67 +237,91 @@ const DataDownloadInnerMenu = ({
...
-    <MenuItem
+    <Button
+      variant="text"
+      size="small"
+      endIcon={isDownloading ? <CircularProgress /> : null}
       onClick={async () => {
         setIsDownloading(true);
         await download();
         onDownloaded();
         setIsDownloading(false);
```
ptbrowne, line 239: Can you wrap download() and onDownloaded() inside a try, with the setIsDwonloading(false) inside a finally ? This way if there is an error, the spinner will be removed from the button.

app/charts/table/table-content.tsx
```diff
@@ -93,25 +92,12 @@ export const TableContent = ({ children }: { children: React.ReactNode }) => {
...
-                            name={
-                              column.isSortedDesc
-                                ? "sortDescending"
-                                : "sortAscending"
-                            }
-                            size={SORTING_ARROW_WIDTH}
-                          />
-                        </Box>
-                      )}
+                      <TableSortLabel
```
ptbrowne, line 95: Didn't know this component, nice 👍

app/components/chart-footnotes.tsx
```diff
@@ -33,6 +37,17 @@ export const ChartFootnotes = ({
...
   });
 
+  // Data for data download
+  const filters = useQueryFilters({ chartConfig });
+  const [{ data: visibleData }] = useDataCubeObservationsQuery({
+    variables: { locale, iri: dataSetIri, dimensions: null, filters },
+  });
+  const [{ data: allData }] = useDataCubeObservationsQuery({
+    variables: { locale, iri: dataSetIri, dimensions: null, filters: null },
+  });
```
ptbrowne, line 47: I think it would be good to [pause](https://formidable.com/open-source/urql/docs/basics/react-preact/#pausing-usequery) those requests unless the menu has been open once. This way we do not download the data if there is no filter.

app/components/chart-footnotes.tsx
```diff
@@ -118,17 +118,22 @@ export const ChartFootnotes = ({
...
             <>
               <Box sx={{ display: "inline", mx: 2 }}>·</Box>
-              <Link
-                sx={{ typography: "caption" }}
+              <Button
+                component="a"
+                variant="text"
                 color="primary"
+                size="small"
+                sx={{ fontWeight: "regular", padding: 0 }}
```
ptbrowne, line 126: Nit: Again, this fontWeight: regular :) Should font-weight regualr be the default for Button with variant "text" ? @AnninaWalker do you have a recommendation ?

ptbrowne (approved):
I am a bit concerned with the requests for the full dataset not being paused. I think it would be good to solve this problem before merging.

The rest looks good to me 👍

app/components/chart-footnotes.tsx
bprusinowski: It's because there is a `fontWeight: bold` when `size="small"` - I can create another variant or change this? I am not sure is size small should implicate a bold font weight, IMO it's slightly confusing, what do you think? 😅

app/components/data-download.tsx
ptbrowne: I meant that this was a component that I like to have in my codebase, not that it was inside MUI. Although I see it in the Chinese version of the docs ? https://mui.com/zh/material-ui/api/loading-button/

app/components/chart-footnotes.tsx
bprusinowski: Ok, so I'll change that and use sx={{fontWeight: "bold"}} for the stepper buttons.

app/components/chart-footnotes.tsx
AnninaWalker: @ptbrowne Not exactly sure what you mean with variant "text"? Buttons fontweight should be regular yes.

AnninaWalker:
@bprusinowski I discovered one thing. I the published chart I see now this grey area at the bottom of the card. Not sure where this is coming from.
Screenshot 2022-04-12 at 16 18 29

app/components/chart-footnotes.tsx
ptbrowne: The variant "text" is when the button has not outline nor fill : https://mui.com/material-ui/react-button/#basic-button. I had put fontWeight: bold onto the button size small + variant text since in the stepper, the links were bold. But now Bartosz has changed the logic to keep fontWeight: regular as the default, and have fontWeight: bold as overrides for the stepper buttons. Does it sound OK to you ?

AnninaWalker: Yes that sounds ok 👍

bprusinowski:
@AnninaWalker could you share a link and information on which browser you used when you encountered this bug? I can't replicate it :(

app/components/data-download.tsx
bprusinowski: I refactored the code to actually download the data when user clicks a button, thanks for pointing this out :)

app/components/chart-footnotes.tsx
bprusinowski: 👌

app/components/data-download.tsx
```diff
@@ -199,27 +182,40 @@ const DataDownloadMenuSection = ({
...
+      return fetch("/api/download", {
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify({
+          columnKeys,
+          data,
+          fileFormat,
+        }),
```
ptbrowne: ~~ There is something I do not follow here, Why do we send back the data to the `/api/download` endpoint ? ~~

OK I checked in the code and it seems like we need to send back the data for the data to be converted into an Excel file. Ideally I guess all the data download should be done at the server side so that we do not have to do a download and then an upload and then again a download. What's your take on this ?

Since this PR is not about this per se, maybe we should keep the change for a next PR.

app/components/data-download.tsx
```diff
@@ -228,15 +224,35 @@ const DownloadMenuItem = ({
...
         setIsDownloading(true);
+        urqlClient
+          .query(DataCubeObservationsDocument, {
+            locale,
+            iri: dataSetIri,
+            dimensions: null,
+            filters,
+          })
+          .toPromise()
+          .then(async (result: OperationResult<DataCubeObservationsQuery>) => {
```
ptbrowne, line 303: Using an await here would be clearer here.

app/components/data-download.tsx
bprusinowski: I am not sure if caching would work in this case? For TEST environment I took the advantage of caching to fetch SparqlEditorUrl, which is the same query to fetch filtered data. I would need to check if this is the case, otherwise we would duplicate these queries – if it's not the case, then I agree it makes sense to also change that :)

AnninaWalker:
@bprusinowski I used Google Chrome Version 100.0.4896.75 (Official Build) (x86_64)
Still see the bug on your deployment link

app/components/data-download.tsx
AnninaWalker: @bprusinowski I used Google Chrome Version 100.0.4896.75 (Official Build) (x86_64)
Still see the bug on your deployment link

AnninaWalker: Screenshot 2022-04-13 at 14 37 49

bprusinowski: I can't reproduce this on any of the browsers (Chrome, Safari, Firefox, Edge) – maybe it's a rare thing only happening on this version of Chrome on Intel chips? I don't really know how I can proceed with this at this point, maybe we'll see if it also occurs on TEST environment after deploying?

PR #476, opened 2022-03-29
Enable internal links inside documentation

ptbrowne:
Previously, in the catalog documentation heading had the slug="wat" (hardcoded) which had the effect of generating anchor links with `page?a=wat`.

For anchor links to correctly work

- `useBrowserHistory` is set in catalog's options so that the routes have "real URLs", which also has the effect to disable the `href=page?a=slug` query parameter behavior (now links have href=page#slug).
- For next.js to be aware of the docs/ pages, `[...rest].tsx` is created and it used the component exported in `docs/index.tsx`
- I used github-slugger to slugify the content of the headings.

Can be tested on https://visualize-ad-docs-inter-irh9oi.herokuapp.com/docs/charts/rdf-to-visualize#possible-chart-types (you should land at the "Possible chart types" heading).

Also, per the suggestion of @wereHamster , I refactored relative links into "@/" links for easier moving around files.


app/pages/docs/index.tsx
```diff
@@ -17,48 +17,55 @@ import {
...
+import Slugger from "github-slugger";
+
+const slugger = new Slugger();
 
 const pages: ConfigPageOrGroup[] = [
-  { path: "/", title: "Introduction", content: require("../docs/index.mdx") },
+  {
+    path: "/",
+    title: "Introduction",
+    content: require("../../docs/index.mdx"),
```
wereHamster, line 30: Maybe you can set up paths in tsconfig and then use absolute paths in import declarations / require calls.

```suggestion
    content: require("@/docs/index.mdx"),
```

After changing in tsconfig:

```
    "paths": {
      "@/components/*": ["./components/*"],
      "@/docs/*": ["./docs/*"]
    }
```

ptbrowne: Yes, that was something that was bothering me for a while actually. Did it in [`e713737` (#476)](https://github.com/visualize-admin/visualization-tool/pull/476/commits/e713737036415d4eae50dd8c8b68299aefc3728d), thanks for the suggestion.

app/browser/dataset-browser.tsx
```diff
@@ -1,6 +1,6 @@
-import { AppLayout } from "../components/layout";
-import { SelectDatasetStep } from "../configurator/components/select-dataset-step";
-import { ConfiguratorStateProvider } from "../src";
+import { AppLayout } from "@/components/layout";
+import { SelectDatasetStep } from "@/configurator/components/select-dataset-step";
+import { ConfiguratorStateProvider } from "@/src";
```
bprusinowski, line 3: 👌

PR #497, opened 2022-04-11
fix: Do not show loading when dataset browse query is in error

ptbrowne:
Previously, even if the query for loading datasets was in error, the spinner
was shown. Now, a proper error is shown.﻿


bprusinowski (approved):
💯

PR #507, opened 2022-04-14
WIP: Maintain chart config when switching chart type

### LINKED ISSUE #500: Maintain Filters when changing to another chart type (Step 1a)

**Usecases and impact**
Examples of how the feature would be beneficial and an estimation of how much impact it would have.

**Additional context**
If the feature concerns the editor, please provide a dataset, class of datasets.
If the feature concerns publication, please provide an example of where the chart is published.

bprusinowski:
Closes #500.

Tasks:
- [ ] implement a way to dynamically retain chart config options
  - [x] x & y fields
  - [ ] color mappings
  - [ ] filters
  - [ ] interactive filters

### DIFF
--- a/app/charts/index.ts
+++ b/app/charts/index.ts
@@ -1,11 +1,15 @@
 import { get, groupBy } from "lodash";
 import {
+  AreaConfig,
   ChartConfig,
   ChartType,
   ColumnConfig,
   GenericFields,
   InteractiveFiltersConfig,
   LineConfig,
+  MapConfig,
+  PieConfig,
+  ScatterPlotConfig,
   TableColumn,
 } from "../configurator";
 import { mapColorsToComponentValuesIris } from "../configurator/components/ui-helpers";
@@ -121,9 +125,7 @@ export const getInitialConfig = ({
         filters: {},
         interactiveFiltersConfig: INITIAL_INTERACTIVE_FILTERS_CONFIG,
         fields: {
-          x: {
-            componentIri: getTimeDimensions(dimensions)[0].iri,
-          },
+          x: { componentIri: getTimeDimensions(dimensions)[0].iri },
           y: { componentIri: measures[0].iri },
         },
       };
@@ -133,9 +135,7 @@ export const getInitialConfig = ({
         filters: {},
         interactiveFiltersConfig: INITIAL_INTERACTIVE_FILTERS_CONFIG,
         fields: {
-          x: {
-            componentIri: getTimeDimensions(dimensions)[0].iri,
-          },
+          x: { componentIri: getTimeDimensions(dimensions)[0].iri },
           y: { componentIri: measures[0].iri, imputationType: "none" },
         },
       };
@@ -266,6 +266,7 @@ export const getChartConfigAdjustedToChartType = ({
 
   mkChartConfigAdjuster({
     chartAdjustConfig: CHART_ADJUST_CONFIGS[chartType],
+    pathOverrides: PATH_OVERRIDES_CONFIG[chartType],
     oldChartConfig: chartConfig,
     newChartConfig,
     dimensions,
@@ -276,7 +277,6 @@ export const getChartConfigAdjustedToChartType = ({
 };
 
 const CHART_ADJUST_CONFIGS = {
-  area: {},
   bar: {},
   column: {
     fields: {
@@ -284,8 +284,12 @@ const CHART_ADJUST_CONFIGS = {
         componentIri: ({
           oldValue,
           newChartConfig,
+          dimensions,
         }: FieldAdjustParams<ColumnConfig>) => {
-          newChartConfig.fields.x.componentIri = oldValue;
+          // When switching from a scatterplot, x is a measure.
+          if (dimensions.map((d) => d.iri).includes(oldValue)) {
+            newChartConfig.fields.x.componentIri = oldValue;
+          }
         },
       },
       y: {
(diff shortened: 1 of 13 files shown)


app/configurator/configurator-state.spec.tsx
```diff
@@ -362,3 +379,85 @@ describe("moveField", () => {
...
+    ).iri;
+
+    // It should not retain GeoShapesDimension on the X axis when switching to
+    // a line chart.
+    expect(newConfig.fields.x.componentIri).toEqual(expectedLineXComponentIri);
+    expect(newConfig.fields.x.componentIri).not.toEqual(
+      oldConfig.fields.x.componentIri
+    );
+  });
+});
```
bprusinowski, line 463: Is this the right way to test it? I wanted to mimic actual user behavior of switching chart types in a chain, but maybe this should be more explicit with actual chart config values visible to the user, e.g.

`const expectedChartConfig = {...}`

?

app/charts/index.ts
```diff
@@ -685,31 +666,35 @@ const mkChartConfigAdjuster = ({
...
-          dimensions,
-          measures,
-        });
-      } else {
-        go({
-          path: newPath,
-          field: v,
-        });
+      if (v !== undefined) {
+        if (shouldAdjustConfigField(newPath, v)) {
```
ptbrowne: Nit: I just thought that maybe isConfigLeaf could be maybe more understandable ?

app/charts/index.ts
```diff
@@ -480,7 +480,14 @@ const CHART_ADJUST_CONFIG: {
...
         newChartConfig,
       }: FieldAdjustParams<ColumnConfig>) => {
-        newChartConfig.fields.segment = oldValue;
+        newChartConfig.fields.segment = {
+          ...oldValue,
+          type: "stacked",
+          sorting: oldValue.sorting || {
+            sortingOrder: "asc",
+            sortingType: "byDimensionLabel",
+          },
```
ptbrowne, line 477: Nit: Extract this as DEFAULT_SORTING ? Isn't there alrady a default like this somewhere that we could reuse ?

app/charts/index.ts
```diff
@@ -536,7 +547,15 @@ const CHART_ADJUST_CONFIG: {
...
       segment: ({ oldValue, newChartConfig }) => {
-        newChartConfig.fields.segment = oldValue;
+        newChartConfig.fields.segment = {
+          componentIri: oldValue.componentIri,
+          palette: oldValue.palette,
+          colorMapping: oldValue.colorMapping,
+          sorting: oldValue.sorting || {
+            sortingOrder: "asc",
+            sortingType: "byDimensionLabel",
+          },
```
ptbrowne, line 546: Nit: Reuse the default here also

app/charts/index.ts
```diff
@@ -261,8 +262,8 @@ export const getChartConfigAdjustedToChartType = ({
     measures,
   });
   const adjustChartConfig = mkChartConfigAdjuster({
-    chartConfigAdjusters: CHART_CONFIGS_ADJUSTERS[chartType],
-    pathOverrides: PATH_OVERRIDES_CONFIG[chartType],
+    adjusters: chartConfigsAdjusters[chartType],
```
ptbrowne, line 265: Thanks for simplifying the names 👍

app/configurator/interactive-filters/editor-time-brush.tsx
```diff
@@ -145,9 +147,23 @@ export const EditorBrush = ({
...
+
+    if (interactiveFiltersConfig) {
+      const { from, to } = interactiveFiltersConfig.time.presets;
+
+      if (from && to) {
+        const fromDate = new Date(from);
+        const toDate = new Date(to);
+        defaultSelection = [timeScale(fromDate), timeScale(toDate)];
+      }
+    }
```
ptbrowne, line 165: 👍

app/configurator/configurator-state.spec.tsx
ptbrowne: I think it's OK right now.

One possibility I see would be to have it more declarative when you say what should happen.
Then you can build multiple scenarios more easily.

```
const checks = [
  {
    type: 'pie',
    checks: [
      ['y.componentIri', 'y.componentIri', true] 
    ]
  },
  { 
    type: 'column',
    checks: [['segment.componentIri', 'segment.componentIri', true]]
},
{
  type: 'map',
  checks:[
    ['areaLayer.componentIri', 'x.componentIri', false]
  ]
}
   
for (let nextChartType of nextChartTypes) {
   ... // code the expect etc :)
}

```

ptbrowne (approved):
Very nice ! I have made some remarks but they are mostly nits. It was very nice to read the PR as everything was nicely separated, thanks !

app/charts/index.ts
```diff
@@ -245,6 +247,113 @@ export const getInitialConfig = ({
...
+  oldChartConfig,
+  newChartConfig,
+  dimensions,
+  measures,
+}: {
+  chartAdjustConfig: ChartAdjustConfig;
+  oldChartConfig: ChartConfig;
+  newChartConfig: ChartConfig;
+  dimensions: DataCubeMetadata["dimensions"];
+  measures: DataCubeMetadata["measures"];
```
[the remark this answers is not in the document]
bprusinowski: Yes, they were not used in this commit, but they are used in the following commits when switching chart type to a scatterplot :D

app/charts/index.ts
```diff
@@ -281,42 +281,63 @@ const CHART_ADJUST_CONFIGS = {
...
-        componentIri: (
-          previousIri: string,
-          oldChartConfig: ChartConfig,
-          newChartConfig: ColumnConfig
-        ) => {
-          newChartConfig.fields.x.componentIri = previousIri;
+        componentIri: ({
+          oldValue,
+          newChartConfig,
+        }: FieldAdjustParams<ColumnConfig>) => {
```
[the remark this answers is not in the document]
bprusinowski: The type name has changed in the following commits

app/charts/index.ts
```diff
@@ -631,6 +624,16 @@ const CHART_ADJUST_CONFIG = {
...
+type FieldAdjustParams<NewChartConfigType> = {
+  oldValue: any;
+  oldChartConfig: ChartConfig;
+  newChartConfig: NewChartConfigType;
+  dimensions: DataCubeMetadata["dimensions"];
+  measures: DataCubeMetadata["measures"];
+};
+type FieldAdjuster<NewChartConfigType> = (
+  params: FieldAdjustParams<NewChartConfigType>
+) => void;
```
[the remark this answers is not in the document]
bprusinowski: I have moved them to config-types afterwards :)

app/charts/index.ts
bprusinowski: Good point! 👌

app/charts/index.ts
```diff
@@ -277,6 +277,161 @@ export const getChartConfigAdjustedToChartType = ({
...
+      type: (params: FieldAdjustParams<ChartConfig>) => void;
+      from: (params: FieldAdjustParams<ChartConfig>) => void;
+      to: (params: FieldAdjustParams<ChartConfig>) => void;
+    };
+  };
+  dataFilters: {
+    active: (params: FieldAdjustParams<ChartConfig>) => void;
+    componentIris: (params: FieldAdjustParams<ChartConfig>) => void;
+  };
+} = {
```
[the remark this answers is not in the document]
bprusinowski: I tried to base this type on `InteractiveFiltersConfig` type, but there is a mismatch between legend fields (it's treated as a literal, so we don't go down to `active` & `componentIri` fields anymore and I also had some problems with excluding array type:

```
type NestedFieldAdjusters<T> = {
  [K in keyof T]: T[K] extends object
    ? NestedFieldAdjusters<T[K]>
    : FieldAdjuster<ChartConfig, T[K]>;
};

export type InteractiveFiltersAdjusters =
  NestedFieldAdjusters<InteractiveFiltersConfig>;
```

which works ok, but not for a dataFilters.componentIris field (and a legend field, but I could change it to use active & componentIri fields to make it compatible.

Do you have an idea of how to do this elegantly? I don't want to spend more time on this and maybe you've done something like this before 😅

app/configurator/configurator-state.tsx
```diff
@@ -602,7 +602,7 @@ const reducer: Reducer<ConfiguratorState, ConfiguratorStateAction> = (
         const { chartType, dataSetMetadata } = action.value;
 
         draft.chartConfig = getChartConfigAdjustedToChartType({
-          chartConfig: draft.chartConfig,
+          chartConfig: current(draft.chartConfig),
```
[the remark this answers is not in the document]
bprusinowski: I have introduced immer there, does it look better now?

app/configurator/configurator-state.spec.tsx
bprusinowski: Great idea, done, thanks! 💯

app/charts/index.ts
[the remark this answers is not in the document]
ptbrowne: I think to work it out, I'd have to invest more time on this, I cannot right figure out like this how to do. The recursive field technique that you used with conditional matching on `extends object` looks good to me. I think that we can invest more time in the future if we feel that editing the types is too painful.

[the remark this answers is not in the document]
bprusinowski: Sure, makes sense! I will leave it without changes at this point then.

app/configurator/configurator-state.tsx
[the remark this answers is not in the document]
ptbrowne: Yes I think so :)

PR #509, opened 2022-04-14
Small improvements

ptbrowne:
Simple things that I noticed while working on other things.

- feat: Only show top 7 items in nav (design alignment)
- feat: Add footer to search results page
- feat: Add test to search with diacritics


bprusinowski (approved):
Overall LGTM, but the closing animation isn't very smooth, there is clear "bump" at the end – maybe this could be improved somehow?

https://user-images.githubusercontent.com/52032047/163436182-c5aefee1-3882-470f-ba7b-97efaae54f92.mov

ptbrowne:
Very good point @bprusinowski, thanks for pointing it out. Now I use Reorder components and it works better.

PR #512, opened 2022-04-19
Fix scatterplot crash

ptbrowne:
- fix: Prevent scatterplot chart from crashing when data segment value does not have an IRI
- fix: Prevent margins due to empty interactive filters component
- refactor: Typo in type


bprusinowski (approved):
💯

PR #514, opened 2022-04-21
Visualization layer under label layer

ptbrowne:
- fix: Missing forward ref lead to console error
- fix: Missing key property lead to React warning
- fix: Warning in theme (spacing should use pixels instead of rem)
- feat: Unexpected error has correct severity, draft warning is in yellow
- feat: Add error message to Unexpected error to make it easier to debug
- fix: Missing key
- wip


bprusinowski (approved):
LGTM, it's really cool to be able to control the order of layers 💯

app/charts/map/map.tsx
```diff
@@ -204,17 +206,30 @@ export const MapComponent = () => {
...
-      features: features.areaLayer?.shapes?.features.filter(
+    // Sort for smaller shapes to be over larger ones, to be able to use tooltip
+    const sortedFeatures = orderBy(
+      features.areaLayer?.shapes?.features.filter(
         ({ properties: { hierarchyLevel } }: GeoFeature) =>
           hierarchyLevel === areaLayer.hierarchyLevel
       ),
+      (feature: GeoFeature) => {
+        return area(feature);
+      },
```
bprusinowski, line 234: Nit: Maybe [d3.geoArea](https://github.com/d3/d3-geo#geoArea) could be used here so we don't need to install another package?

PR #519, opened 2022-04-26
Sort imports

ptbrowne:
- chore: Configure eslint to sort and group imports
- chore: Run yarn lint --fix

Eslint is used to sort imports as the import/order plugin can be configured
and it does not seem like Organize imports from Vscode can.

`@/` imports are correctly recognized as "internal" and not from node_modules.

Here is the relevant VSCode config:

```
  "editor.codeActionsOnSave": {
    "source.fixAll.eslint": true
  }
```

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,10 +1,22 @@
 {
   "extends": "next",
-  "plugins": ["unused-imports"],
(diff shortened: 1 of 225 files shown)


bprusinowski (approved):
Great, thanks! 💯

PR #522, opened 2022-04-26
feat: Add chart type selection tabs UI

bprusinowski:
This PR:
- adds a possibility of switching chart type by using popover menu
- adds basic tabs logic (which would need to be enhanced when chart composition is implemented)
- aligns design of ChartTypeSelector to match the one from mockups


app/components/chart-selection-tabs.tsx
```diff
@@ -0,0 +1,201 @@
...
+    <Flex sx={{ width: "100%", overflowX: "auto" }}>
+      <Flex sx={{ gap: 1 }}>
+        {/* TODO: Generate dynamically when chart composition is implemeneted */}
+        <Tab
+          iconName={getIconName(chartType)}
+          selected
+          editable={editable}
+          onActionButtonClick={onActionButtonClick}
+        />
+        <Tab iconName="add" disabled narrow />
```
ptbrowne, line 133: Maybe we should add a tooltip (MuiTooltip) here "Soon you will be able to add multiple charts to a visualisation ?"

app/components/chart-selection-tabs.tsx
```diff
@@ -0,0 +1,201 @@
...
+        borderTopLeftRadius: 12,
+        borderTopRightRadius: 12,
+
+        boxShadow: 6,
+      }}
+    >
+      {children}
+    </ButtonBase>
+  );
+};
```
ptbrowne, line 201: Have you seen the Tabs component from MUI ? I think it would be nice to use them and theme them in the federal theme.

app/components/chart-selection-tabs.tsx
AnninaWalker: I wouldn't add a tooltip. As long as we are still in the conceptional phase with the publish a composition feature I don't want to hint on it .

ptbrowne: At the end of the sprint, we should we publish to production right ? In this case either we should remove the Add button or add a tooltip no ? Otherwise we have a disabled button it would be a bit weird no ?

AnninaWalker: Ah I thought I have agreed with Bartosz that we dont show the Add Button. This should be removed

bprusinowski: I think I lost this along the way - will hide the button :)

AnninaWalker: No worries ;-)

CHANGELOG.md
```diff
@@ -13,7 +13,9 @@ You can also check the [release page](https://github.com/visualize-admin/visuali
 - Fixed scrolling of table preview & other smaller UI things in published charts
 - Only the top 7 nav items are shown in the search nav bar (design alignment), a button is there to display more
 - Added the footer in the search page
-- Maps now have labels and the visualization layers (scatterplot and areas) are shown under map labels. When zooming via the buttons, we now have a smooth transition.
+- Maps now have labels and the visualization layers (scatterplot and areas) are shown under map labels. When zooming via the buttons, we now have a smooth transition
+- Removed chart selection step; selection of a chart type can now be done at any point when creating a chart (added a new chart selection tabs UI)
+- Implemented a logic to keep as much chart cofiguration as possible when switching a chart type
```
ptbrowne, line 18:
```suggestion
- Implemented logic to keep as much chart configuration as possible when switching chart type
```

bprusinowski: Thanks 🙇

app/docs/chart-selection-tabs.docs.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+import { ChartSelectionTabs } from "@/components/chart-selection-tabs";
+import { ConfiguratorStateConfiguringChart } from "@/configurator";
+import { ConfiguratorStateProvider } from "@/src";
+import palmerPenguinsFixture from "@/test/__fixtures/int/scatterplot-palmer-penguins.json";
+
+export default () => markdown`
+> Chart Selection Tabs are used in ChartPreview and ChartPublished to enable selecting a chart type in any stage of creating a chart.
+
+They can be either _editable_, to display a button to show ChartTypeSelector (used when creating a chart) or not, to disable that functionality (used in published charts).
```
ptbrowne, line 13: Ah yes I was wondering what was the use of the editable prop.

Nit: It could be nice to have it as prop documentation

```
/** Tabs are not editable when they are published **/
editable: boolean
```

so that you get autocomplete documentation.

This is not important in this PR, just for the future, I think it's good to take the habit to document props when they are not obvious (I say this also for me ;)).

ptbrowne (approved):
LGTM nice !

bprusinowski:
Thanks @Rdataflow! Indeed the whole new switch chart feature makes the tool more user-friendly by allowing the user to experiment with different chart types without losing all of the chart config; and new UI also makes it more fun to do so :)

PR #525, opened 2022-04-27
feat: Tabs UI refinements

### LINKED ISSUE #515: Switch visualization: Switch chart type via tab (1b)
**Problem statement**
Currently an editor looses all the filter settings when changing to another chart type.

**Expected behaviour**

The editor should be able – while visualizing a chart – to switch to another visualization type anytime, while keeping the filter and configuration settings, and without having to go back one step to select another chart type.

**Concept**

Instead of having to choose a chart type first, the user will directly land on the editing page, after selecting a dataset. Per default, the chart type that makes the most sense for the selected dataset is shown (in this example, it is the bar chart).

See Design Mockups and Documentation [here](https://invis.io/9811YXLK6VU5#/464464049_presentation-Switch-Visualization-6-01)

Dependency: work can begin as soon as we receive approval of the "publish composition" designDerived from the following [issue](https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/299)

**Additional Context**
Derived from the following [issue](https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/299)
Issue on Gitlab: https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/348

`+ button` not needed for this implementation as it is needed for the composition

bprusinowski:
Closes #515.

Small adjustments to the tabs UI based on conversation in #515.

### DIFF
--- a/app/components/chart-selection-tabs.tsx
+++ b/app/components/chart-selection-tabs.tsx
@@ -3,7 +3,9 @@ import React, {
   createContext,
   Dispatch,
   ReactNode,
+  useCallback,
   useContext,
+  useEffect,
   useState,
 } from "react";
 
@@ -78,6 +80,14 @@ const TabsEditable = ({ chartType }: { chartType: ChartType }) => {
   const [tabsState, setTabsState] = useTabsState();
   const [popoverAnchorEl, setPopoverAnchorEl] =
     useState<HTMLButtonElement | null>(null);
+  const handleClose = useCallback(() => {
+    setPopoverAnchorEl(null);
+    setTabsState({ isPopoverOpen: false });
+  }, [setTabsState]);
+
+  useEffect(() => {
+    handleClose();
+  }, [handleClose, configuratorState.chartConfig.chartType]);
 
   return (
     <>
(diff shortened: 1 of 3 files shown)


app/configurator/components/chart-type-selector.tsx
```diff
@@ -192,6 +185,21 @@ export const ChartTypeSelector = ({
...
+            sx={{
+              position: "absolute",
+              bottom: 48,
+              right: 48,
+              color: "hint.main",
+              display: "inline-block",
+              ml: 1,
+            }}
+          />
+        ) : null}
```
ptbrowne, line 202: I wonder in fact why we have the spinner for possible filters fetching here ? It's already shown in the fitlers control section no ?

ptbrowne (commented):
LGTM, only a small question for the spinner, but since this was already there, I am fine with the proposed solution.

ptbrowne:
@bprusinowski I had a small question on the spinner so I did not approve so as not to trigger the auto merge, but this is LGTM if you think we should keep the spinner.

app/components/chart-selection-tabs.tsx
```diff
@@ -78,6 +80,14 @@ const TabsEditable = ({ chartType }: { chartType: ChartType }) => {
   const [tabsState, setTabsState] = useTabsState();
   const [popoverAnchorEl, setPopoverAnchorEl] =
     useState<HTMLButtonElement | null>(null);
+  const handleClose = useCallback(() => {
+    setPopoverAnchorEl(null);
+    setTabsState({ isPopoverOpen: false });
```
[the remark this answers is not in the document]
ptbrowne: Fine by me.

app/configurator/components/chart-type-selector.tsx
bprusinowski: Done!

ptbrowne (commented):
LGTM

PR #529, opened 2022-04-28
feat: Include tables in chart config retaining logic

bprusinowski:
This PR introduces segment retaining when switching from and to a table chart.

**Non table chart with segment -> Table chart:** _Segment is used to choose a grouping column_
**Table chart with at least one grouping column -> Non table chart that supports segment:** _First grouping column is used as a segment (if possible, in tables it's possible to use TemporalDimensions or Measures as grouping columns, which is not permitted in segments)_

### DIFF
--- a/app/charts/index.ts
+++ b/app/charts/index.ts
@@ -259,26 +259,26 @@ export const getInitialConfig = ({
 
 export const getChartConfigAdjustedToChartType = ({
   chartConfig,
-  chartType,
+  newChartType,
   dimensions,
   measures,
 }: {
   chartConfig: ChartConfig;
-  chartType: ChartType;
+  newChartType: ChartType;
   dimensions: DataCubeMetadata["dimensions"];
   measures: DataCubeMetadata["measures"];
 }): ChartConfig => {
+  const oldChartType = chartConfig.chartType;
   const initialConfig = getInitialConfig({
-    chartType,
+    chartType: newChartType,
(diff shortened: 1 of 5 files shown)


app/charts/index.ts
```diff
@@ -639,37 +639,96 @@ const chartConfigsAdjusters: ChartConfigsAdjusters = {
...
+    },
+    area: {
+      "fields.segment": "fields",
+    },
+    scatterplot: {
+      "fields.segment": "fields",
+    },
+    pie: {
+      "fields.segment": "fields",
+    },
```
ptbrowne, line 845: Nit: wondering if there should be a `default` destination chart so that we can remove some duplication here.

ptbrowne (commented):
It'd be good to have some tests around the behavior since the behavior is complex.

app/charts/index.ts
bprusinowski: Hmmm, I am not sure about that, as we exclude Map chart in this particular case, so we still would need to define either exclusions (Map) or inclusions (all expect Map). In fact, no override field is defined for all chart types, so not sure is there should be a default one to fallback to, as in such cases it should return undefined to proceed normally. What do you think?

ptbrowne: OK I agree, thanks.

ptbrowne (approved):

PR #530, opened 2022-04-28
Published chart responsiveness

ptbrowne:
- fix: On mobile chart published do not overflow
- fix: When downsizing the window, a published chart will resize correctly
- fix: Correctly display chart actions on mobile


bprusinowski (approved):
LGTM, nice!

PR #531, opened 2022-04-28
feat: Cypress makes screenshots, then we use argos CI to do comparison

ptbrowne:
Previously, we'd use Cypress to make screenshots and compare them to a screenshot saved in the repository.
This tactic is cumbersome since we have to remember to update the screenshots when we make a change
and it is cumbersome to see the screenshots since we have to download the artifacts from the github workflow
to see the problem.
Here, Cypress takes the screenshots and saves them to the `cypress/screenshots` folder. This folder is then uploaded to argos platform where they will be compared to the screenshots coming from the main branch.
Then if Argos founds an error, it will be reported on the PR as a Github check. The interesting part is that we  can then go to the failing check, see directly the screenshots in a web interface, and accept the changes if they correspond to something that we really want. This makes the workflow smoother.

See the build here for this branch: https://www.argos-ci.com/visualize-admin/visualization-tool/builds/4

I do not know why right know we have no comparison screenshots, I guess that when we merge it should be OK 🤔


bprusinowski (approved):
Wow, looks great! So for each PR new screenshots will be compared with these from the main branch?

ptbrowne:
Yep and there should be a check on Github.

image

Then you can click on details and see something like this : https://www.argos-ci.com/cozy/cozy-ui/builds/259 where the 3 images are side by side (original from main/master, your branch, and the diff).

PR #535, opened 2022-04-28
feat: Cache node_modules and .next/cache folder

ptbrowne:
I saw a warning during build from next telling we should cache things.
Apparently we did not cache node_modules.
Using a cache should speed things up.

* I upgrade setup-node action and tells to cache things related to yarn.
* I also cache .next/cache and the cache key depends on yarn.lock and source files


bprusinowski (approved):
💯

Issue #556, opened 2022-05-18
Problem with Data Type "mm.yyyy"


ortnever:
Here is the bug I mentioned yesterday during the weekly call. It would be great if it could be fixed so that this dataset could be published on PRO soon.
This dataset contains a time dimension of type mm.yyyy. Zazuko has added this data type in the Creator cube. The transformation works and the values of this dimension appear correctly in the Designer cube (I added screen shot from CC below).
On Visualize the date (time reference) is displayed correctly in the Bar chart and line charts but is not displayed correctly in the first table and in the table chart: sometimes just yyyy, sometimes dd.mm.yyyy. 

**To Reproduce**
1. Go to 'test.visualize.admin.ch
2. Search "Air pollution along the A2 and A13"
3. See Value in "Temporal Reference"
![image](https://user-images.githubusercontent.com/79834090/168993089-c1fc0f9d-4a9e-46b0-a6a6-4b343c8d5e79.png)
![image](https://user-images.githubusercontent.com/79834090/169005637-8c5359a6-dc8a-4fb6-ba08-fca902ffb812.png)
![image](https://user-images.githubusercontent.com/79834090/168994822-66a01660-5683-44a4-a4a6-fdd8783bf3b0.png)

**Expected behavior**
the dimension "time reference" must appear everywhere in the format mm.yyyy in Visualize.

**Screenshots**
Screenshots from Cube Creator INT
![image](https://user-images.githubusercontent.com/79834090/168995526-3094fe32-19ab-432d-87d3-9baeaf1c6279.png)
![image](https://user-images.githubusercontent.com/79834090/168995361-a5274fc4-ffad-42bf-926c-5cc55cf12b7b.png)
![image](https://user-images.githubusercontent.com/79834090/168994123-17f23924-b803-4259-9ff2-a028736701b5.png)

**Environment (please complete the following information):**
 - Visualize TEST v3.6.4 ([2b571b5]
 - Browser : edge

Issue #498, opened 2022-04-11
Year filter do not show all available years - problem in the data ?


FabianCretton:
**Describe the bug**
Dataset "Öffentliche Ausgaben TEST 2" has observations from 1995 to 2020, every year.
When preparing a "table" visualisation, all years are visible, but when preparing a column graph for instance, the Filter only displays 1995 and 2020 (and not 1996, 1997....2019)

Here is a final visualization with the year filter: https://test.visualize.admin.ch/v/UNdsaAoMsc4o

How to find out where the problem is, is there a problem in the data ?
I made a first check, and did not find any error easily.

For information: the original cube has two measures: in % and In Mio. CHF (CHF) -> in the current published version, version 3, I did remove the "in %" measure, suspecting a decimal value to be wrong. But keeping only the "integer" value of "In Mio. Chf" did not help.

**Environment (please complete the following information):**
 - Visualize environment and version: TEST v3.4.8 ([2a8c061]
 - Browser and version: Firefox 99.0

AFoletti:
I have the exact same issue for the "Einmalvergütung für Photovoltaikanlagen" dataset

bprusinowski:
Hi @FabianCretton and @AFoletti!

Thanks for raising this issue – I checked both mentioned cubes and they are missing a dataKind property on a Year dimension, which has to be set on a TemporalDimension in Step 3 in Cube Creator in order to be correctly interpreted by Visualize (just setting a data type of a year is not enough).

@ortnever this is also the case for Heavy Metal Soil Contamination dataset on PROD.

Please let me know if this solves the issue :)

FabianCretton:
Thank you for your input @bprusinowski 
I can add the dataKind (I did already but did not publish yet). 
However, I do not understand what is happening, as, from my understanding, those values could be anything else then dates and could have no dataKind actually, no ?
So, why are there only two values present in the filter, this seems strange, no ?

FabianCretton:
For information, the new cube publication, with the dataKind, is working and displays all the years: https://test.visualize.admin.ch/v/Ot9GV0D1_ONb

I am still interested for an answer about what was happening:
However, I do not understand what is happening, as, from my understanding, those values could be anything else then dates and could have no dataKind actually, no ?
So, why are there only two values present in the filter, this seems strange, no ?

AFoletti:
I would have liked to test myself, but after the change I could not publish the data anymore, so I can't confirm.
[...]

PR #559, opened 2022-05-19
Cell value formatters

ptbrowne:
Fixes https://github.com/visualize-admin/visualization-tool/issues/556

To allow for proper formatting of values inside cells, we instantiate
dimension values formatters at a high level (parent table) instead
of relying on useFormatDateAuto. The metadata of the column (dimension.timeUnit
to know which format to apply, dimension.timeFormat to parse) is used
to create the proper formatter.

### DIFF
--- a/app/configurator/components/ui-helpers.ts
+++ b/app/configurator/components/ui-helpers.ts
@@ -49,7 +49,11 @@ import { useMemo } from "react";
 
 import { ChartProps } from "../../charts/shared/use-chart-state";
 import { Observation } from "../../domain/data";
-import { DimensionMetaDataFragment, TimeUnit } from "../../graphql/query-hooks";
+import {
+  DimensionMetaDataFragment,
+  TemporalDimension,
+  TimeUnit,
+} from "../../graphql/query-hooks";
 import { IconName } from "../../icons";
 import {
   getD3FormatLocale,
@@ -62,6 +66,8 @@ import {
   TableColumn,
   TableFields,
 } from "../config-types";
+
+export type DateFormatter = (d: string | Date | null) => string;
 
 // FIXME: We should cover more time format
 const parseSecond = timeParse("%Y-%m-%dT%H:%M:%S");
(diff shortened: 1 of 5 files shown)


app/configurator/components/ui-helpers.ts
```diff
@@ -91,16 +97,71 @@ const getFormattersForLocale = memoize((locale) => {
     minute: format("%d.%m.%Y %H:%M"),
     hour: format("%d.%m.%Y %H:%M"),
     day: format("%d.%m.%Y"),
-    month: format("%d.%m.%Y"),
+    month: format("%m.%Y"),
```
bprusinowski, line 100: 👍

bprusinowski (approved):
LGTM!

ptbrowne:
I've added a test for useDimensionFormatters. Can you please take a quick look ?

ptbrowne:
Oups, I've amended my previous commit, please check ui-helpers.spec.ts ;)

app/configurator/components/ui-helpers.ts
```diff
@@ -91,16 +108,74 @@ const getFormattersForLocale = memoize((locale) => {
...
+const dateFormatterFromDimension = (
+  dim: TemporalDimension,
+  localFormatters: LocalDateFormatters,
+  formatDateAuto: (d: Date | string | null) => string
+) => {
+  if (
+    dim.timeFormat &&
+    dim.timeUnit &&
+    localFormatters[dim.timeUnit.toLowerCase() as keyof typeof localFormatters]
+  ) {
```
bprusinowski, line 130: After taking a second look at this, do we even use timeFormat defined in CC somewhere? Here it looks we only check if it was defined, but is it taken into consideration when formatting the dates?

ptbrowne: Not that I know of, since we use d3 for parsing, we'd have to convert the parsing spec from CC format to d3 format. For now we only rely on the timeUnit. Let's see if we encounter cases where the timeFormat has a different time format than what's defined based on the timeUnit.

@l00mi we currently use the below code to know how to parse dates coming from a cube ? Can we rely on this ?

```
const parseSecond = timeParse("%Y-%m-%dT%H:%M:%S");
const parseMinute = timeParse("%Y-%m-%dT%H:%M");
const parseDay = timeParse("%Y-%m-%d");
const parseMonth = timeParse("%Y-%m");
const parseYear = timeParse("%Y");

const dateParsers = {
  Year: parseYear,
  Month: parseMonth,
  Day: parseDay,
  Minute: parseMinute,
  Second: parseSecond,
};

// When parsing observations
const dateParser = dateParsers[dimension.timeUnit]
```

bprusinowski: I wonder then why we throw a fatal error when timeFormat is not defined in CC, when we do not make use of it (I know it's not in the scope of this PR, but thinking out loud since the topics are connected). I think that either then we should take it into consideration (if @l00mi agrees) or remove the constraint of having non-empty timeFormat  – yesterday Veronique couldn't load a cube because of it, despite having other fields filled up correctly.

bprusinowski: (it's defined as mandatory field in GQL)

ptbrowne: Oh I made a mistake.

1. timeFormat is returned by us, from the resolvers, based on the data type of the dimension.

```
const timeFormats = new Map<string, string>([
  [ns.xsd.gYear.value, "%Y"],
  [ns.xsd.gYearMonth.value, "%Y-%m"],
  [ns.xsd.date.value, "%Y-%m-%d"],
  [ns.xsd.dateTime.value, "%Y-%m-%dT%H:%M:%S"],
]);
```

2. This way I do not need the conversion table and can directly us the timeFormat to build the d3 parser.

---

For Veronique's problem, it's not clear for me why the time format was not returned, based on this information @bprusinowski what do you think ?

bprusinowski: Cool – I also forgot it's done like that, I had the assumption it is fetched from CC 😆 I re-visited Veronique's problem and [it was related to missing dataKind](https://github.com/visualize-admin/visualization-tool/issues/498#issuecomment-1129790684) so all good from my side!

l00mi: Sorry, missed this, but yes it is all based on [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601). 

It should be as simple as `const date1 = new Date(isoStr1);` see also https://bobbyhadz.com/blog/javascript-convert-iso-string-to-date-object.

l00mi: Ah you mean the ticks, yes that is what we propose, but we can never be sure that this are the only once used by others. So best have a fallback to one unit in case it is not "known".
Screenshot 2022-05-19 at 15 20 42

PR #560, opened 2022-05-19
fix: Infinite loop when hovering over chart

bprusinowski:
Although it probably wasn't noticeable by the end user (but surely by developers as lots of console errors), we had an infinite loop running when hovering over a chart (even when the mouse stopped moving). It was due to useState + useEffect without an empty dependency array. In addition, this PR improves performance of finding closest value for horizontal tooltip by only looking for these when X values changes (and not on every mouse move as previously).


ptbrowne:
💯 Thanks !!

PR #568, opened 2022-05-23
fix: Scroll jumping problem when switching from table to chat view in published chart

ptbrowne:
Wrap both chart and table with a box that takes the height of the chart


bprusinowski (approved):
👌

Issue #564, opened 2022-05-20
Problems with the filter on a "yyyy-mm" time reference


ortnever:
I have on test.visualize a new cube with a time reference of type yyyy-mm. In the creator cube I defined the scale of measure Intervall and the Data kind Time description + Time Precison : month. It seems to me that this is correct?
However the filter on the time reference does not work correctly in Visualize. I do not see all the dates.

**To Reproduce**
1. Go to 'test.visualize.admin.ch
2. Select German navigation (DE)
3. Search "Luftbelastung" 
4. Open dataset "Luftbelastung entlang der A2 und der A13 - Monatliche Mittelwerte - NO und NO2"
5. If you let "zeitliche Referenz" on the horizontal axis, the filter on the right side is displayed correctly. The only small thing is that the beginning of the scale is 2011 instead of 01.2011, but this is a detail.
![image](https://user-images.githubusercontent.com/79834090/169503157-793b3bfe-74e0-4c09-ba63-4a788f681693.png)
6. Now choose another dimension on the horizontal axis, for example "Schadstoffe".
The dimension "time reference" is displayed in the filters on the left, but you can only choose 2 values: 01.2011 and 02.2011
![image](https://user-images.githubusercontent.com/79834090/169500490-a5078f76-fe49-4c42-a76c-20dd7228eb68.png)

8. Change the type graphic on map
9. Same problem occures

**Expected behavior**
all available values of the dimension "zeitliche referenz" should be displayed in the drop down list.

**Screenshots**

**Environment (please complete the following information):**
 - test.visualize.admin.ch, v3.7.0 [d8c018a]
 - edge

**Additional context**

ptbrowne:
It seems like the dataset need to be republished @ortnever, I cannot find it.

The problem seems similar to https://github.com/visualize-admin/visualization-tool/issues/498, I will take a look when the dataset is republished.

ortnever:
Sorry @ptbrowne . It's published as Draft.
![image](https://user-images.githubusercontent.com/79834090/169773078-55ebbe95-480e-44a5-bc7d-548f9b6cfe4f.png)

ptbrowne:
@ortnever Please tell us if it's working for you. If not I will re-open.

ptbrowne:
It's still deploying.

ortnever:
@ptbrowne Yes, it works. Thank you very much ! 
There is just a little detail in the Time-scala on the right side, but not really important (2011 instead 01.2011)

[...]

PR #570, opened 2022-05-24
Show all monthly values inside filter

ptbrowne:
Fix https://github.com/visualize-admin/visualization-tool/issues/564

Previously, only the min / max values were shown inside a temporal filter with timeUnit='Month'. Now we use a select for this case.

You can test by starting from [this dataset](https://visualization-tool-git-filter-month-select-ixt1.vercel.app/browse/dataset/https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2FUBD010705%2F3?previous=%7B%22includeDrafts%22%3Atrue%2C%22order%22%3A%22SCORE%22%2C%22search%22%3A%22Luftbelastung%22%7D) and selecting a x-axis different from the date, so that the date is available in the filters.

### DIFF
--- a/app/rdf/queries.ts
+++ b/app/rdf/queries.ts
@@ -261,7 +261,8 @@ export const getCubeDimensionValues = async (
   if (
     typeof dimension.minInclusive !== "undefined" &&
     typeof dimension.maxInclusive !== "undefined" &&
-    data.timeUnit !== "Day"
+    data.timeUnit !== "Day" &&
+    data.timeUnit !== "Month"
   ) {
     const min = parseObservationValue({ value: dimension.minInclusive }) ?? 0;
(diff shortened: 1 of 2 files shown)


app/rdf/queries.ts
```diff
@@ -258,10 +258,12 @@ export const getCubeDimensionValues = async (
   filters?: Filters
 ): Promise<DimensionValue[]> => {
   const { dimension, cube, locale, data } = rdimension;
+  console.log("TU", data.timeUnit);
```
bprusinowski, line 261: 👀

bprusinowski (approved):
Looks good! 👍

app/rdf/queries.ts
ptbrowne: 👍

PR #581, opened 2022-06-09
Support for hierarchies

ptbrowne:
Following the work that had been done to display hierarchical data inside the filters, the hierarchies are now fetched
through the data of the cube.

Only level 1 and level 2 are fetched at the moment until we have datasets with more levels.
Since hierarchical values are referenced through named nodes inside the observations, it was also necessary to change formatters for labels so that they get the label from the dimension values. This is why we have to keep an index of values by IRI in various states that use segments (to display correctly the labels in the legend) for example.

I expect there will be more work on this to support hierarchical values as part of the dynamic filters and to group values by parent inside the legend.

### DIFF
--- a/app/graphql/resolvers.spec.ts
+++ b/app/graphql/resolvers.spec.ts
@@ -8,7 +8,6 @@ import { unversionObservation as unversionObservation_ } from "../rdf/query-dime
 
 import { Query } from "./resolvers";
 
-
 const getCubeObservations = getCubeObservations_ as unknown as jest.Mock<
   typeof getCubeObservations_
 >;
@@ -29,6 +28,7 @@ jest.mock("../rdf/queries", () => ({
 }));
 
 jest.mock("../rdf/query-cube-metadata", () => ({}));
+jest.mock("../rdf/query-hierarchies", () => ({}));
 
 describe("possible filters", () => {
   beforeEach(() => {
--- a/app/graphql/resolvers.ts
+++ b/app/graphql/resolvers.ts
@@ -5,8 +5,6 @@ import { keyBy } from "lodash";
 import { topology } from "topojson-server";
 import { parse as parseWKT } from "wellknown";
 
-import { queryHierarchy } from "@/rdf/query-hierarchies";
-
 import { Filters } from "../configurator";
 import {
   DimensionValue,
@@ -35,6 +33,7 @@ import {
 } from "../rdf/query-cube-metadata";
 import { unversionObservation } from "../rdf/query-dimension-values";
 import { RawGeoShape } from "../rdf/query-geo-shapes";
+import { queryHierarchy } from "../rdf/query-hierarchies";
 import cachedWithTTL from "../utils/cached-with-ttl";
 import {
   makeCubeIndex as makeCubeIndexRaw,


app/locales/locales.ts
```diff
@@ -31,19 +33,12 @@ export const locales = ["de", "fr", "it", "en"] as const;
...
-  it: catalogIt,
-  en: catalogEn,
-});
-i18n.activate(defaultLocale);
+export const activate = async (locale: typeof locales[number]) => {
+  i18n.loadLocaleData(locale, localeData[locale]);
+  const messages = await import(`./${locale}/messages`);
+  i18n.load(locale, messages);
+  i18n.activate(defaultLocale);
+};
```
bprusinowski, line 41: Wouldn't this mean that the messages will now be imported each time when a user switches a locale?

app/pages/_app.tsx
```diff
@@ -86,3 +87,12 @@ export default function App({ Component, pageProps }: AppProps) {
     </>
   );
 }
+
+export const getServerSideProps: GetServerSideProps = async (ctx) => {
+  await activate(ctx.router.locale);
+  return {
+    props: {
+      test: "ok",
+    },
```
bprusinowski, line 96: 👀

app/rdf/query-hierarchies.ts
```diff
@@ -0,0 +1,88 @@
...
+  o: Record<string, any>
+): o is Record<string, string> & { label: string; value: string } => {
+  return o.label && o.value ? true : false;
+};
+
+export const queryHierarchy = async (
+  dimension: string
+): Promise<HierarchyValue[]> => {
+  const query = DESCRIBE`?hierarchy ?level1 ?level2`.WHERE`    
+      SELECT DISTINCT ?hierarchy ?level1 ?level2
```
bprusinowski, line 33: Why is there a DESCRIBE and then SELECT inside a query? Shouldn't this just be SELECT.DISTINCT`?hierarchy ?level1 ?level2...`? What is the difference when using this notation?

app/rdf/query-hierarchies.ts
```diff
@@ -0,0 +1,88 @@
...
+        ?level1 ${ns.schema.hasPart} ?level2 .
+    
+        filter(isiri(?level2))
+    
+        OPTIONAL { ?this ${ns.schema.name} ?order0 } 
+      }
+    `;
+  const stream = await query.execute(sparqlClientStream.query);
+  const dataset = await fromStream(rdf.dataset(), stream);
+  const cf = clownface({ dataset });
```
bprusinowski, line 51: It this way more efficient than the way we did this in the past (just query.execute and the parsing the data)? Or was there an other need to do so?

app/charts/table/table.tsx
```diff
@@ -45,6 +46,23 @@ const TableContentWrapper = forwardRef<HTMLDivElement, $FixMe>(
...
+      borderBottom: "1px solid",
+      borderBottomColor: "grey.400",
+      "&:first-of-type": {
+        borderTop: "1px solid",
+        borderTopColor: "grey.400",
+      },
+    },
+  };
+});
+
```
bprusinowski, line 65: 👍

bprusinowski (approved):
I couldn't check a preview deployment, but the code looks good 💯

app/locales/locales.ts
ptbrowne: I removed this, it was part of an experimentation to load locales asynchronously, but it shouldn't be in this PR. 

---

To answer your original question, I would think that the asynchronous `import` gets converted to a call that fetches from webpack's runtime cache, and if the module has not been loaded, it would load it across the network. I will check this :)

app/pages/_app.tsx
ptbrowne: Removed from this PR.

app/rdf/query-hierarchies.ts
ptbrowne: I used the same technique as did Tomas from Zazuko here.

First we select the IRIs we are interested in (inside the inner select), and then we describe them to get all the attributes. We end up with an in-memory graph full of interconnected objects, and we use clownface to navigate it and reconstruct the hierarchy.

I *think* we could have a way to get tabular data right away to end up with a table with columns (iri, label, parentIri) and then reconstruct the hierarchy tree, I have to try.

bprusinowski: I see, thanks for explanation 🙏

PR #585, opened 2022-06-13
Hierarchical filters only shows values appearing in data

ptbrowne:
Since a hierarchy can encompass values that are much wider than what's in the data, we need to prune the tree to remove leafs whose value do not appear in the data. To do this, we query for distinct values inside the dimension and filter the hierarchical tree accordingly.

### DIFF
--- a/app/rdf/query-hierarchies.ts
+++ b/app/rdf/query-hierarchies.ts
@@ -1,4 +1,4 @@
-import { DESCRIBE } from "@tpluscode/sparql-builder";
+import { DESCRIBE, SELECT } from "@tpluscode/sparql-builder";
 import clownface from "clownface";
 import { ascending } from "d3";
 import { uniqBy } from "lodash";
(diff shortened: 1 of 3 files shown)


app/rdf/query-hierarchies.ts
```diff
@@ -46,11 +56,14 @@ export const queryHierarchy = async (
...
       }
     `;
+
+  const dimensionValues = new Set(await queryDimensionValues(dimension));
   const stream = await query.execute(sparqlClientStream.query);
   const dataset = await fromStream(rdf.dataset(), stream);
   const cf = clownface({ dataset });
 
   const schemaName = ns.schema.name as unknown as NamedNode<string>;
+  console.log(dimensionValues);
```
bprusinowski, line 66: 👀

app/rdf/query-hierarchies.ts
```diff
@@ -26,6 +26,16 @@ const hasValueAndLabel = (
   return o.label && o.value ? true : false;
```
bprusinowski, line 26: I know it was not a part of this PR, but wouldn't just o.label && o.value work the same? 😅

bprusinowski (approved):
LGTM!

app/rdf/query-hierarchies.ts
ptbrowne: Actually only the o.label && o.value would return a string instead of a boolean ;) I replaced by !!(o.value && o.label). Thanks :)

PR #586, opened 2022-06-15
Cube checker

ptbrowne:
- feat: Add a way to check the cube health
- feat: When redirecting to home page due to bad cube, show an error message
- feat: Add translations

When cannot load a cube, we have a message on home page:

image

We can see the cube in a new cube checker page that checks some of the attributes of the cube to see if they are correct.

## Bad cube 

image

## Good cube

image


bprusinowski (approved):
LGTM, great addition 👌 

Just an extra minor thing, when you change the language on the _cube-checker page, the content doesn't stay in place but moves – maybe it could be good to make the width of the container fixed?

PR #592, opened 2022-07-14
fix: Screenshots + pie and scatterplot legend fixes 

ptbrowne:
- Pie state and scatterplot state lacked a function in their state to support new hierarchy changes
- Screenshots were no longer made after the database cleanup


.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "tabWidth": 2,
+  "useTabs": false
+}
```
bprusinowski, line 4: 🚀

bprusinowski (approved):
LGTM!

PR #594, opened 2022-07-15
fix: When looking for versioned/unversioned cube through IRI, remove draft filter

ptbrowne:
When coming from the cube creator directly, we want to be able to see a versioned/unversioned iri even if they are in draft status.


bprusinowski (approved):
👍

PR #595, opened 2022-07-19
Improve behavior of hierarchy checkboxes

ptbrowne:
Previously, when toggling a parent node, we would toggle only the direct children.
And when a node had no children toggled, the checkbox was shown as unchecked.
This caused problems when using a hierarchy with 3 levels like the Öffentliche
ausgaben dataset from @fabiancretton.

Here we introduce a new centralized state for those checkboxes and a new behavior
with 3 states ("mixed" / "checked" / "unchecked").

The behavior is still not completely satisfactory in our opinion and maybe not
self explanatory enough. We will have to work again on it with @AnninaWalker  but
the behavior now is better than before, especially for hierarchy with 3 levels.

- fix: When looking for versioned/unversioned cube through IRI, remove draft filter
- chore: Remove console log
- feat: Add checkbox indeterminate to the federal theme
- feat: Better behavior for hierarchy in filters toggling
- test; Add E2E test for hierarchy filters in chart edition


bprusinowski (approved):
LGTM, works like a charm ✨ There is just one thing – when testing the feature using Bathing water quality (with Hierarchie) dataset, nothing changes when checking for example "BAQUA_FR" (from an initial state to a checked state) apart from the number of selected filters. Will this depend on a dataset (that the "whole" value is then added to a chart)? Or is this an unexpected behavior? Because in such case maybe it would make sense to "omit" the initial state and display the checkboxes as toggled?

ptbrowne:
MMh in this case, I think when querying for the hierarchy I need to return whether the value is present in the dataset or not. If it's not present, I think we can omit the "indeterminate" state. I will add this to the PR, I had the plan to keep it for later, but I think it's good to have in this PR.

bprusinowski (approved):
🎉

PR #605, opened 2022-07-26
Add handling of different sources (SPARQL, SQL) along with different URLs

### LINKED ISSUE #196: Add SPARQL endpoint to config
As discussed offline with @ptbrowne, we should probably add the SPARQL endpoint of a chart to the configuration, so in the future we could target different endpoints from any Visualize env. The main goal would be to allow users to preview data from any endpoint, e.g. Lindas INT cubes on Visualize PROD, or to test new Visualize features with production data (by using Visualize INT with Lindas PROD). By storing the endpoint in the config, the charts will always use the same data source once published, and the config will be more portable regardless of environment configuration for testing purposes and reproducing bugs.

Each Visualize env should still have a default endpoint (via SPARQL_ENDPOINT config var), and charts should display a message if their endpoint differs from the default one.

If we do this, we need to make sure that there's a fallback solution for configs that don't have this property yet.

Gitlab Issue: 
- https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/320

Related discussion : https://github.com/visualize-admin/visualization-tool/discussions/598#discussioncomment-3196686

- [x] Support for endpoint property in all queries
- [x] When no endpoint property is passed, the default endpoint is used
- [x] Default endpoint is configurable via environment variable
- [x] Default endpoint is production
- [x] Endpoint can be selected via a setting in the UX (as a MVP, a gear icon in the top bar opening a small popup menu with a endpoint select would be good)
- [x] When a chart is saved, the endpoint is saved there
- [x] When a chart does not have a endpoint, default endpoint is used
- [ ] All existing charts are migrated to include their endpoint
- [x] A mention should be shown on existing charts when a non trusted endpoint is used

bprusinowski:
Closes #196.

TODO:
- [ ] chart configs migrations
- [x] disable the possibility of changing env type when configuring / viewing a published chart
- [x] add mention for unsafe endpoints
- [x] check cypress tests
- [x] change env variables on Vercel (and maybe somewhere else?) from SPARQL_ENDPOINT to ENDPOINT
- [x] use the data source defined in chart config to set a data source if viewing a published chart
- [x] what should happen when several charts that use different data sources are embedded on a page? right now we set the source in localStorage, it would then get overwritten


app/components/data-source-menu.tsx
```diff
@@ -100,8 +100,12 @@ export const DataSourceMenu = () => {
 
   return (
     <Flex sx={{ alignItems: "center", gap: 1 }}>
-      <Typography variant="h4">Datasource:</Typography>
-      <MiniSelect
+      <Label htmlFor="dataSourceSelect">
+        <Typography sx={{ fontWeight: "bold", color: "grey.900" }}>
+          Data source:
+        </Typography>
```
ptbrowne, line 88: To translate ?

app/components/form.tsx
```diff
@@ -41,7 +41,7 @@ export const Label = ({
     htmlFor={htmlFor}
     variant={smaller ? "caption" : "body1"}
     display="flex"
-    sx={{ mb: 1, color: "grey.600" }}
+    sx={{ color: "grey.600" }}
```
ptbrowne, line 44: Isn't color='textSecondary' sufficient ?

app/graphql/resolvers/utils.ts
```diff
@@ -0,0 +1,20 @@
...
+  const [type, url] = endpoint.split("+") as [DataSourceType, string];
+
+  return { type, url };
+};
+
+export const convertSourceToEndpoint = (source: DataSource): string => {
+  const { type, url } = source;
+
+  return `${type}+${url}`;
+};
```
ptbrowne, line 20: Nit on the naming, having stringifySource and parseSource  (or stringifyEndpoint parseEndpoint) would be a bit clearer and would remove the need to have both source and endpoint.

ptbrowne (approved):
👏

app/components/data-source-menu.tsx
bprusinowski: Yup 😄

app/components/form.tsx
bprusinowski: It is, good catch, thanks!

app/graphql/resolvers/utils.ts
bprusinowski: Good point, will adjust the logic 👌

app/pages/embed/[chartId].tsx
```diff
@@ -43,13 +43,14 @@ const EmbedPage = (props: PageProps) => {
   const {
     config: {
       key,
-      data: { dataSet, meta, chartConfig },
+      data: { dataSet, dataSource, meta, chartConfig },
```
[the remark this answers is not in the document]
Comment (bprusinowski) [reply app/pages/embed/[chartId].tsx:46]: 👍

app/components/data-source-menu.tsx
```diff
@@ -57,17 +59,17 @@ export const useDataSource = () => {
 export const DataSourceProvider = ({ children }: { children: ReactNode }) => {
   const [source, setSource] = useState<DataSource>(DEFAULT_DATA_SOURCE);
   const handleSourceChange = (source: DataSource) => {
-    localStorage.setItem("endpoint", convertSourceToEndpoint(source));
+    saveDataSourceToLocalStorage(stringifyDataSource(source));
```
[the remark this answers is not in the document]
bprusinowski: 👍

app/pages/api/graphql.ts
```diff
@@ -58,43 +75,22 @@ const server = new ApolloServer({
...
-      loaders: Loaders | undefined;
-      sparqlClient: ParsingClient | undefined;
-      sparqlClientStream: StreamClient | undefined;
+  context: async () => {
+    const ctx = {} as {
+      setupping: Promise<{
+        loaders: Loaders | undefined;
+        sparqlClient: ParsingClient | undefined;
+        sparqlClientStream: StreamClient | undefined;
+      }>;
```
ptbrowne, line 86: To avoid repetition, I guess we can use ReturnType<typeof setupContext> here.

app/components/chart-published.tsx
```diff
@@ -62,28 +61,27 @@ export const ChartPublished = ({
...
   configKey: string;
 }) => {
   const locale = useLocale();
-  const parsedDataSource = parseDataSource(dataSource);
-  const isDataSourceTrusted = React.useMemo(() => {
-    return DATA_SOURCE_OPTIONS.find((d) => d.value === dataSource)?.isTrusted;
-  }, [dataSource]);
+  // const isDataSourceTrusted = React.useMemo(() => {
+  //   return DATA_SOURCE_OPTIONS.find((d) => d.value === dataSource)?.isTrusted;
+  // }, [dataSource]);
```
ptbrowne, line 78: To remove 🧹

app/components/chart-published.tsx
```diff
@@ -142,15 +140,15 @@ export const ChartPublishedInner = ({
...
+        {/* {!isDataSourceTrusted && (
           <Box sx={{ mb: 4 }}>
             <HintBlue iconName="hintWarning">
               <Trans id="data.source.notTrusted">
                 This chart is not using a trusted data source.
               </Trans>
             </HintBlue>
           </Box>
-        )}
+        )} */}
```
ptbrowne, line 151: Why the comment ?

ptbrowne (approved):

app/components/chart-published.tsx
bprusinowski: Whoops, this shouldn't be commented out, thanks for catching! Same for the above comment.

PR #609, opened 2022-08-10
fix: Wrap published chart in Config provider

ptbrowne:
This way, the legend can use the chart config
Elements that do not need to change the chart config can use the
readonly version of the useReadOnlyConfiguratorState function


bprusinowski (approved):
LGTM, thanks!

PR #611, opened 2022-08-11
feat: Search directly with SPARQL

ptbrowne:
Instead of building a lunr index inside the application server and using it to search cubes, we use the full text capabilities of stardog.
A search debug page with interesting queries is there to quickly see if everything is working correctly.


ptbrowne:
Putting the work on pause while we search for an improvement on the performance of the search of the query, see discussion here : https://zulip.zazuko.com/#narrow/stream/32-bar-ld-ext/topic/full.20text.20search.20performance/near/281509

app/graphql/context.tsx
```diff
@@ -16,3 +17,5 @@ const client = createClient({
 export const GraphqlProvider = ({ children }: { children: ReactNode }) => {
   return <Provider value={client}>{children}</Provider>;
 };
+
+export type GraphQLContext = ReturnType<typeof createContext>;
```
bprusinowski, line 93: 💯

app/utils/search.ts
```diff
@@ -161,3 +176,234 @@ export const searchCubes = (
...
+  );
+
+  // Reduce to a single weighted score
+  const weights: Record<ScoreKey, number> = {
+    scoreName: 5,
+    scoreDescription: 2,
+    scoreTheme: 1,
+    scorePublisher: 1,
+    scoreCreator: 1,
+  };
```
bprusinowski: 👌

bprusinowski (approved):

PR #613, opened 2022-08-12
Fix color segmentation

ptbrowne:
- refactor: Remove unused props
- fix: Correct margin for labels
- fix: Interlace would fail if array length was 0
- refactor: Remove unneeded Box
- refactor: Use Select + MenuItem for color palette selector


bprusinowski (approved):
LGTM!

PR #615, opened 2022-08-12
feat: Do not hierarchy level section for maps as this is handled in filters

ptbrowne:
As hierarchy is now handled at dimension level and not anymore as part of the geoshapes, we do not need the hierarchy options for the maps.
Fixes were made on the multi filters display so that they handle cases without color mapping

### DIFF
--- a/app/charts/map/map-state.tsx
+++ b/app/charts/map/map-state.tsx
@@ -221,10 +221,9 @@ const useMapState = (
     [symbolErrorDimension?.unit, formatNumber, getSymbolError]
   );
 
-  const getDataByHierarchyLevel = useCallback(
+  const getData = useCallback(
     ({
       geoDimensionIri,
-      hierarchyLevel,
       getLabel,
     }: {
       geoDimensionIri: string;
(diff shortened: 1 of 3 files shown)


app/configurator/components/ui-helpers.ts
```diff
@@ -939,6 +909,44 @@ const divergingPaletteKeys = [
...
+export const getColorInterpolator = (
+  palette?: SequentialPaletteType | DivergingPaletteType
+): ((t: number) => string) => {
+  const interpolator = interpolatorByName[palette!] ?? defaultInterpolator;
+  // If the palette is sequential, we artificially clamp the value not to display too
+  // white a value
+  const isSequential = palette
+    ? // @ts-ignore
+      sequentialPaletteKeys.includes(palette)
+    : false;
```
bprusinowski, line 944: Nit: I wonder if here it wouldn't be enough to just do sequentialPaletteKeys.includes(palette), as even if it's undefined it should work the same way, no?

app/configurator/components/ui-helpers.ts
```diff
@@ -939,6 +909,44 @@ const divergingPaletteKeys = [
...
+  return isSequential
+    ? (n: number) => interpolator(n * 0.8 + 0.2)
+    : interpolator;
+};
```
bprusinowski, line 948: Was there a particular place in the app that led to this change?

bprusinowski (approved):
Not sure about starting the color interpolators from 0.2, but otherwise LGTM!

app/configurator/components/ui-helpers.ts
ptbrowne: Ah yes I added it because of typescript but then since there was the other error that types were incompatible between sequentialPaletteKeys ("blue" | "red" etc...) and palette (string), I added the typescript ignore.

app/configurator/components/ui-helpers.ts
ptbrowne: I saw a comment from @rdataflow at some point that said that the left side of the palette was too white. I am removing this commit from the PR and we can discuss it.

PR #616, opened 2022-08-12
Allow blocking of map controls

### LINKED ISSUE #440: Development: Block map controls to enhance user-friendlyness
The new worldwide map is great and opens up a tons of interesting possibilities. However, sometimes you just want to show a fixed region (say, Switzerland...) and save your user the hassle of getting back to the right place if he mistakenly scrolled or zoomed away.

At the publishing stage, I would like to have the option to completely block the possibility to pan and zoom, so that the map remains for the user fixed in the same configuration I set up.
This also means that during a window resize the map zoom must automatically adapt in order to keep the same region visible in the box.

At least the ability to block panning, if not zooming.

**Additional context**
None

bprusinowski:
Closes #440.

This PR introduces two radio buttons in the base layer settings in map charts which allow users to "block" and "unblock" map controls. When controls are blocked (`locked`), it's not possible to interact with the map (except for displaying map tooltip) and the viewport of the map is persisted during window resizes. Current behavior is used when controls aren't blocked (`dynamic`).

TODO:
- [x] translations,
- [x] design alignment,
- [x] saving of currentBBox to chart config on mode change (dynamic -> locked).


app/charts/map/layer.tsx
```diff
@@ -11,7 +11,7 @@
 import assert from "assert";
 
 import isEqual from "lodash/isEqual";
-import { Map } from "maplibre-gl";
+import { Map } from "mapbox-gl";
```
ptbrowne, line 14: Why ? Don't we use maplibre ?

app/charts/map/map-state.tsx
```diff
@@ -342,12 +342,22 @@ const useMapState = (
...
-    chartProps.baseLayer.bbox ||
-    getBBox(
-      areaLayer.show ? features.areaLayer?.shapes : undefined,
-      symbolLayer.show ? features.symbolLayer?.points : undefined
-    );
+  const bbox = useMemo(() => {
+    if (chartProps.baseLayer.bbox !== undefined) {
+      return chartProps.baseLayer.bbox;
+    } else {
+      return getBBox(
```
ptbrowne, line 349: Nit:  `return chartProps.baseLayer.bbox ?? getBBox(...)`

app/charts/map/ref.ts
```diff
@@ -0,0 +1,13 @@
+import { Map } from "mapbox-gl";
```
ptbrowne, line 1: I think it's helpful here to add a small comment to show that we are not so happy with the solution.

```
/**
 * We need to access the map from the map controls. Until we have a good solution for
 * sibling components, we use a global non observable ref.
 */
```

app/charts/map/map.tsx
```diff
@@ -114,21 +130,27 @@ export const MapComponent = () => {
...
+      ? BASE_VIEW_STATE
+      : getInitialViewState(bbox, bounds);
+  }, [isViewStateLocked, bbox, bounds]);
   const locale = useLocale();
 
   const [, dispatchInteraction] = useInteraction();
   const [, setMapTooltipType] = useMapTooltip();
 
-  const [viewState, setViewState] = useState(INITIAL_VIEW_STATE);
+  const [viewState, setViewState] = useState(initialViewState);
```
ptbrowne, line 164: We don't need a memo if we use the value only for the initial value of the state. Pass directly the useMemoed function to useState in this case.

app/locales/fr/messages.po
```diff
@@ -120,59 +120,67 @@ msgid "button.share"
...
 
-#: app/configurator/map/map-chart-options.tsx:170
-#: app/configurator/map/map-chart-options.tsx:394
+#: app/configurator/map/map-chart-options.tsx:109
+msgid "chart.map.layers.base.view.dynamic"
+msgstr "Vue dynamique"
+
+#: app/configurator/map/map-chart-options.tsx:118
+msgid "chart.map.layers.base.view.locked"
+msgstr "Vue verrouillée"
```
ptbrowne, line 163: The translation here is not immediately understandable in French. While trying to find a good other translation, I wondered if in fact we should not have "Locked map" with a [Switch](https://mui.com/material-ui/react-switch/#main-content). This way we do not have to find another word for the "dynamic" here (I find dynamic even in English a bit vague in this case). Since we have only two states, the Switch seems warranted.

app/charts/map/map.tsx
```diff
@@ -473,24 +501,26 @@ const useStyles = makeStyles<Theme>((theme) => ({
...
   return (
-    <Box className={classes.zoomButtons}>
-      <ZoomButton iconName="add" handleClick={zoomIn} />
-      <ZoomButton iconName="minus" handleClick={zoomOut} />
+      <MapControlButton iconName="circle" handleClick={refresh} />
+      <MapControlButton iconName="add" handleClick={zoomIn} />
+      <MapControlButton iconName="minus" handleClick={zoomOut} />
     </Box>
   );
 };
```
ptbrowne, line 530: Nit: This MapControlButtons is not very useful IMHO. Better to have the buttons directly in the parent, this way we do not have to drill props.

app/charts/map/map.tsx
```diff
@@ -175,6 +187,18 @@ export const MapComponent = () => {
     mapNodeRef.current = mapRef;
   };
 
+  const refresh = () => {
```
ptbrowne, line 202: refresh or reset ? Can you add a small comment here, it is not immediately clear to me what this should do.

app/charts/map/map.tsx
```diff
@@ -160,8 +161,19 @@ export const MapComponent = () => {
...
+    if (bbox) {
+      if (!isViewStateLocked) {
+        newViewState = constrainZoom(viewState, bbox);
+      } else {
+        newViewState = viewState;
+      }
+    }
+
+    if (newViewState) {
+      initialViewStateRef.current = newViewState;
```
ptbrowne: Since we update the ref here, it does not seem that "initial" is the right word here (I would expect that a variable named with initial would not change during the lifecycle of the component).

I find this useEffect a bit hard to understand, do you think it would be easier to understand if it was extracted with a good name ? What is the end goal of this useEffect ?

app/charts/map/map.tsx
```diff
@@ -513,7 +515,8 @@ const MapControlButtons = ({
   const classes = useStyles();
 
   return (
-      <MapControlButton iconName="circle" handleClick={refresh} />
+    <Box className={classes.mapControlButtons}>
```
ptbrowne, line 524: Nit: a `div` is sufficient here.

app/charts/map/map.tsx
```diff
@@ -187,13 +197,15 @@ export const MapComponent = () => {
...
   const refresh = () => {
-    if (initialViewStateRef.current) {
-      const { longitude, latitude, zoom } = initialViewStateRef.current;
+    if (firstViewStateRef.current) {
+      const { longitude, latitude, zoom } = firstViewStateRef.current;
       const newViewState = {
         center: [longitude, latitude] as LngLatLike,
         zoom,
-        duration: 500,
+        duration: DURATION,
```
ptbrowne, line 208: Nit: FLY_TO_DURATION

app/charts/map/map.tsx
```diff
@@ -133,31 +146,30 @@ export const MapComponent = () => {
     bounds,
     bbox,
   } = useChartState() as MapState;
-  const initialViewStateRef = useRef<MinMaxZoomViewState | null>(null);
+
   const isViewStateLocked = controlsType === "locked";
+  // Needed to be able to "refresh" the map to its initial position.
+  const firstViewStateRef = useRef<MinMaxZoomViewState | null>(null);
```
ptbrowne, line 152: Another ref, I really wonder if the logic behind the view states should be extracted in its own hook.

app/configurator/map/map-chart-options.tsx
```diff
@@ -104,7 +104,8 @@ const BaseLayerSettings = memo(() => {
           field={null}
           path="baseLayer.show"
         />
-        <Flex>
+        {/* FIXME: change the styles of MUI components to match design specs */}
```
ptbrowne, line 107: Which MUI components ?

ptbrowne (commented):
👍  Thanks.

- [ ] I'd like that we revisit the map control (see comment https://github.com/visualize-admin/visualization-tool/pull/616#discussion_r946436433), instead of two radio buttons, could we use a switch ? I think it would simplify since we would not have to find two different words ("dynamic" for me is a bit vague, while I immediately understand the contrary of "locked") @AnninaWalker 
- [ ] I think the main problematic point for me here is the view state logic and the refs. I think with those useEffects and useRef, that it is a bit hairy. At least I'd like to have tests on thi the behavior.

Otherwise LGTM, thanks !

app/locales/fr/messages.po
ptbrowne: I'd use "Carte verrouillée" in French.

app/charts/map/layer.tsx
bprusinowski: In fact the type of the map retrieved from the events (onLoad, onMove, etc) is the type from MapBox and they are not the same :) In fact in this particular place there are no TS errors, but to properly type things in other places (as mapRefs) it had to be a type from MapBox.

app/configurator/map/map-chart-options.tsx
bprusinowski: In the map control design file on Sketch, for example the Radio Button looks differently and also the margin between button and its label weren't matching the specs. Also I had some problems to align the buttons in this case, as for example the radio button has its own internal padding of 9px, while checkbox does not. If they are used just as <Checkbox /><RadioButton /> next to each other, the aren't aligned correctly. I think it would make sense to align this and also check other chart options components, as I think there are also some inconsistencies. I think it would make sense to include this in potential technical improvements.

ptbrowne: OK thanks for the explanation, can you add those tasks there : https://github.com/visualize-admin/visualization-tool/issues/620 ?

app/charts/map/map.tsx
bprusinowski: I'll refactor the code, thanks for suggestions 🙇

app/configurator/map/map-chart-options.tsx
bprusinowski: I will, yes!

app/charts/map/map.tsx
```diff
@@ -134,6 +130,43 @@ let globalScatterplotLayerId = 0;
 
 const FLY_TO_DURATION = 500;
 
+const useViewState = ({
```
ptbrowne, line 133: Can you add a docstring, I think it will help nail down the responsibilities of this hook.

bprusinowski: Sure!

app/charts/map/helpers.spec.ts
```diff
@@ -1,45 +1,108 @@
...
+      initialProps: {
+        width,
+        height,
+        lockedBBox: undefined,
+        featuresBBox: undefined,
+      },
     });
 
-    expect(viewState).toEqual(BASE_VIEW_STATE);
+    // If featuresBBox was not privded, defaultViewState should equal BASE_VIEW_STATE.
```
ptbrowne, line 36:
```suggestion
    // If featuresBBox was not provided, defaultViewState should equal BASE_VIEW_STATE.
```

app/charts/map/helpers.spec.ts
```diff
@@ -1,45 +1,108 @@
...
+    expect(result.current.defaultViewState.latitude).toEqual(
+      BASE_VIEW_STATE.latitude
+    );
+    expect(result.current.defaultViewState.longitude).toEqual(
+      BASE_VIEW_STATE.longitude
+    );
+    expect(result.current.defaultViewState.zoom).toEqual(BASE_VIEW_STATE.zoom);
```
ptbrowne, line 43: Nit: I think the 3 expect calls could be done via one toEqual call (expect(defaultVIewState).toEqual(BASE_VIEW_STATE)

app/charts/map/helpers.spec.ts
```diff
@@ -1,45 +1,108 @@
...
+    // different than BASE_VIEW_state.
+    expect(result.current.defaultViewState.latitude).not.toEqual(
+      BASE_VIEW_STATE.latitude
+    );
+    expect(result.current.defaultViewState.longitude).not.toEqual(
+      BASE_VIEW_STATE.longitude
+    );
+    expect(result.current.defaultViewState.zoom).not.toEqual(
+      BASE_VIEW_STATE.zoom
+    );
```
ptbrowne, line 57: Same here, 1 call would be sufficient.

app/charts/map/map.tsx
```diff
@@ -63,21 +66,31 @@ export const MapComponent = () => {
...
       const newViewState = {
         center: [longitude, latitude] as LngLatLike,
         zoom,
-        duration: FLY_TO_DURATION,
+        duration: RESET_DURATION,
       };
       mapNodeRef.current?.flyTo(newViewState);
     }
-  };
+  }, [defaultViewState]);
```
ptbrowne, line 88: I think useEvent would be better to use here. Then you would not need the dependencies array, it would make it clearer what is the value of reset.

And then it would make the useEffect below also clearer since the dependencies of the useEffect could be `[defaultViewState]` and then it's clear that "when the default view state changes, we reset", otherwise we have to think "reset changes when default view state changes, so the use effect will be called when default view state changes".

ptbrowne (approved):
Thanks it is already a lot more clearer with the lockViewState and the defaultViewState ! Some nits but LGTM.

app/charts/map/map.tsx
bprusinowski: Yes, I also wasn't happy with this useEffect hook, thanks for the suggestion!

ptbrowne (approved):
🎉 Thanks

Issue #538, opened 2022-05-03
choropleth color palette begin (white) misses contrast to background / basemap


Rdataflow:
**Is your feature request related to a problem? Please describe.**
choropleth color palette begin (white) misses contrast to background / basemap

**Describe the solution you'd like**
A color palette not beginning in white but with a color contrasting the light basemap.

**Describe alternatives you've considered**
appearingly none

**Usecases and impact**
lowest values of maps don't have contrast to appear as a measurement

**Additional context**
![image](https://user-images.githubusercontent.com/26090254/166484452-0bdf1b0b-fbab-4bbc-85ba-46c60183b258.png)
https://test.visualize.admin.ch/de/v/uAhcqNB7it-d

@zellersabine @afoletti

Rdataflow:
@ptbrowne cc @AnninaWalker

AFoletti:
I am open to any solution, but it is true that this is an issue.
Today, I often pick the color ramp not because of the color itself, but looking for hte better contrast possible with the basemap. Which is kind of a pity :)

ptbrowne:
@AFoletti Please tell us if the artificial clamping of the sequential color palette solves your problem. It is deployed on test.admin.ch.

AFoletti:
Hello @ptbrowne 
Looks way better, many thanks!

PR #617, opened 2022-08-12
feat: Artificially clamp range of sequential palettes for colors to be less white

ptbrowne:
Here we articifically clamp the left side of the sequential color palette as it is
too white and is not visible on maps.

I think I saw somewhere (maybe on GitLab) an issue that was talking about this,
and I experienced the problem myself.

I'd like to have your views on these. cc @rdataflow @anninawalker @bprusinowski﻿

### DIFF
--- a/app/configurator/components/ui-helpers.ts
+++ b/app/configurator/components/ui-helpers.ts
@@ -890,36 +890,6 @@ export const categoricalPalettes: Array<{
 
 export const getDefaultCategoricalPalette = () => categoricalPalettes[0];
 
-export const getColorInterpolator = (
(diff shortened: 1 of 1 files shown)


Rdataflow:
@ptbrowne LGTM let's go with this 👍 
Closes: https://github.com/visualize-admin/visualization-tool/issues/538

bprusinowski (approved):
👍

If the problem only appears for map charts, another solution would be to include border to shapes so they are clearly separated from the background.

ptbrowne:
Let's try this simple and easily reversible solution first, and see if it solves the problem. If in the future, we see that it causes problem, we can very easily go back ;)

PR #624, opened 2022-08-17
Better initial state for segment

ptbrowne:
When selecting a dimension for segmentation, due to missing sourceUrl and sourceType in the client.query call, we would not retrieve the hierarchy and select 7 values without care for the hierarchy. Now we select 7 leaves.
I've added also proper types to client.query calls so now we are warned.

### DIFF
--- a/app/configurator/configurator-state.tsx
+++ b/app/configurator/configurator-state.tsx
@@ -678,16 +678,15 @@ const reducer: Reducer<ConfiguratorState, ConfiguratorStateAction> = (
           dataSetMetadata,
           selectedValues: actionSelectedValues,
         } = action.value;
+        const component = dataSetMetadata.dimensions.find(
+          (dim) => dim.iri === componentIri
+        );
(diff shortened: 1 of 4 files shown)


app/configurator/configurator-state.tsx
```diff
@@ -671,15 +672,22 @@ const reducer: Reducer<ConfiguratorState, ConfiguratorStateAction> = (
...
+          if (field === "segment") {
+            const component = dataSetMetadata.dimensions.find(
+              (dim) => dim.iri === componentIri
             );
-            const selectedValues = action.value.selectedValues
-              ? action.value.selectedValues
-              : component?.values.slice(0, 7) || [];
+            const selectedValues = actionSelectedValues
+              ? actionSelectedValues
+              : component?.values.slice(0, SEGMENT_CHILDREN_INITIAL_LIMIT) ||
```
bprusinowski: 🎉

bprusinowski (approved):
Nice!

PR #625, opened 2022-08-17
DnD empty groups

ptbrowne:
Add empty component for when there are no groups in table chart options. Still draft, need input from @AnninaWalker.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 You can also check the [release page](https://github.com/visualize-admin/visualization-tool/releases)
 
+## Unreleased
+
+- Fixed organization search
+- Explanation on how to create groups for tables
+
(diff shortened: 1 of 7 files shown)


app/themes/federal.tsx
```diff
@@ -165,18 +165,18 @@ theme.typography = merge(theme.typography, {
     fontWeight: "bold",
   }),
   body1: createTypographyVariant(theme, {
-    fontSize: [14, 16],
-    lineHeight: [22, 24],
+    fontSize: 14,
+    lineHeight: [24],
     fontWeight: "regular",
   }),
```
bprusinowski, line 171: I think it should be fontSize: [16] here, no?

ptbrowne: I am keeping this for another PR, I need to see how to see this in the sketch file.

ptbrowne: I extracted the commit

bprusinowski (approved):
👌

PR #626, opened 2022-08-17
Make chart choice more visible

ptbrowne:
Previously, we had extracted the chart type choice inside the tabs. Here we revert this choice and put it back in the left panel.

### DIFF
--- a/app/themes/federal.tsx
+++ b/app/themes/federal.tsx
@@ -41,7 +41,7 @@ export const theme = createTheme({
     },
     divider: "#CCCCCC",
     action: {
-      hover: "#d8e8ef",
+      hover: "#F2F7F9",
     },
     secondary: {
       main: "#757575",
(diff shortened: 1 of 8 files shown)


app/themes/federal.tsx
```diff
@@ -163,18 +163,18 @@ theme.typography = merge(theme.typography, {
...
+    lineHeight: [24],
     fontWeight: "regular",
   }),
   body2: createTypographyVariant(theme, {
-    fontSize: [12, 14],
-    lineHeight: [18, 20],
+    fontSize: [14],
+    lineHeight: [20],
     fontWeight: "regular",
   }),
```
bprusinowski, line 174: Shouldn't body2 have different fontSize that body1? Or are the same in design specs?

bprusinowski (approved):
LGTM, but there is something wrong with translations (see Vercel preview – controls.section.chart.type) 👍

app/themes/federal.tsx
ptbrowne: Extracted the commit, I will see this with @AnninaWalker

PR #628, opened 2022-08-17
types: Update urql types

bprusinowski:


ptbrowne (approved):
💯 Nice I had only searched for client.query( so I had missed those cases 👍

PR #633, opened 2022-08-18
feat: Save DataSource label in the URL

bprusinowski:
To fix some bugs when sharing links to dataset previews, data source is now also stored in URL.

### DIFF
--- a/app/components/data-source-menu.tsx
+++ b/app/components/data-source-menu.tsx
@@ -77,6 +77,8 @@ export const DataSourceProvider = ({ children }: { children: ReactNode }) => {
     setSource(source);
   };
 
+  const router = useRouter();
+
   useEffect(() => {
     const dataSource = retrieveDataSourceFromLocalStorage();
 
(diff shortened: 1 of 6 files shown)


app/components/data-source-menu.tsx
```diff
@@ -44,12 +45,14 @@ export const DATA_SOURCE_OPTIONS: ({ isTrusted: boolean } & Option)[] = [
...
-const DATA_SOURCE_URLS: string[] = DATA_SOURCE_OPTIONS.map((d) => d.value);
+
+const SOURCES_BY_LABEL = keyBy(SOURCE_OPTIONS, (d) => d.label);
+const SOURCES_BY_VALUE = keyBy(SOURCE_OPTIONS, (d) => d.value);
 
 export const useIsTrustedDataSource = (dataSource: DataSource) => {
   return useMemo(() => {
     const stringifiedDataSource = stringifyDataSource(dataSource);
-    return DATA_SOURCE_OPTIONS.find((d) => d.value === stringifiedDataSource)
+    return SOURCE_OPTIONS.find((d) => d.value === stringifiedDataSource)
```
ptbrowne, line 55: You can use SOURCE_BY_VALUE here.

app/components/data-source-menu.tsx
```diff
@@ -77,19 +80,45 @@ export const DataSourceProvider = ({ children }: { children: ReactNode }) => {
...
       saveDataSourceToLocalStorage(DEFAULT_DATA_SOURCE);
     }
   }, []);
 
+  useEffect(() => {
+    if (router.isReady) {
+      const routerSourceLabel = router.query.dataSource as string;
+      const routerSource = SOURCES_BY_LABEL[routerSourceLabel]?.value;
+
+      const strSource = stringifyDataSource(source);
```
ptbrowne, line 100:
```suggestion
      const providerSource = stringifyDataSource(source);
```

ptbrowne: I find the logic here a bit hard to follow. I think it would be beneficial to try to refactor this into a more meaningful chunk whose responsibility are well defined.

```
useSyncRouterParam({
  param: 'dataSource',
  value: providerSourceLabel,
  onParamChange: (routerParam) => handleSourceChange(parseDataSource(routerParam))
})
```

With the above example, we have clear boundaries between the thing knowing the router API and charged with keeping things in sync = useSyncRouterParam, and the thing that the provider knows, the business logic = the onParamChange function. What do you think ?

ptbrowne (approved):
LGTM ! Thanks 🎉

PR #634, opened 2022-08-18
fix: Use a proper time unit to format brush extent dates

### LINKED ISSUE #571: Wrong label for min value in time slider when dateUnit is months
@ptbrowne Yes, it works. Thank you very much ! 
There is just a little detail in the Time-scala on the right side, but not really important (2011 instead 01.2011)

image

_Originally posted by @ortnever in https://github.com/visualize-admin/visualization-tool/issues/564#issuecomment-1135777555_

bprusinowski:
Closes #571.

I think it makes more sense to use the same date format as visible above the time slider, see picture below. I don't think we should use formatDateAuto when we have an easy access to a timeUnit of given TemporalDimension.

Screenshot 2022-08-18 at 12 54 27

### DIFF
--- a/app/configurator/components/filters.tsx
+++ b/app/configurator/components/filters.tsx
@@ -823,8 +823,10 @@ export const TimeFilter = ({
 
     return (
       <Box>
-        {timeFormatUnit(timeRange[0], timeUnit)} –{" "}
-        {timeFormatUnit(timeRange[1], timeUnit)}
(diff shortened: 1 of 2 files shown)


ptbrowne:
👍 LGTM. Nit: The text looks a little big, is it body1 ? I think we should use body2 here.

bprusinowski:
Good point, I changed it to body2 👍

PR #637, opened 2022-08-19
fix: Reflect state change immediately in dropdown

ptbrowne:
Rework the hook synchronizing the router state to send back a local
state, this way we do not have to "wait" for router state to be updated
for the value in the dropdown to be updated

Relation to localStorage is also simplified: localStorage is used only
at init, and on change of


bprusinowski (approved):
💯

PR #638, opened 2022-08-19
WIP: Use atoms in data source

bprusinowski:


app/domain/data-source/helpers.ts
```diff
@@ -44,12 +44,8 @@ export const isDataSourceChangeable = (pathname: string) => {
   }
 };
 
-export const retrieveStringifiedDataSourceFromLocalStorage = () => {
-  return localStorage.getItem("dataSource");
-};
-
 export const retrieveDataSourceFromLocalStorage = () => {
-  const dataSource = retrieveStringifiedDataSourceFromLocalStorage();
+  const dataSource = localStorage.getItem("dataSource");
```
ptbrowne, line 48: Whys is retrieveDataSourceFromLocalStorage still needed at this point ?

bprusinowski: It's used in ConfiguratorState (when initialising a new chart it needs to access a data source), see e.g. here: https://github.com/visualize-admin/visualization-tool/blob/main/app/configurator/configurator-state.tsx#L765, and as this is not a React component, we can't use useAtom to retrieve current value :( But maybe it could be possible to use a value from URL in this case :)

ptbrowne: Couldn't we use the initial state for this ? https://github.com/visualize-admin/visualization-tool/blob/main/app/configurator/configurator-state.tsx#L1191. The initial state would be dynamic there and would use the provider to get the data source ? I do not like too much that the configurator state has to look at the localStorage, only the datasource provider should now that something is happening with localStorage.

ptbrowne: https://github.com/immerjs/use-immer/blob/master/src/index.ts#L11 the initial arg can be a function :+1: which would make it easy to base the initial state on the provider I think.

Issue #640, opened 2022-08-22
Endpoint selections break the docs


ptbrowne:
The docs are white on https://test.visualize.admin.ch/docs?dataSource=Int, Firefox.

It seems like the useRouteState ends up in an infinite loop.

<img width="494" alt="image" src="https://user-images.githubusercontent.com/465582/185931089-16ff11ca-d305-4fa2-95d9-689013a91ea6.png">

It is reproducible.

ptbrowne:
Fixes https://github.com/visualize-admin/visualization-tool/pull/641

ptbrowne:
Fixed in #641

PR #641, opened 2022-08-22
Fix docs

ptbrowne:
- fix: Fix docs/ route infinite loop
- fix: Do not save in localStorage nor in URL the dataSource if it's the same as the default one


ptbrowne:
Fixes https://github.com/visualize-admin/visualization-tool/issues/640

bprusinowski (approved):
🎉

PR #646, opened 2022-08-23
Support partial search terms

ptbrowne:
Add wildcards to the Lucene query used in the search to support
partial terms.

- [x] "bath" should return the "bathing water quality" dataset
- [x] "tari" should return the electricity tariff datasets
﻿


bprusinowski (approved):
💯

PR #647, opened 2022-08-23
fix: Design issues

bprusinowski:
This PR fixes padding between base layer options in maps + adds padding to other options sections in the right panel, as they went missing after recent changes. Also, it fixes sizing issues with table chart sorting options.

### DIFF
--- a/app/configurator/components/chart-controls/section.tsx
+++ b/app/configurator/components/chart-controls/section.tsx
@@ -47,6 +47,7 @@ export const ControlSectionContent = ({
   role,
   ariaLabelledBy,
   children,
+  gap = "default",
   px,
   sx,
 }: {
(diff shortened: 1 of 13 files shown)


app/configurator/components/chart-controls/section.tsx
```diff
@@ -47,13 +47,15 @@ export const ControlSectionContent = ({
...
   children,
+  gap = "default",
   px,
   sx,
 }: {
   component?: ElementType;
   role?: string;
   ariaLabelledBy?: string;
   children: ReactNode;
+  gap?: "big" | "default" | "none";
```
ptbrowne, line 58: Let's use "large" instead of "big" to keep the same convention as MUI.

ptbrowne:
The first commit could be splitted between I think the fix on the map options spacing and the other change and introduction of a new property "gap" on the ControlSectionContent. It's not clear to me why this was introduced, I can only understand that some ControlSectionContent need a gap between their children, sometimes they do not.

The last commit could be splitted between the refactor and the fix. It's hard to review here.

I think in general that commit description can be a bit improved.

Thanks for the refactors though. As this is only styling I LGTM now :)

bprusinowski:
Thanks for a reminder, I have a tendency to mix stuff in the commits :) I hope it's better now!

ptbrowne (approved):

PR #648, opened 2022-08-23
Format currencies improvements

ptbrowne:
When we detect a currency in the data (from the unit in qudt.unit), we format the data with currencyExponent digits.

In this issue, the usecase of data where the floating precision is more than currencyExponent (for example for currency exchange datasets) is not tackled. It would be possible to tackle this when the dimension metadata informs us on the precision necessary.

Fixes https://github.com/orgs/visualize-admin/projects/1/views/18

This can be tested with the "State accounts - Office" dataset where amounts should have a precision of 2 digits after comma since the currency is CHF, and the currency exponent of CHF is 2. 

image

### DIFF
--- a/app/configurator/components/ui-helpers.spec.ts
+++ b/app/configurator/components/ui-helpers.spec.ts
@@ -56,6 +56,14 @@ describe("useDimensionFormatters", () => {
           isNumerical: true,
           isKeyDimension: false,
         } as DimensionMetaDataFragment,
+        {
+          iri: "iri-currency",
+          isNumerical: true,
+          isKeyDimension: false,
+          isCurrency: true,
+          currencyExponent: 1,
+          __typename: "Measure",
+        } as DimensionMetaDataFragment,
       ])
     );
     return { formatters };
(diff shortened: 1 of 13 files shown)


app/configurator/components/ui-helpers.spec.ts
```diff
@@ -75,6 +83,11 @@ describe("useDimensionFormatters", () => {
     const { formatters } = setup();
     expect(formatters["iri-number"]("2.33333")).toEqual("2,33");
   });
+
+  it("should work with currencies", () => {
+    const { formatters } = setup();
+    expect(formatters["iri-currency"]("20002.3333")).toEqual("20'002,3");
+  });
```
bprusinowski, line 90: Just wondering, if you're using locale in this formatter (currencyFormatter), can the test fail in some scenarios if run with a locale that use different formatting? Or will it be the same in all cases?

bprusinowski (approved):
LGTM 👌

app/rdf/query-hierarchies.ts
```diff
@@ -59,6 +60,29 @@ const toTree = (
...
 };
 
+const findHierarchyForDimension = (cube: Cube, dimensionIri: string) => {
+  const newHierarchy = cube.ptr
+    .any()
+    .has(ns.sh.path, rdf.namedNode(dimensionIri))
+    .has(ns.cubeMeta.inHierarchy)
+    .out(ns.cubeMeta.inHierarchy)
+    .toArray()
+    .shift();
```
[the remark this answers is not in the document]
ptbrowne: To get the first element, it's clearer with [0] I updated the code.

app/configurator/components/ui-helpers.spec.ts
ptbrowne: It could fail if run with a different locale, but for now we do not test for this.

PR #651, opened 2022-08-23
refactor: Use Zustand for DataSource

bprusinowski:
In this PR I've tried to use Zustand to manage DataSource.

Additionally, I re-introduced keeping of DataSource in the URL even when it's a default one, as otherwise we enable a bug of sharing a link to a dataset without DataSource in URL (so a default one should be used), but if the person we send a link to has another DataSource saved in their localStorage, it would led to a GQL error (as we do not reset localStorage to a default DataSource when entering a link w/o DataSource in URL).

I like this approach, as almost everything is handled directly by a store, with only exception of syncing URL in DataSourceMenu component. If this approach works for you, I will also add / adjust tests :)


app/stores/data-source.ts
```diff
@@ -0,0 +1,62 @@
...
+  DEFAULT_DATA_SOURCE,
+  getURLParam,
+  parseDataSource,
+  stringifyDataSource,
+} from "@/domain/datasource";
+import { isRunningInBrowser } from "@/lib/is-running-in-browser";
+
+type DataSourceStore = {
+  dataSource: DataSource;
+  setDataSource: (value: string) => void;
```
ptbrowne, line 17: It's a bit unfortunate that we get passed a full DataSource but when we set, we set a string. It'd be good to have symmetry here, we receive a DataSource and set a DataSource.

app/stores/data-source.ts
```diff
@@ -0,0 +1,62 @@
...
+      },
+      get,
+      api,
+      []
+    );
+
+    let dataSource = DEFAULT_DATA_SOURCE;
+
+    if (isRunningInBrowser()) {
+      const urlDataSource = getURLParam("dataSource");
```
ptbrowne: Shouldn't it be `name` instead of `"dataSource"` ?

It seems the persist is a bit generic (for localStorage) but not fully.

app/stores/data-source.ts
```diff
@@ -0,0 +1,62 @@
...
+  stringifyDataSource,
+} from "@/domain/datasource";
+import { isRunningInBrowser } from "@/lib/is-running-in-browser";
+
+type DataSourceStore = {
+  dataSource: DataSource;
+  setDataSource: (value: string) => void;
+};
+
+const customPersist =
```
ptbrowne: Could you add a docstring explaining the behavior, how things are stringified in localStorage and how is it stringified in the router (if I am not mistaken, they were stringified differently, maybe it'd be the opportunity to store the same thing in localStorage and urlParam).

app/stores/data-source.ts
```diff
@@ -14,8 +14,10 @@ type DataSourceStore = {
   setDataSource: (value: string) => void;
 };
 
+const STORAGE_KEY = "dataSource";
+
 const customPersist =
-  (name: string, config: StateCreator<DataSourceStore>) =>
+  (config: StateCreator<DataSourceStore>) =>
```
ptbrowne, line 45: Yes I think it's better not to have the name here since in fact the persist is not generic.

app/stores/data-source.ts
```diff
@@ -17,7 +17,13 @@ type DataSourceStore = {
 
 const STORAGE_KEY = "dataSource";
 
-const customPersist =
+/**
+ * Custom middleware that saves data source to localStorage.
+ *
+ * On initialization it tries to first retrieve data source from URL; if it's
+ * not here, tries with localStorage, otherwise uses a default data source.
+ */
```
ptbrowne, line 43: Would be helpful to know how the value is stored in the router and in localStorage.

app/stores/data-source.ts
```diff
@@ -17,6 +17,10 @@ type DataSourceStore = {
 
 const STORAGE_KEY = "dataSource";
 
+const saveToLocalStorage = (value: DataSource) => {
+  localStorage.setItem(STORAGE_KEY, stringifyDataSource(value));
```
ptbrowne: Could we use only the label here to have the same mechanism as the router ?

ptbrowne (commented):
I think we can continue with this approach 👍

ptbrowne (approved):
👍 LGTM Thanks it's cleaner like this 🎉

ptbrowne (approved):
👏

PR #653, opened 2022-08-24
refactor: Remove unused variable

ptbrowne:
- refactor: Remove unused variable
- refactor: Rename variable for clarity
- refactor: Extract hierarchyMetadata fragment + rename dimensionMetaData into dimensionMetadata
- feat: Setting default filters for area chart through hierarchy
- fix: Remove throw error when not finding a hierarchy as this is expected in most cases
- refactor: Remove unused commented code
- feat: Consider data type to set dimension resolution + do not trim currency values


app/rdf/parse.ts
```diff
@@ -236,6 +236,13 @@ export const parseCubeDimension = ({
   const rawOrder = dim.out(ns.sh.order).value;
   const order = rawOrder !== undefined ? parseInt(rawOrder, 10) : undefined;
 
+  const name = dim.out(ns.schema.name, outOpts).value ?? dim.path?.value!;
+  const resolution =
+    dataType?.equals(ns.xsd.int) || dataType?.equals(ns.xsd.integer)
+      ? 0
+      : undefined;
+
+  console.log({ name, resolution });
```
bprusinowski, line 246: 👁️‍🗨️

ptbrowne:
I see that the formatting code is not localized yes, I guess this is because across Switzerland we want to use formatting rules coming from the German part.

PR #654, opened 2022-08-24
chore: General cleanup

bprusinowski:
In this PR I've reorganised some files (e.g. consolidated lib and utils folder, moved generic types to one file) and removed unused params / functions / variables / fixed map data, etc.


ptbrowne (approved):
🎉

PR #658, opened 2022-08-24
feat: Reintroduce chart selection popup

bprusinowski:
This PR brings chart selection via tab back.

### DIFF
--- a/app/components/chart-selection-tabs.tsx
+++ b/app/components/chart-selection-tabs.tsx
@@ -3,7 +3,6 @@ import React, {
   createContext,
   Dispatch,
   ReactNode,
-  useCallback,
   useContext,
   useEffect,
   useState,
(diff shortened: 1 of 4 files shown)


app/components/chart-selection-tabs.tsx
```diff
@@ -83,10 +83,9 @@ const useStyles = makeStyles<Theme, { editable: boolean }>((theme) => ({
     padding: `${theme.spacing(1)} ${theme.spacing(3)}`,
     borderRadius: 3,
     transition: "0.125s ease background-color",
-    "&:hover": ({ editable }) => {
-      return editable
-        ? { backgroundColor: theme.palette.grey[200] }
-        : undefined;
+    "&:hover": {
+      backgroundColor: ({ editable }) =>
+        editable ? theme.palette.grey[200] : undefined,
```
ptbrowne, line 88: Ideally this patch should be merged with the previous commit so as not to have a commit that is not working.

ptbrowne (approved):
LGTM Thanks !

app/components/chart-selection-tabs.tsx
bprusinowski: Good point, thanks!

PR #660, opened 2022-08-25
Prefer useStyles instead of sx when there is a large number of properties

ptbrowne:
- Do not use `sx` for more than 5 properties (we will be able to lower the number), prefer to use useStyles as it leads to less entangled code
- Add custom lint rules checking usage of `sx` and `makeStyles`


bprusinowski (approved):
Very nice 👌

PR #663, opened 2022-08-25
Map default state

ptbrowne:
The selection of default state would only work on the second attempt.
Before I had put the logic at the wrong place in getAdjustedConfig where in fact it needs to be in getInitialConfig to work 
correctly.
I do not know how I missed that 🤷

### DIFF
--- a/app/configurator/config-form.tsx
+++ b/app/configurator/config-form.tsx
@@ -294,8 +294,6 @@ export const useActiveFieldField = ({
   };
 };
 
-// const chartType =
-
 // Specific ------------------------------------------------------------------
 export const useChartType = ({
   metadata,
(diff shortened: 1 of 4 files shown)


app/configurator/config-form.tsx
```diff
@@ -295,44 +294,39 @@ export const useActiveFieldField = ({
   };
 };
 
+// const chartType =
+
```
bprusinowski, line 298: 👁️

app/configurator/config-form.tsx
```diff
@@ -298,23 +298,24 @@ export const useActiveFieldField = ({
...
   value: ChartType;
   onChange: (chartType: ChartType) => void;
 } => {
   const [state, dispatch] = useConfiguratorState();
   const onChange = useEvent((chartType: ChartType) => {
-    if (!metaData) {
+    if (!metadata) {
       return;
     }
+    console.log("on change", metadata);
```
bprusinowski, line 313: 👁️‍🗨️

bprusinowski (approved):
Looks and works good! 👌

PR #667, opened 2022-08-30
feat: Map Chart improvements

bprusinowski:
This PR adds a functionality of mapping dimensions & measures to SymbolLayer colors. It also introduces improvements and fixes to the Map Chart itself.

### Main change
Instead of a single color stored in a `symbolLayer` field inside `chartConfig`, we now can store:
- single color (along with opacity): `fixed` mode,
- measure: `continuous` mode,
- other dimensions: `categorical` mode.

### Other changes
- Added Slider component,
- Used correct font sizes for small inputs,
- Re-introduced diverging color palettes for measures,
- Allowed to use multiple MultiFilters,
- Updated Deck.gl to newest version which exposes types,
- Allowed to have multiple points at the same coordinates (SymbolLayer),
- Added new Size icon,
- Introduced horizontal display of ColorLegend items.

### Bug fixes
- Default palette in ColorRamp is now chosen correctly,
- Radius for undefined observations is now correctly parsed as 0 (previously there were small black dots, now they are gone),
- Used a correct `measureIri` to format SymbolLayer errors,
- Fixed a bug where Map Chart was not resized properly after going fullscreen and vice versa,
- Introduced re-computation of colorMapping of initial load to show all the colors,
- Overlapping shapes are now drawn in the correct order, so tooltip works correctly (DeckGLOverlay),
- ColorLegend's title is only rendered when a title is actually present, so there is no unnecessary left margin.

### TODO
- [ ] Migrate existing charts to the new SymbolLayer color modes,
- [ ] Allow setting opacity for `categorical` & `continuous` modes,
- [ ] Parse OrdinalMeasures so they can be used in charts,
- [ ] Enable InteractiveFilters for Map Chart (??),
- [ ] Check why in some cases Map Chart is being re-rendered when changing SymbolLayer color dimension to categorical dimension,
- [ ] Think about how to better handle Map Chart (`fields.symbolLayer.colors`) vs all the other charts (`fields.segment`) problem.


app/charts/map/map-state.tsx
```diff
@@ -320,6 +322,12 @@ const useMapState = (
     return rgb ? [rgb.r, rgb.g, rgb.b] : [0, 0, 0];
   };
 
+  const colors = symbolLayer.colors;
+  const getSymbolColor =
+    colors.type === "single"
+      ? () => convertHexToRgbArray(colors.value)
+      : () => [0, 0, 0, 255 * 0.1];
```
ptbrowne: I think we can memo here so that getSymbolColor does not change everytime.

app/rdf/query-geo-shapes.ts
```diff
@@ -139,72 +98,3 @@ export const createGeoShapesLoader =
...
-
-    return go({ tree, level: level + 1, iris: parentIris });
-  };
-
-  const lowestLevelIris = narrowerValues
-    .filter((d) => !d.narrower)
-    .map((d) => d.iri);
-
-  return go({ tree: [], iris: lowestLevelIris });
-};
```
ptbrowne, line 210: 👍  Thanks 😅

app/charts/shared/colors.tsx
```diff
@@ -5,3 +5,16 @@ export const convertHexToRgbArray = (hex: string): [number, number, number] => {
...
+    case 3:
+      return `rgb(${rgbArray.join(",")})`;
+    case 4:
+      return `rgba(${rgbArray.join(",")})`;
+    default:
+      throw new Error(
+        `You need to pass 3 or 4 arguments when converting RGB array to HEX, while ${rgbArray.length} were provided.`
+      );
+  }
+};
```
ptbrowne, line 23: Should be fixed up into the first commit of this PR.

app/charts/map/map-state.tsx
```diff
@@ -311,11 +312,47 @@ const useMapState = (
...
-  const getSymbolColor =
-    colors.type === "single"
-      ? () => convertHexToRgbArray(colors.value)
-      : () => [0, 0, 0, 255 * 0.1];
+  // @ts-ignore
+  // console.log(
+  //   [...dimensions, ...measures].find(
+  //     (d) => d.iri === symbolLayer.colors.componentIri
+  //   )
+  // );
```
ptbrowne: 👁️

app/charts/map/map-state.tsx
```diff
@@ -311,11 +312,47 @@ const useMapState = (
...
+        const color = convertHexToRgbArray(colors.value);
+        return (_: Observation) => color;
+      case "categorical":
+        const component = dimensions.find((d) => d.iri === colors.componentIri);
+        const componentValuesByLabel = keyBy(component?.values, (d) => d.label);
+        return (d: Observation) => {
+          const label = d[colors.componentIri] as string;
+          const value = componentValuesByLabel[label].value;
+          const color = colors.colorMapping?.[value];
+          return color ? convertHexToRgbArray(color) : [0, 0, 0, 255 * 0.1];
```
ptbrowne: The rgb arrays could be built outside the function (`const colorMappingArray = mapValues(colorMapping, convertHexToRgbArray)`). This way we have only to build `number of colors` arrays instead of `number of observatiosn` arrays.

ptbrowne:
Awesome work already :)

app/charts/map/map-state.tsx
bprusinowski: Wow, of course it should be done like this 😆 Thanks for spotting!

app/charts/shared/colors.tsx
bprusinowski: Good point, done!

app/charts/index.ts
```diff
@@ -287,7 +287,10 @@ export const getInitialConfig = ({
             componentIri: geoCoordinates[0]?.iri || geoShapes[0]?.iri || "",
             measureIri: measures[0].iri,
             hierarchyLevel: 1,
-            color: "#1f77b4",
+            colors: {
+              type: "single",
+              value: "#1f77b4",
+            },
```
ptbrowne: This is not a refactor since it's changing the schema. It should be more prominent from the commit title + description that this commit is "dangerous".

app/charts/map/map.tsx
```diff
@@ -170,6 +171,7 @@ export const MapComponent = () => {
           });
         }
       },
+      // @ts-ignore
```
ptbrowne: It'd be good to add the reason why we ignore here.

app/charts/map/get-base-layer-style.ts
```diff
@@ -33,17 +33,17 @@ export const emptyStyle = {
...
 
-export const getBaseLayerStyle = (props: Props) => {
-  const { locale, showLabels } = props;
+const getBaseLayerStyle = ({
+  locale,
+  showLabels,
+}: {
+  locale: Locale;
+  showLabels: boolean;
+}): MapboxStyle => {
```
ptbrowne, line 46: Nit: The commit is a refactor.

app/charts/map/map.tsx
```diff
@@ -109,11 +106,11 @@ export const MapComponent = () => {
...
-  const baseLayerStyle = useMemo(() => {
-    return getBaseLayerStyle({ locale, showLabels: !areaLayer.show });
-  }, [locale, areaLayer.show]);
-
-  const mapStyle = showBaseLayer ? baseLayerStyle : emptyStyle;
+  const mapStyle = useMapStyle({
+    locale,
+    showBaseLayer,
+    showLabels: !areaLayer.show,
+  });
```
ptbrowne, line 154: 👍

app/charts/map/map.tsx
```diff
@@ -111,6 +106,9 @@ export const MapComponent = () => {
     showBaseLayer,
     showLabels: !areaLayer.show,
   });
+
+  const onHover = useOnHover();
```
ptbrowne, line 156: Not sure about this particular refactor. Here I have to dig into another file to understand that the tooltip is managed there. I am not even sure that the dispatch interaction and management of the tooltip are related so I think it's a bit awkard to have them tied together.

app/charts/map/constants.ts
```diff
@@ -1,4 +1,5 @@
 export const DEFAULT_SYMBOL_LAYER_FIXED_COLOR_OPACITY = 80;
+export const DEFAULT_COLOR = [0, 0, 0, 255];
```
ptbrowne, line 2: Can this be squashed with the commit that needs it ?

app/charts/map/helpers.ts
```diff
@@ -178,6 +179,17 @@ export const getBBox = (
...
+  props: MapboxOverlayProps & {
+    interleaved?: boolean;
+  }
+) {
+  // @ts-ignore
+  const overlay = useControl<MapboxOverlay>(() => new MapboxOverlay(props));
+  overlay.setProps(props);
+
+  return null;
+}
```
ptbrowne, line 192: Can you add a comment describing what is the purpose of MapboxOverlay ? I see fixing overlapping symbols in the issue, but I do not understand how is the overlay solving the problem.

On another note, this commits seems also to contain refactors.

app/configurator/map/map-chart-options.tsx
```diff
@@ -529,6 +541,16 @@ const NoGeoDimensionsWarning = () => {
   );
 };
 
+const NotEnoughMeasuresWarning = () => {
+  return (
+    <Box sx={{ my: 3, py: 3, px: 5, width: "80%" }}>
+      <Trans id="chart.map.warning.notEnoughMeasures">
+        In this dataset there are not enough measures to display as symbols.
```
ptbrowne, line 548: It's not clear to me here. What does it mean ?

app/configurator/map/map-chart-options.tsx
```diff
@@ -371,11 +375,20 @@ export const SymbolLayerSettings = memo(
...
+          .filter((d) => d.iri !== areaLayer.measureIri)
+          .map((d) => ({
+            value: d.iri,
+            label: d.label,
+          }));
+      } else {
+        return metaData.measures.map((d) => ({
+          value: d.iri,
+          label: d.label,
+        }));
```
ptbrowne, line 389: I think it makes it a bit clearer the intent here if we only put the `if` on the filter.

```
const filterMeasures = areaLayer.selectable ?
   d => d.iri !== areaLayer.measureIri : d => true

return metaData.measures
          .filter(filterMeasures)
          .map((d) => ({
            value: d.iri,
            label: d.label,
          }));
```

app/charts/shared/legend-color.tsx
```diff
@@ -39,7 +39,7 @@ const useStyles = makeStyles<Theme>(() => ({
     minHeight: "20px",
     gap: "1rem 1.5rem",
     display: "grid",
-    gridTemplateColumns: "repeat(auto-fill, minmax(300px, 1fr))",
+    gridTemplateColumns: "repeat(auto-fill, 1fr)",
```
ptbrowne, line 42: Why this change ? I think it's cleaner no ?

ptbrowne (approved):
I think it was a larger feature than expected, and it will be a very good feature. Kudos 💯 !

On the form of the PR, in the future, I think it'd be good to separate between the refactor PRs, and the features, and to have smaller PRs (aim for 5-10 commits).
It will be clearer and it would lead to less chance of the PR being blocked.
For me the most important here would be to have tests, maybe E2E for this particular feature.

app/charts/map/helpers.ts
bprusinowski: I added a small comment, but generally I also don't fully understand how exactly it fixed the problem – I think the problem was in a way we plotted symbols previously that was not 100% correct and this approach was recommended on DeckGL's site, thus "fix" comment.

app/charts/shared/legend-color.tsx
bprusinowski: It's not needed anymore, I will remove this commit – initially the color legend was vertical and with a min-width of 300px; other legends (e.g. for symbol size) were plotted on the right side of it, and the space between them was not looking natural for some datasets. But now as it's plotted below other maps' legend (and with a horizontal layout) this in fact is not needed.

app/themes/federal.tsx
```diff
@@ -432,6 +432,8 @@ theme.components = {
       sizeSmall: {
         height: 40,
         minHeight: 40,
+        fontSize: `${theme.typography.body2.fontSize} !important`,
+        lineHeight: `${theme.typography.body2.lineHeight} !important`,
```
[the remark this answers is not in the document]
bprusinowski: I've tried with different ways to apply this and I didn't find a way other than this :( Let me know if I should remove this and add an issue to tackle this properly 👍

app/charts/map/map-legend.tsx
```diff
@@ -113,14 +113,33 @@ export const MapLegend = () => {
...
+              <>
+                <Typography component="div" variant="caption">
+                  {symbolLayer.colors.componentLabel}
+                </Typography>
+                <ContinuousColorLegend
+                  palette={symbolLayer.colors.palette}
+                  domain={symbolLayer.colors.domain}
+                  getValue={(d: Observation) =>
+                    // @ts-ignore
+                    d[symbolLayer.colors.componentIri] as number
```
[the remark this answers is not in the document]
bprusinowski: Here we do `symbolLayer.colors.type === "continuous"` at top, but it's "forgotten" in this particular place, and TS error appears (it still thinks `symbolLayer.colors` can be either fixed, categorical or continuous, while it only can be continuous. Would it be better to the do (`symbolLayer.colors as` ...)?

app/charts/map/helpers.ts
ptbrowne: OK thanks for the explanation.

app/charts/map/map-state.tsx
```diff
@@ -154,6 +176,130 @@ const getAreaColorScale = ({
...
+        },
+        dimensions
+      );
+    case "continuous":
+      return getContinousSymbolColors(
+        colors as {
+          type: "continuous";
+          componentIri: string;
+          palette: string;
+        },
```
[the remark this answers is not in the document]
bprusinowski: The type narrowing does not work correctly in this case (in second and third case it doesn't narrow properly). Do you have an idea why?

app/components/form.tsx
```diff
@@ -94,6 +101,68 @@ export const Radio = ({
...
+          name={name}
+          id={`${name}-${value}`}
+          size="small"
+          value={value}
+          min={min}
+          max={max}
+          step={step}
+          disabled={disabled}
+          // @ts-ignore
+          onChange={onChange}
```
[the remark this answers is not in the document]
bprusinowski: Apparently there is some type mismatch:

Type 'Fn<[e: ChangeEvent<HTMLInputElement>], void> & FormEventHandler<HTMLDivElement>' is not assignable to type '((event: Event, value: number | number[], activeThumb: number) => void) | undefined',

I wasn't able to type it properly, let me know if you have an idea how to improve this.

app/charts/map/map.tsx
bprusinowski: It's due to typed version of Deckl.gl which started to throw type errors here, while it was (and still is) working properly. I tried to fix them, but wasn't able to do it in reasonable amount of time – I'll add a todo here to remember to come back to it.

PR #671, opened 2022-08-31
feat: Provide query timings in debug search page

ptbrowne:
To provide the debug search page with the actual graphql queries that were sent to the sparql server, I use the extensions part of the graphql response.

I use the context to store the particular query times (at each query I push the meta information in context.queries)
Then I use the formatResponse callback from apollo-server to add to the extensions part of the response the queries that were stored in the context while the response is processed.

### DIFF
--- a/app/graphql/context.tsx
+++ b/app/graphql/context.tsx
@@ -90,7 +90,10 @@ const createContextContent = async ({
 export const createContext = () => {
   let setupping: ReturnType<typeof createContextContent>;
 
-  return {
+  const ctx = {
+    // Stores meta information on queries that have been made during the request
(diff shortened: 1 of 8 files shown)


app/graphql/context.tsx
```diff
@@ -90,14 +90,19 @@ const createContextContent = async ({
...
+
     setup: async ({
       variableValues: { locale, sourceUrl },
     }: GraphQLResolveInfo) => {
       setupping = setupping || createContextContent({ locale, sourceUrl });
       return await setupping;
     },
   };
+
+  return ctx;
```
bprusinowski, line 105: 👌

bprusinowski (approved):
Cool! 🎉

PR #677, opened 2022-09-01
Fix link unversioned dataset

ptbrowne:
After the implementation of the datasource functionality, we moved the redirect to the front-end as we needed to access the localStorage. It seems that when we did that we missed one of the redirect.

I also experimented an issue with the router.replace function (an error while access /browser/dataset/iri, see the description of the commit), so I simplified the saving of the dataSource param in the URL, and used directly history.pushState.

Fixes https://github.com/visualize-admin/visualization-tool/issues/288

### How to test

Going to the following URLs should correctly redirect to the versioned IRI and the cube preview should be displayed.

- `/en/browse/dataset/https%3A%2F%2Fculture.ld.admin.ch%2Fsfa%2FStateAccounts_Function?dataSource=Int`
- `/en/browse/dataset/https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fubd000501?dataSource=Int`

### DIFF
--- a/app/domain/datasource/index.spec.ts
+++ b/app/domain/datasource/index.spec.ts
@@ -3,6 +3,15 @@ import mittEmitter from "next/dist/shared/lib/mitt";
 import { SingletonRouter } from "next/router";
 
 import { createUseDataSourceStore } from "@/stores/data-source";
+import { setURLParam } from "@/utils/router/helpers";
+
+jest.mock("@/utils/router/helpers", () => {
+  const original = jest.requireActual("@/utils/router/helpers");
+  return {
+    getURLParam: original.getURLParam,
+    setURLParam: jest.fn(),
+  };
+});
 
 const createRouter = ({ query }: { query: Record<string, string> }) => {
   const router = {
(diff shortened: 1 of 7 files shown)


app/utils/router/helpers.ts
```diff
@@ -18,3 +18,16 @@ export const updateRouterQuery = (
...
+  qs.delete(param);
+  qs.append(param, value);
+  const newUrl =
+    window.location.protocol +
+    "//" +
+    window.location.host +
+    window.location.pathname +
+    `?${qs.toString()}`;
+  window.history.replaceState({ path: newUrl }, "", newUrl);
+};
```
bprusinowski, line 33: I really like how much simpler things became because of this 👏 

Nit: I think it should be enough to just do `?${qs}`, as when it's interpolated by default it looks for .toString() method

app/domain/datasource/index.spec.ts
```diff
@@ -17,26 +26,6 @@ const createRouter = ({ query }: { query: Record<string, string> }) => {
...
-
-    // @ts-ignore
-    delete window.location;
-    // @ts-ignore
-    window.location = {
-      href: url.toString(),
-    };
-
-    router.events.emit("routeChangeComplete");
-  };
```
bprusinowski, line 39: 💯

bprusinowski (approved):
LGTM, and 💯 for the tests 👌

app/utils/router/helpers.ts
ptbrowne: Done, and used string interpolation instead of the + ;)

PR #678, opened 2022-09-01
feat: Add E2E test for unversioned link to dataset

ptbrowne:
Add a simple test to ensure that we do not have the unversioned link bug again in the future

### DIFF
--- a/cypress/integration/unversioned.spec.ts
+++ b/cypress/integration/unversioned.spec.ts
@@ -1,0 +1,10 @@
+describe("Unversioned dataset", () => {
+  it("should be possible to open a link to an unversioned dataset", () => {
+    cy.visit(
+      `/en/browse/dataset/https%3A%2F%2Fculture.ld.admin.ch%2Fsfa%2FStateAccounts_Function?dataSource=Int`
+    );
+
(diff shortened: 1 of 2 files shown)


cypress/integration/unversioned.spec.ts
```diff
@@ -0,0 +1,21 @@
+const resizeObserverLoopErrRe = /> ResizeObserver loop/;
+Cypress.on("uncaught:exception", (err) => {
+  if (resizeObserverLoopErrRe.test(err.message)) {
+    return false;
+  }
+});
```
bprusinowski, line 6: Maybe `(err) => !err.message.includes("ResizeObserver loop")` would be simpler? Or does it catch more cases right now (not sure about the ">" sign at the beginning)?

bprusinowski (approved):
LGTM!

cypress/integration/unversioned.spec.ts
ptbrowne: Yep.. Not sure why I used a regexp in the first test, I removed it from there, I do not think it should be fired, and modified the original code. 🤞

bprusinowski: Cool!

Issue #683, opened 2022-09-07
very slow search if several keywords are entered


ortnever:
The search takes more than one minute  if several keywords are entered. However, the result is good.
I had once an "error message" but I cannot reproduce it.

**To Reproduce**
1. Go to 'test.visualize.admin.ch', source:INT
2. Search "Emissions de gaz à effet de serre".
3. Result are displayed after 1 minute 20 sec.

Just for information, the print screen of the failed search. But it happened only once:
![image](https://user-images.githubusercontent.com/79834090/188919249-385fecb6-b851-403f-8336-8353b5335f10.png)

**Environment (please complete the following information):**
 - Visualize environment and version: v3.7.10 ([bfb9035]
 - Browser and version : Edge

adintegra:
@ptbrowne what do you think about trying a pure SPARQL approach to the search, similar to what they have done [here](https://environment.ld.admin.ch/datasets/)? Something along [these lines](https://s.zazuko.com/4dc43F) maybe. 

It's a slightly blunter instrument and probably has some tradeoffs I'm missing right now, but curious to hear your thoughts.

ptbrowne:
This is fixed via #726 .

PR #686, opened 2022-09-07
feat: Split words of query and send 1 query per word

ptbrowne:
This PR tries to improve the search times when there are several words.
Instead of considering the full query, we send several query for each word, ignoring very small words that are not capitalized.

- feat: Split words of query and send 1 query per word
- feat: Default debug search uses drafts
- fix: Ignore search terms that are not capitalized and under 2 letters of length

This is related to https://github.com/visualize-admin/visualization-tool/issues/683

### DIFF
--- a/app/components/debug-search.tsx
+++ b/app/components/debug-search.tsx
@@ -177,7 +177,7 @@ const Search = ({
 };
 
 const DebugSearch = () => {
-  const [includeDrafts, setIncludeDrafts] = useState(false);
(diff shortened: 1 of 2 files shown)


ptbrowne:
@l00mi When you have further information on whether the stardog full text search or our queries can be improved, could you post it here ? Thanks 🙇

l00mi:
Yes will do. Thank you for this inputs.

PR #689, opened 2022-09-08
Hierarchy last level

ptbrowne:
- refactor: Add types to jest config
- fix: Update cube hierarchy query


.github/workflows/lint-test.yml
```diff
@@ -55,6 +55,19 @@ jobs:
           start: "env PORT=3333 npm run start"
           wait-on: "http://localhost:3333"
           record: false
+      - name: Upload Cypress screenshots
+        uses: actions/upload-artifact@v2
+        if: failure()
```
bprusinowski, line 60: It means that screenshots will be only sent on failure?

bprusinowski (approved):

PR #695, opened 2022-09-08
feat: Add chartConfig versioning

bprusinowski:
As there can be breaking changes to the chartConfig schema (as was the case for SymbolLayer colors recently), we should have a mechanism of migrating old chartConfig to new ones, so the charts created with older visualize versions would still work in the current version.

This PR is a proposal to handle such situations.

### To clarify
Should we save migrated chartConfigs of published charts in the database?

### TODO
- [x] add tests

### DIFF
--- a/app/utils/chart-config/versioning.spec.ts
+++ b/app/utils/chart-config/versioning.spec.ts
@@ -1,0 +1,66 @@
+import { decodeChartConfig } from "@/configurator/config-types";
+
+import { migrateChartConfig } from "./versioning";
+
+describe("config migrations", () => {
+  const oldConfig = {
+    version: "1.0.0",
+    chartType: "map",
+    interactiveFiltersConfig: {
+      legend: {
+        active: false,
+        componentIri: "",
+      },
+      time: {
+        active: false,
+        componentIri: "",
+        presets: {
+          type: "range",
+          from: "",
+          to: "",
+        },
+      },
+      dataFilters: {
+        active: false,
+        componentIris: [],
+      },
+    },
+    filters: {},
+    fields: {
+      areaLayer: {
+        show: true,
+        componentIri: "GeoShapesDimensionIri",
+        measureIri: "MeasureIri",
+        colorScaleType: "continuous",
+        colorScaleInterpolationType: "linear",
+        palette: "oranges",
+        nbClass: 5,
+      },
+      symbolLayer: {
+        show: true,
+        componentIri: "GeoCoordinatesDimensionIri",
+        measureIri: "MeasureIri",
+        color: "blue",
+      },
+    },
+    baseLayer: {
+      show: true,
+    },
+  };
+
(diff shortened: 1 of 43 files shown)


app/utils/chart-config/versioning.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    baseLayer: {
+      show: true,
+    },
+  };
+
+  it("should migrate to newest config", () => {
+    const migratedConfig = migrateChartConfig(oldConfig);
+    const decodedConfig = decodeChartConfig(migratedConfig);
+
+    expect(decodedConfig).not.toBeUndefined();
```
ptbrowne, line 55:
```
    expect(decodedConfig).toMatchSnapshot();
```

? 

I think snapshots are perfectly suited for those cases. This way as reviewer I could see the result.

app/utils/chart-config/versioning.ts
```diff
@@ -0,0 +1,148 @@
...
+      return newConfig;
+    },
+  },
+];
+
+export const migrateChartConfig = (
+  config: any,
+  toVersion: string = CHART_CONFIG_VERSION
+): any => {
+  const fromVersion = config.version || "1.0.0";
```
ptbrowne, line 111: Nit: I think having an options object as 2nd parameters with { toVersion?: undefined, fromVersion?: undefined } would be a bit more versatile.

app/utils/chart-config/versioning.ts
```diff
@@ -0,0 +1,148 @@
...
+        const { fields } = newConfig;
+        const { symbolLayer } = fields;
+        newConfig = {
+          ...newConfig,
+          fields: {
+            ...fields,
+            symbolLayer: {
+              show: symbolLayer.show,
+              componentIri: symbolLayer.componentIri,
+              measureIri: symbolLayer.measureIri,
```
ptbrowne, line 69: Nit: Spreading the symbolLayer could even be clearer ? Actually I think that using immer could be perfect.

app/utils/chart-config/versioning.ts
```diff
@@ -0,0 +1,148 @@
...
+    return config;
+  }
+
+  const currentMigration = migrations.find(
+    (migration) => migration[direction === "up" ? "from" : "to"] === fromVersion
+  );
+
+  if (currentMigration) {
+    const newConfig = currentMigration[direction](config);
+    return migrateChartConfig(newConfig, toVersion);
```
ptbrowne, line 124: Took me some time to understand how we would go to the latest version, I was searching for a `while toVersion !== curVersion`, but now I noticed the recursive call, guess I am tired 😅

ptbrowne (approved):
Great improvement 💯 Some smaller nits, LGTM !

ptbrowne (approved):
👏

Issue #688, opened 2022-09-08
Button "Back" no longer works


ortnever:
Button "Back" no longer works

**To Reproduce**
1. Go to 'test.visualize.admin.ch" source : INT
2. Open any dataset
3. Start a visualization
4. Click on 'back'

![image](https://user-images.githubusercontent.com/79834090/189054664-7818651f-3eee-4699-8010-37c226da40c8.png)

**Environment (please complete the following information):**
 - Visualize environment and version: v3.7.10 ([bfb9035]
 - Browser and version: Edge

PR #696, opened 2022-09-09
Fix back button

ptbrowne:
- refactor: Use useEvent instead of useCallback
- refactor: Remove duplicate route-state
- fix: Repair back button

When clicking on the UI "Back" button when configuring the chart, the URL changed, but the view was not updated.
I am not too sure why the view is not updated while the route was changed. I could not reproduce 100% the bug but I think that with router.push we'll have more luck, I am wondering if it's not related with the history mangling that we do for the datasource (I would guess that since we manually change it, react-router or history have trouble finding back the previous state 🤷 ).

Should fix https://github.com/visualize-admin/visualization-tool/issues/688


app/utils/router/use-route-state.spec.tsx
```diff
@@ -4,7 +4,7 @@ import { useRouter } from "next/router";
 
 import { DEFAULT_DATA_SOURCE } from "@/domain/datasource";
 
-import { useRouteState } from "./router/use-route-state";
+import { useRouteState } from "./use-route-state";
 
 jest.mock("next/router", () => {
   return {
```
bprusinowski, line 10: 👌

bprusinowski (approved):
Nice, seems to work fine 🎉

PR #697, opened 2022-09-09
fix: Undefined Scatterplot points

### LINKED ISSUE #669: Scatter plot displays points with missing values
Scatter plot displays points, where one dimension is a missing value as 0.

One example here (can be found [here](https://test.visualize.admin.ch/en/v/CNsDvACFEMu3))
image

The expected behavior is for the point to not show up at all. It seems to me, that the visualization tool is aware of that, as hovering over the point displays the point coordinates of the closest point instead (as can be seen in the screenshot). Nevertheless, the point shows up with the missing value replaced as 0 which is not ideal.

The issue happens on the test environment with version 3.7.10 and both in the Edge Browser (104.0.1293.70) as well as Chrome (105.0.5195.53)

bprusinowski:
Fixes #669.


app/charts/scatterplot/scatterplot-state.tsx
```diff
@@ -72,15 +72,14 @@ const useScatterplotState = ({
   const getY = useOptionalNumericVariable(fields.y.componentIri);
   const getSegment = useSegment(fields.segment?.componentIri);
 
-  // All data, sort by segment
-  const sortedData = data.sort((a, b) =>
-    ascending(getSegment(a), getSegment(b))
-  );
+  const plottableSortedData = data
+    .filter((d) => getX(d) !== null && getY(d) !== null)
+    .sort((a, b) => ascending(getSegment(a), getSegment(b)));
```
ptbrowne, line 77: Can use useMemo here ? Since we are dealing with a potentially big array + a sort :)

ptbrowne: Actually I see that this could be done in usePreparedData. I think it makes sense to have it as a default filter, and it'd be memoed directly.

app/charts/scatterplot/scatterplot-state.tsx
```diff
@@ -111,7 +110,7 @@ const useScatterplotState = ({
   const yScale = scaleLinear().domain(yDomain).nice();
 
   const hasSegment = fields.segment ? true : false;
-  const segments = [...new Set(sortedData.map(getSegment))]; // get *all* segments
+  const segments = [...new Set(plottableSortedData.map(getSegment))]; // get *visible* segments
```
ptbrowne, line 113: Nit: This could be memoed actually.

app/charts/scatterplot/scatterplot-state.tsx
bprusinowski: I tried this and realised that then there are situations where we have more items in ColorLegend below the chart than visible on the chart itself, and it shouldn't be like that (as we only filter preparedData, while segments are created basing on "full" data). I refactored the code to have a function that does this filtering across all charts. I think it would make sense to test it out, as I am not sure if it won't break some time filtering stuff 😆

app/charts/shared/chart-helpers.tsx
```diff
@@ -71,28 +71,21 @@ type ValuePredicate = (v: any) => boolean;
...
+  const isPlottable = useCallback((d: Observation) => {
+    for (let p of plotters) {
+      const v = p(d)
+      if (v === undefined || v === null) {
+        return false
+      }
+    }
+    return true
+  }, [plotters]);
+  return useMemo(() => data.filter(isPlottable), [data, isPlottable]);
```
bprusinowski, line 88: 💯

app/configurator/config-form.tsx
```diff
@@ -99,8 +99,8 @@ export const useChartFieldField = ({
       const tree = hierarchyData?.dataCubeByIri?.dimensionByIri
         ?.hierarchy as HierarchyValue[];
 
-      // If the dimension has a hierarchy, we select 7 leaves
-      const leafs = getLeafs(tree);
+      // If the dimension has a hierarchy, we select leaves
+      const leafs = getLeaves(tree);
```
bprusinowski, line 103: `const leaves`?

ptbrowne: Done

ptbrowne (approved):

PR #701, opened 2022-09-12
feat: Remove limit of 7 leafs selected by default

ptbrowne:
Draft﻿


bprusinowski:
I see it's a draft now, but when visiting the preview deployment and creating a chart, it still selects 7 values 😅

ptbrowne:
Continued in #697

PR #704, opened 2022-09-13
feat: Tooltips show numbers that are formatted according to their dimension formatter

ptbrowne:
We have rules for formatting numbers according to what their dimension specifies.
Previously those rules applied only to the table.
Now they apply to the tooltips of the charts.

### DIFF
--- a/app/charts/area/areas-state.tsx
+++ b/app/charts/area/areas-state.tsx
@@ -48,11 +48,10 @@ import {
 } from "@/configurator/components/ui-helpers";
 import { Observation } from "@/domain/data";
 import { useLocale } from "@/locales/use-locale";
+import useChartFormatters from "@/shared/use-chart-formatters";
 import { sortByIndex } from "@/utils/array";
 import { estimateTextWidth } from "@/utils/estimate-text-width";
 import { makeOrdinalDimensionSorter } from "@/utils/sorting-values";
-
-import useChartFormatters from "../shared/use-chart-formatters";
 
 export interface AreasState {
   chartType: "area";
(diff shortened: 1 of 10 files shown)


app/charts/area/areas-state.tsx
```diff
@@ -51,6 +52,8 @@ import { sortByIndex } from "@/utils/array";
 import { estimateTextWidth } from "@/utils/estimate-text-width";
 import { makeOrdinalDimensionSorter } from "@/utils/sorting-values";
 
+import useChartFormatters from "../shared/use-chart-formatters";
```
bprusinowski, line 55: That's totally a huge nit, but I think it could be "@/charts/shared/...", as like with the imports above?

app/charts/map/map-state.tsx
```diff
@@ -446,14 +446,8 @@ const useMapState = (
...
-  const areaMeasureLabel = useMemo(
-    () => measures.find((m) => m.iri === areaLayer.measureIri)?.label || "",
-    [areaLayer.measureIri, measures]
-  );
-  const symbolMeasureLabel = useMemo(
-    () => measures.find((m) => m.iri === symbolLayer.measureIri)?.label || "",
-    [symbolLayer.measureIri, measures]
-  );
+  const areaMeasureLabel = areaMeasureDimension?.label || "";
+  const symbolMeasureLabel = symbolMeasureDimension?.label || "";
```
bprusinowski, line 450: 🙇

bprusinowski (approved):
LGTM!

PR #705, opened 2022-09-13
Refactor chart state

ptbrowne:
- feat: Tooltips show numbers that are formatted according to their dimension formatter
- refactor: Use useMemo to group scales / data preparation


bprusinowski (approved):
Nice, I like the approach of grouping stuff needed for particular concepts (data, scales, etc) in shared useMemo 👍

PR #707, opened 2022-09-14
fix: Use correct default color

ptbrowne:
- fix: Use correct default color
- feat: Add small transition on map legend indicator
- chore: Defensive programming on eslint rule


bprusinowski (approved):
💯

PR #709, opened 2022-09-14
chore: Improve E2E deployment

ptbrowne:
Instead of building the app on Github, uploading the asset, retrieving the asset on a cypress action etc... we directly test against the vercel deployment.

This should be quicker than the alternative as Vercel deployment is very fast.
It also nicely decouples each task.

[See explanatory blog post here](https://glebbahmutov.com/blog/develop-preview-test/#bonus-print-the-github-event-and-values)


.github/workflows/e2e.yml
```diff
@@ -7,8 +7,17 @@ jobs:
     if: github.event.deployment_status.state == 'success'
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 12.x
```
bprusinowski, line 13: Shouldn't it be Node 16 now?

bprusinowski (approved):
LGTM, thanks for providing a blog post about this 🎉

PR #710, opened 2022-09-14
feat: Apply hierarchical dimension first in filters & select top most…

### LINKED ISSUE #711: In left filters, initial value should be topmost value for dimension with hierarchy
> As a user, I expect to have left filters initially set to topmost parent if values available (i.e. Switzerland or Total) to get a good overview of the dataset before digging into a specific slice (i.e. my municipality or my special selection).  reasoning: in the majority of the cases this will be a good initial default. for some rare edge cases i.e. values in % this might feel a bit odd

ptbrowne:
Apply hierarchical dimension first in filters & select top most value.

- For NFI datasets, Switzerland should be selected first in the left filters : https://visualization-tool-git-feat-topmost-map-layer-ixt1.vercel.app/en/create/new?cube=https://environment.ld.admin.ch/foen/nfi/49-19-None-None-44/cube/1
- For Public accounts datasets, Total should be selected first in the left filters : https://visualization-tool-git-feat-topmost-map-layer-ixt1.vercel.app/en/create/new?cube=https://environment.ld.admin.ch/foen/fab_Offentliche_Ausgaben_test3/8

Smaller fixes on color palette to make it coherent with other fields (checked with @AnninaWalker)

- feat: Use Select from MUI to implement ColorRampField
- fix: Missing gutter
- feat: Add name of palette
- feat: Better spacing for ListSubheader
- feat: Make color palette field width 100%
- feat: Remove downshift

Fix #711

### DIFF
--- a/app/configurator/components/chart-controls/color-ramp.tsx
+++ b/app/configurator/components/chart-controls/color-ramp.tsx
@@ -1,8 +1,13 @@
 import { Trans } from "@lingui/macro";
-import { Box, Button, Typography } from "@mui/material";
-import { useSelect } from "downshift";
+import {
+  Box,
+  Select,
+  MenuItem,
+  ListSubheader,
+  SelectProps,
+} from "@mui/material";
 import get from "lodash/get";
-import { useCallback, useEffect, useMemo, useRef } from "react";
+import { useEffect, useMemo, useRef } from "react";
 
 import { Label } from "@/components/form";
 import {
@@ -15,7 +20,7 @@ import {
   Palette,
   sequentialPalettes,
 } from "@/configurator/components/ui-helpers";
-import { Icon } from "@/icons";
+import useEvent from "@/utils/use-event";
 
 // Adapted from https://observablehq.com/@mbostock/color-ramp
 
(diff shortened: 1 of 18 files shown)


app/configurator/components/chart-controls/color-ramp.tsx
```diff
@@ -90,118 +95,61 @@ export const ColorRampField = ({
...
     palettes.find((d) => d.value === currentPaletteName) || defaultPalette;
 
-  const onSelectedItemChange = useCallback(
-    ({ selectedItem }) => {
-      if (selectedItem) {
+  const onSelectedItemChange: SelectProps<typeof currentPalette>["onChange"] =
+    useEvent((_ev, selectedItem) => {
+      const value = _ev.target.value as typeof currentPalette["value"];
+      console.log({ value, _ev, selectedItem });
+      // ^?
```
bprusinowski, line 103: 👁️‍🗨️

bprusinowski: I am also not sure if `selectedItem` is used anywhere?

bprusinowski (commented):
Really nice changes, especially the ColorRamp is much cleaner with MUISelect 👏 But if you select a measure as color in a SymbolLayer and then disable a layer, you are still able to open and change palette, which shouldn't be possible. Also, I am not sure if we need palette names (as they are not sure descriptive for the end users anyway IMO), but it's also fine to have them :)

ptbrowne:
> But if you select a measure as color in a SymbolLayer and then disable a layer, you are still able to open and change palette, which shouldn't be possible.

Ah yes I had forgotten a `disabled`. 🙇 

> Also, I am not sure if we need palette names (as they are not sure descriptive for the end users anyway IMO), but it's also fine to have them :)

OK this was a test, I think it'd be better with names but you're right that we should find better names.

PR #742, opened 2022-09-27
Chart loading UI

ptbrowne:
When loading charts, chart elements (filters, chart, footnotes) were jumping around due to space between.
The original goal of the space between was to put the footnotes at the bottom of the chart container.
It is now achieved by having a growing container for the loading.

At the same time, I refactored to use the same component for the loading logic for all charts so that we are not repeating ourselves on all chart types.


app/charts/area/chart-area.tsx
```diff
@@ -35,6 +27,8 @@ import {
 } from "@/graphql/query-hooks";
 import { useLocale } from "@/locales/use-locale";
 
+import { ChartLoadingWrapper } from "../scatterplot/chart-loading-wrapper";
+
```
bprusinowski, line 31: I think this wrapper shouldn't be located in scatterplot 😅

ptbrowne: Done :)

PR #743, opened 2022-09-27
fix: Map offset in edition page

ptbrowne:
The map offset problem would disappear after a window resize.
Here, we trigger a resize after the style data has been loaded
and it fixes the offset problem between viz layer and base layer.


bprusinowski (approved):
Nice 👌

Issue #730, opened 2022-09-21
Inconsistencies in the number of results displayed by category


ortnever:
The number of "draft" datasets shown per category in the left panel does not match the number of results displayed in the center panel. 

**To Reproduce**
1. open int.visualize.admin.ch
2. check the box "include draft datasets".
3. Click on a category in the left panel
4. Compare the number of results.

![image](https://user-images.githubusercontent.com/79834090/191467702-34635aa1-f2a8-4040-88da-8ca44535ff06.png)

![image](https://user-images.githubusercontent.com/79834090/191467844-41213b7a-f2db-44e9-9280-a77c789b6352.png)

 - Visualize environment and version: v3.9.2 ([3fc980f]
 - Browser and version : Edge 105.0.1343.27

sosiology:
@ortnever we will try to tackle this issue this week, so it can be tested before we deploy to production next week.

ptbrowne:
Fixed by https://github.com/visualize-admin/visualization-tool/pull/745

sosiology:
@ortnever the bug has been resolved and can be reviewed and tested on int.visualize.admin.ch

ortnever:
@sosiology @ptbrowne 
yes, everything is ok now. I think the search works very well now. Thank you very much for all enhancements.

PR #745, opened 2022-09-28
Search improvements

### LINKED ISSUE #740: Search : different problems
I published a dataset with Title "untitre" and   following "Categorie" and "Organisation", as Draft. 

image

According to the documentation the search functionality includes these cube fields in its scope:

✅ Title: schema:name : ok
✅ Description: schema:description : ok
✅ Publisher: dcterms:publisher : ok
✅ Organisation label: dcterms:creator/schema:name : ok

⚠️ Theme label: dcat:theme/schema:name (tested with "Crime, criminal justice" and "Trade") 

- [ ] No result with "Trade"
-  1 result with "Crime" or "criminal justice" : ok

- [ ] If I search all drafts I do not find the dataset and the categories "Crime, criminal justice" and "Trade", as well as the Organisation "Swiss post" are not displayed in the left panel.

image

If I search "crime", I find the dataset and the categories "Crime, criminal justice" and "Trade" are displayed in the left panel.
image

 - Visualize environment and version: v3.9.2 ([3fc980f]
 - Browser and version: Edge

ptbrowne:
Fix #740 and #730

- The inconsistency between the left panel and the number of results was due to the query for the results not returning datasets without theme or creator.
- There was also a problem that we did not search via theme / creator if we already had results through name description. To fix this problem, we only do one query. Since some datasets can miss the theme / creator / publisher, we have to do a `bound()` check.

### DIFF
--- a/app/rdf/query-search.ts
+++ b/app/rdf/query-search.ts
@@ -1,4 +1,5 @@
-import { DESCRIBE, SELECT } from "@tpluscode/sparql-builder";
+import { TemplateResult } from "@tpluscode/rdf-string/lib/TemplateResult";
+import { DESCRIBE, SELECT, sparql } from "@tpluscode/sparql-builder";
 import clownface from "clownface";
 import { descending } from "d3";
 import { Cube } from "rdf-cube-view-query";
@@ -10,7 +11,7 @@ import { truthy } from "@/domain/types";
 import { DataCubeSearchFilter } from "@/graphql/resolver-types";
 import { ResolvedDataCube } from "@/graphql/shared-types";
 import * as ns from "@/rdf/namespace";
-import { parseCube, parseIri, parseVersionHistory } from "@/rdf/parse";
+import { parseCube, parseIri } from "@/rdf/parse";
 import { fromStream } from "@/rdf/sparql-client";
 
 import { computeScores, highlight } from "./query-search-score-utils";
@@ -23,7 +24,7 @@ const makeInFilter = (varName: string, values: string[]) => {
   return `
     ${
       values.length > 0
-        ? `FILTER (
+        ? `FILTER (bound(?${varName}) &&
     ?${varName} IN (${values.map(toNamedNode)})
   )`
         : ""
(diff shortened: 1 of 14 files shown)


app/rdf/query-search.ts
```diff
@@ -290,7 +291,7 @@ export const searchCubes = async ({
 
   // Sort the cubes per score using previously queries scores
   const results = cubes
-    .filter((c) => !!c?.data)
+    .filter((c): c is ResolvedDataCube => !!c?.data)
```
bprusinowski, line 252: 👌

app/rdf/query-search.ts
```diff
@@ -72,10 +73,17 @@ const enhanceQuery = (rawQuery: string) => {
   return enhancedQuery;
 };
 
-const contains = (left: string, right: string) => {
+const icontains = (left: string, right: string) => {
```
bprusinowski, line 76: `icontains` as a shortcut for...?

cypress/integration/selectors.ts
```diff
@@ -2,15 +2,16 @@ type Cy = Cypress.Chainable;
...
   edition: {
-    findLeftPanel: (cy: Cy) => cy.get('[data-name="panel-left"]'),
-    findRightPanel: (cy: Cy) => cy.get('[data-name="panel-right"]'),
-    findConfiguratorFilters: (cy: Cy) =>
-      cy.findByTestId("configurator-filters"),
-    findChartFiltersList: (cy: Cy) => cy.findByTestId("chart-filters-list"),
+    configFilters: (cy: Cy) => cy.findByTestId("configurator-filters"),
+    chartFilters: (cy: Cy) => cy.findByTestId("chart-filters-list"),
   },
 };
```
bprusinowski, line 16: 🙇

bprusinowski: I think also an option to consider would be to not have callbacks here (e.g. `selectors.search.searchInput: "dataset-search"`). It would require us to convert all of them to use test-ids, which would be good thing I believe, and then we could do `cy.findByTestId(selectors.search.input)`, for example. But this is not a scope of this PR and I am not sure if this would be better, let me know what you think :)

bprusinowski (approved):
LGTM! I like that the highlighting is back 💯

One thought: previously words were marked with yellow background, which maybe was too much, but right now I am not sure the bolded part is distinctive enough either 😅

app/rdf/query-search.ts
ptbrowne: insensitive

cypress/integration/selectors.ts
ptbrowne: I think for some selectors we can use findByLabelText or findByPlaceholder, this is why I chose to support the full `cy` object here. IMHO this is better that test id since we do not have to add them, and test ids should only be an escape hatch when we have no proper user seeable thing that can be targeted.

bprusinowski: Right, I didn't think about selecting by text (and I based the comment on [this section](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements), where they recommend to use custom data attributes). But I'm fine with the current approach, just wanted to hear your opinion, thanks!

ptbrowne: I am basing myself on the guiding principles of testing library, not of Cypress. This is why I added the methods from testing library. I think it's better if we do not have to add a lot of data attributes everywhere, and test as much with "human" visible attributes.

https://testing-library.com/docs/queries/about#priority

PR #746, opened 2022-09-29
feat: Add dimension descriptions to more places in UI

### LINKED ISSUE #744: Metadata: Information from schema:description missing in the UI
**Describe the Enhancement**
metadata is still missing on certain filters and legends in both the published chart and editor mode. 
Reference: https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/350#note_8705

**To Reproduce**
1. Go to 'https://int.visualize.admin.ch/en/create/n9fdoq_th318?dataSource=Int'

**Expected behavior**
Metadata should be shown on the following filters and legends (see yellow highlighted areas in screenshots below)
image

 - Visualize environment and version: INT v3.9.2 ([3fc980f], Datasource: INT
 - Browser and version [e.g. Chrome 95]

**Additional context**
Related to: https://github.com/visualize-admin/visualization-tool/issues/341

bprusinowski:
Closes #744.

### DIFF
--- a/app/charts/bar/bars-state.tsx
+++ b/app/charts/bar/bars-state.tsx
@@ -41,6 +41,7 @@ export interface BarsState {
   getSegment: (d: Observation) => string;
   segments: string[];
   xAxisLabel: string;
+  xAxisDescription: string | undefined;
   colors: ScaleOrdinal<string, string>;
 }
 
@@ -64,9 +65,10 @@ const useBarsState = ({
   const getY = useStringVariable(fields.y.componentIri);
   const getSegment = useSegment(fields.segment?.componentIri);
 
-  const xAxisLabel =
-    measures.find((d) => d.iri === fields.x.componentIri)?.label ??
-    fields.y.componentIri;
+  const xMeasure = measures.find((d) => d.iri === fields.x.componentIri);
+
+  const xAxisLabel = xMeasure?.label ?? fields.y.componentIri;
+  const xAxisDescription = xMeasure?.description || undefined;
 
   // Sort data
   const sortingType = fields.y.sorting?.sortingType;
@@ -123,6 +125,7 @@ const useBarsState = ({
     getSegment,
     segments,
     xAxisLabel,
+    xAxisDescription,
     colors,
   };
 };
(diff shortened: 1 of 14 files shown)


app/charts/bar/bars-state.tsx
```diff
@@ -64,9 +65,10 @@ const useBarsState = ({
   const getY = useStringVariable(fields.y.componentIri);
   const getSegment = useSegment(fields.segment?.componentIri);
 
-  const xAxisLabel =
-    measures.find((d) => d.iri === fields.x.componentIri)?.label ??
-    fields.y.componentIri;
+  const xMeasure = measures.find((d) => d.iri === fields.x.componentIri);
+
+  const xAxisLabel = xMeasure?.label ?? fields.y.componentIri;
+  const xAxisDescription = xMeasure?.description as string | undefined;
```
ptbrowne, line 71: Nit: Not sure the type assertion is good since we can have null right ?

```
xMeasure?.description || undefined
```

achieves the same thing without type assertion.

app/charts/shared/axis-height-linear.tsx
```diff
@@ -55,9 +58,23 @@ export const AxisHeightLinear = () => {
...
+              textDecoration="underline"
+            >
+              {yAxisLabel}
+            </text>
+          </Tooltip>
+        ) : (
+          <text x={0} y={0} dy={labelFontSize} fontSize={labelFontSize}>
+            {yAxisLabel}
+          </text>
+        )}
```
ptbrowne, line 77: Nit: A MaybeTooltip component that would do the conditional wrapping based on whether `title` exists could be cleaner since we would be sure that text elements are changed at the same time.

ptbrowne (approved):
👏 Really nice. Only 1 suggestion about the MaybeTooltip that could be tackled if you have time. I remember that I had to do the wrapping the way you did and I was a bit unconfortable with the repeated element.

ptbrowne (approved):

app/charts/shared/axis-height-linear.tsx
bprusinowski: I have added it, thanks for suggestion 🙇 Is it fine like this? I also thought about having a prop like `wrapper=((children) => ...)` on the MaybeTooltip, but then I think we would need a useCallback for each case like this to prevent re-rendering, which would also not be great I think 😅

ptbrowne: Yes I think this is fine ! Thanks !

PR #747, opened 2022-09-29
Using playwright for tests

ptbrowne:
Using playwright instead of Cypress for tests.

## Verdict

* The run of tests takes 3m37s vs 5m43s for Cypress. And one test was added taking 54s, so it's more like 2m40s for Playwright vs 5m43s for Cypress 
* I feel it is speedier while developing
* The API while necessitating quite a lot of `await` is closer to Javascript, instead of the DSL of puppeteer.  For the test checking the number of search results of a category, then clicking on the category to see if it contains the same number of results, I am not sure how I would have done that in Cypress, maybe with an alias, while in Puppeteer, it is straightforward as you only have to assign to a variable.


e2e/filters.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+  await selectors.chart.loaded(screen, page);
+
+  const filters = await selectors.edition.configFilters(screen);
+
+  await within(filters).findByText("1. production region", undefined, {
+    timeout: 30 * 1000,
+  });
+  await within(filters).findByText("2. stand structure");
+  await within(filters).findByText("3. Inventory");
+  await within(filters).findByText("4. evaluation type");
```
bprusinowski, line 21: So here it's basically a hidden "expect text to not be undefined" and it would fail if the text is not found?

e2e/fixtures/map-waldflasche-chart-config.json
```diff
@@ -1,6 +1,6 @@
 {
   "state": "CONFIGURING_CHART",
-  "dataSet": "https://environment.ld.admin.ch/foen/nfi/49-19-None-None-44/cube/1",
+  "dataSet": "https://environment.ld.admin.ch/foen/nfi/49-19-44/cube/1",
```
bprusinowski, line 3: 👍

bprusinowski: I think this config can be also enabled [here](https://github.com/visualize-admin/visualization-tool/blob/playwright/app/test/__fixtures/config/int/configs.ts#L31:L36) and changed [here](https://github.com/visualize-admin/visualization-tool/blob/playwright/app/test/__fixtures/config/int/map-waldflasche.json#L5)?

playwright.config.ts
```diff
@@ -23,21 +23,20 @@ const config: PlaywrightTestConfig = {
...
   /* Run tests in files in parallel */
-  fullyParallel: true,
+  fullyParallel: false,
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : undefined,
+  workers: process.env.CI ? 1 : 1,
```
bprusinowski: 👓

bprusinowski (approved):
LGTM! 👍

I like Cypress' chained approach to writing tests (at least I had such impression 😅) more than Playwright's awaits, but if it's more performant then let's go with this approach 👌

e2e/edition.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+  screen,
+  within,
+}) => {
+  const key = "WtHYbmsehQKo";
+  const config = offentlicheAusgabenChartConfigFixture;
+  await loadChartInLocalStorage(page, key, config);
+  await page.goto(`/en/create/${key}`);
+  await selectors.chart.loaded(screen, page);
+
+  // await page.waitForLoadState("networkidle");
```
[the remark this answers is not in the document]
ptbrowne: :+1:

e2e/fixtures/map-waldflasche-chart-config.json
ptbrowne: Done

e2e/chart-snapshots.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+    height: 1024,
+  },
+  "iphone-8, portrait": {
+    width: 375,
+    height: 667,
+  },
+};
+
+test("Chart Snapshots", async ({ page }) => {
+  test.setTimeout(5 * 60 * 1000);
```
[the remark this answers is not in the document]
ptbrowne: Because the Heavy metal dataset is very long to load, and all the charts were tested in the same test, since Cypress was slow to start new tests. 

Now each chart is is its own test, which is better for reporting failures, and lets us adjust to timeout to 1m (which is very long, but we have to do that until we figure out the performance problems of the Heavy metal dataset).

ptbrowne:
> I like Cypress' chained approach to writing tests (at least I had such impression 😅) more than Playwright's awaits

The problem I see with Cypress is that to support the wish of not writing await, they built a whole DSL (should.have.length instead of expect(.length).toBe) and new ways to save values into variables (aliases), instead of just using plain javascript.

This makes the mental model of Cypress harder (what is synchronous, what is not ?), and a new thing to learn whereas Playwright is just Javascript.

I agree though that sometimes the multiple awaits are a bit of a pain, at least they are added automatically by VSCode, so there's that.

Issue #719, opened 2022-09-16
Error when visualizing the EIV data as a scatterplot or pie chart


sosiology:
**Describe the bug**
I get an error when I try to visualize the EIV data as scatterplot or pie chart. I guess the data is not adapted to such plots, but the I should not get stuck with an inresponsive app.

**To Reproduce**
1. Go to 'https://test.visualize.admin.ch/en/browse?includeDrafts=false&search=EIV&order=SCORE&dataSource=Int'
2. Select the data set "Einmalvergütung für Photovoltaikanlagen"
3. click "Start a visualization"
4. Select "scatterplot" or "pie chart"
5. See error "Application error: a client-side exception has occurred (see the browser console for more information)."

**Expected behavior**
The chart types "Scatter plot" or "pie chart" should not be selectable, if it is not possible to visualize the data with these chart types

**Screenshots**
![image](https://user-images.githubusercontent.com/29289679/190581472-77e77f65-1125-4a70-a917-5be9e1b1d9ef.png)

**Environment (please complete the following information):**
 - Visualize environment and version: TEST v3.8.0, Data Source: INT
 - Browser and version: Chrome

**Additional context**
Also reproducable on INT v3.8.0 ([9b6863c](https://github.com/visualize-admin/visualization-tool/commit/9b6863cc990903bd81c2348d79484f78c1551915)) with Data Source: INT

PR #749, opened 2022-09-30
fix: Scatterplot issues with optional segment

bprusinowski:
For some time now I've encountered a problem with Einmalvergütung für Photovoltaikanlagen dataset when I wanted to make a scatterplot chart with it. After an investigation, it turned out that there are no Nominal or Ordinal dimensions in the dataset, but the logic to get initial chart config expected these dimension types to always be present.

Additionally, there was another issue – the possibility of selecting scatterplot chart was only based on a condition of "number of measures higher than 1" (so a case of no Nominal or Ordinal dimensions would again break the chart even if above issue was fixed, because scatterplot state logic also contained assumptions that the segment is there [e.g. missing optional chaining in some places]).

One way of fixing it would be just adjusting the logic to **actually** accept undefined segment, but then we would end up with a chart with only one observation (as in other chart types we have an additional dimension-based "axis", let it be Temporal axis for line & area charts or Nominal X axis for column chart), which allow the user to see bigger number of observations, while in double-measure based scatterplot chart, if no segment was there, we didn't have this additional breakdown.

Taking all these concerns into account, this PR proposes to make segment field mandatory in case of scatterplot chart. Another option to consider would be to somehow allow the user to specify this "hidden dimension" to include within a chart if they didn't want a color assigned to the chart (but just would like to see it when hovering over data points) – but I think it would require some more thought into it, also from design side.

I don't think there is a need to include this in chart config migrations, as if somebody created a scatterplot chart without color segmentation, it would only show one observation anyway.


ptbrowne:
I do not understand why we would only have 1 measure if there was no segment. If there is no segment, and two measures, there would be points with only 1 color no ? Am I missing something ?

image

Related issue : https://github.com/visualize-admin/visualization-tool/issues/719

ptbrowne (approved):
👍

PR #754, opened 2022-09-30
fix: Pie chart initialization

### LINKED ISSUE #719: Error when visualizing the EIV data as a scatterplot or pie chart

bprusinowski:
Fixes #719.

This PR updates the pie chart to also use geoDimension when it's initialized. Also there is a correction to possible chart types logic.


ptbrowne:
Do you think it would make sense to add a test to check the chart initialization logic ? Like open the editor for the dataset and checking that all the possible chart types button are not disabled ?

bprusinowski:
I think a test is definitely a good idea, but I wonder if this should be an e2e test, or would a unit test be enough – I am not sure if there is benefit in having an e2e test for a given dataset, when the logic to display the buttons is solely dependent on one function (`getPossibleChartType`)?

ptbrowne:
If the original issue would have been caught by an unit test, I think this is the best way to go :)

PR #757, opened 2022-10-04
fix: Migrate chart config in embed mode

bprusinowski:


app/pages/embed/[chartId].tsx
```diff
@@ -26,7 +27,18 @@ export const getServerSideProps: GetServerSideProps<PageProps> = async ({
...
+        status: "found",
+        config: {
+          ...config,
+          data: {
+            ...config.data,
+            chartConfig: migrateChartConfig(config.data.chartConfig),
+          },
+        },
+      },
+    };
```
Comment (ptbrowne) [inline app/pages/embed/[chartId].tsx:41]: Could we have the migration directly in getConfig ? This way we would not have to repeat it everywhere we use getConfig ?

ptbrowne (approved):

Issue #734, opened 2022-09-22
Colored Map with Ordinal Scale Measure


sosiology:
### User Story
As a user I would like to show a colored map from my ordinal scale measure (e.g. low, medium, high).\

### Definition of Done
Ordinal measures can be used as color mapping for map chart layers – with `schema:name` and `schema:description` appearing in the tooltip and colors that are pre-defined based on potential `schema:color`.

### Additional Info
* [Example dataset](https://test.visualize.admin.ch/en/browse/dataset/https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Ftest_ordinal_507%2F7)
* Take the following into account:
  * Sort order (i.e. severity) according to `schema:position`
  * Typical attributes shall be displayed on hovering (currently, NaNs see [example](https://test.visualize.admin.ch/en/v/AP2HkMGFzw8w))
    * `schema:name @lang`
    * `schema:description @lang`
  * For the corresponding table view `schema:name` will be enough (same as implemented already for nominal key dimensions in table view)
* Color of levels can differ case-by-case, i.e. danger case here [green, yellow and red](https://www.naturgefahren.ch) while non danger may select other colors

![image](https://user-images.githubusercontent.com/29289679/191741522-64d4e648-7c9a-4072-8513-ed223ea254c9.png)
![image](https://user-images.githubusercontent.com/29289679/191741557-537e7794-9472-4360-967f-6a67f59ed7e4.png)

adintegra:
I couldn't pinpoint the problem exactly (yet), but somewhere when switching from ord1 to ord2 and back to ord1, it seems the filter states get out of sync with what is displayed on the map – at the latest, this is the case at about 0:50 in the video.



https://user-images.githubusercontent.com/6533245/197601655-db66cbe8-071b-4f1f-b64e-9580eab56a70.mp4

bprusinowski:
Thanks again @adintegra for testing and catching this! I moved the issue to #803 and created a PR (#804) which should fix this.

PR #760, opened 2022-10-05
feat: Add support for Ordinal Measures

### LINKED ISSUE #733: Use Measures with Ordinal Scales
### User Story
As a user, I would like Visualize to properly parse measures with ordinal scale type defined in the Cube Creator.

### Definition of Done
Measures with ordinal scale type are treated differently than regular, numeric measures – e.g. it's not possible to use them as linear axis (for example as values in the bar chart).

### Additional info
This issue is about adding technical support for measures with ordinal scale types; tackling it will enable further advancements (such as #734).

Also see [dataset with ordinal measures](https://test.visualize.admin.ch/en/browse/dataset/https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Ftest_ordinal_507%2F7?previous=%7B%22includeDrafts%22%3Atrue%2C%22order%22%3A%22SCORE%22%2C%22search%22%3A%22ord%22%7D&dataSource=Int).

bprusinowski:
Closes #733.

This PR adds support for measures with ordinal scale type. From what I understood from [Cube Schema docs](https://cube.link/#qudt-scaletype) and [Cube Designer docs](https://github.com/zazuko/cube-creator/wiki/3.-Cube-Designer#dimension-type) we need to have a distinction between ordinal dimensions and ordinal measures, mostly because ordinal dimensions are ordinal because of convenience (e.g. you want the Zurich canton to appear on top of other cantons, thus you add ordering) while ordinal measures are ordinal because of necessity (e.g. low, medium or high).

At this point I am not 100% sure if we **really** need to distinguish between them in Visualize – but there seems to be at least one use case where this will be required (we don't allow to color the map by using ordinal dimension, but we want to allow such coloring by ordinal measure in #734).

To make it easier I've also renamed current `Measure` to `NumericalMeasure`.

## How to test

- [x] Check `/en/browse/dataset/https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Ftest_ordinal_507%2F7?previous=%7B%22includeDrafts%22%3Atrue%2C%22order%22%3A%22SCORE%22%2C%22search%22%3A%22ord%22%7D&dataSource=Int` and see that there are no NaNs
- [x] Create a chart from above dataset and check if it's only possible to create a table.


ptbrowne (approved):
LGTM 🎉 

I think that in the places where there are d => d.__typename === "NumericalDimension", we could replace them with a shared helper, that would also be a typeguard.

ptbrowne:
Can you add an E2E test with the test dataset containing ordinals ?
I edited the description to add what I checked.
Only the table is possible to choose as a chart type, which seems OK.

bprusinowski:
Thanks for adding "How to test" section, I forgot about this 🙇 I'll add E2E test 👍

cypress/integration/ordinal-measures.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+beforeEach(() => {
+  cy.viewport("macbook-15");
+});
+
+describe("viewing a dataset with only ordinal measures", () => {
+  const key = "ePUgYyo622qS";
+  const config = testOrd507;
+
+  it("should retrieve dimension values properly", () => {
+    cy.visit(
```
ptbrowne, line 24: No need to loadChartInLocalStorage here ?

ptbrowne (approved):
🎉

cypress/integration/ordinal-measures.spec.ts
bprusinowski: Nope, here we only look at dataset preview :)

PR #764, opened 2022-10-06
feat: Sorting improvements

bprusinowski:
When I was starting to implement Ordinal Measures in maps, I noticed that we do not allow to sort Geo dimensions (by name or measure).

I think we should enable this, as essentially they are just Nominal dimensions with coordinates attached – this PR adds support for sorting Geo dimensions.

Additionally, I've noticed that sorting doesn't work properly when there are undefined values, see screenshots down below – it was related to treating undefined values as NaNs instead of -1s. Also, the switch between Pie chart and Column / Area chart (and vice versa) was resulting in blank sorting by value in the segment field (it's also fixed now).

Screenshot 2022-10-06 at 12 16 29

Screenshot 2022-10-06 at 12 16 37

### How to test
- [ ] Go to `/en/browse/dataset/https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fubd0104prod%2F5?&dataSource=Int`
  - create a new column chart
  - add segmentation by Parameter
  - set sorting by total size
  -  switch to pie chart
  - go to segmentation
  - see that it's correctly set to measure
  - repeat the same process on TEST env to see that it's blank here
- [ ] Go to `/en/browse/dataset/https%3A%2F%2Fenergy.ld.admin.ch%2Fsfoe%2Fbfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen%2F13?dataSource=Int`
  - create new column chart
  - set X axis to Standort der Anlage
  - see that it's possible to sort by this Geo dimension (and that the sorting is correct)
  - repeat the same process on TEST env to see that it's not possible to sort this axis there

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,10 @@ You can also check the [release page](https://github.com/visualize-admin/visuali
 
 ## Unreleased
 
-Nothing yet.
+- Charts:
+  - Enable sorting of Geo dimensions
+  - Maintain segment sorting type correctly when switching from / to Pie chart
+  - Fix sorting by measure when undefined values are present in the data
 
 ## [3.9.5] - 2022-10-04
 
--- a/app/charts/chart-config-ui-options.ts
+++ b/app/charts/chart-config-ui-options.ts
@@ -77,6 +77,19 @@ const SEGMENT_DIMENSION_TYPES: DimensionType[] = [
   "GeoShapesDimension",
 ];
 
+export const AREA_SEGMENT_SORTING: EncodingSortingOption[] = [
+  { sortingType: "byDimensionLabel", sortingOrder: ["asc", "desc"] },
+  { sortingType: "byTotalSize", sortingOrder: ["asc", "desc"] },
+];
+
+export const COLUMN_SEGMENT_SORTING: EncodingSortingOption[] =
+  AREA_SEGMENT_SORTING;
+
+export const PIE_SEGMENT_SORTING: EncodingSortingOption[] = [
+  { sortingType: "byMeasure", sortingOrder: ["asc", "desc"] },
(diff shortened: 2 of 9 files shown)


ptbrowne (approved):
👏 LGTM !

Regarding the segment sorting adjusters: in the future we might want to adjust a particular sorting type to another particular sorting type. We could then use a map inside the options of adjustSegmentSorting, for example to convert byMeasure to byTotalSize explicitly, instead of doing that through the default value. I think it's fine like this right now, but it is something to consider in the future.

I really like the improvements you made on the extraction of constant and types, this makes things much clearer.

Thanks for the explicit test hints, I think they would be a great basis for an automated test after we merge the playwright branch. We could make our own test actions to make it easier to write tests. Then we can directly convert your hints into a test :)

```
  create a new column chart
    add segmentation by Parameter
    set sorting by total size
    switch to pie chart
    go to segmentation
    see that it's correctly set to measure
```

becomes 

```
await actions.createChartFrom('en/browse/dataset/https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fubd0104prod%2F5?&dataSource=Int')
await actions.setEncoding('segment', 'Parameter')
await actions.setSorting('Total size')
await actions.switchToChart('pie')
expect(await selectors.edition.segmentation()).toEqual('Measure')
```

bprusinowski:
Thanks! I was thinking about such mapping (e.g. byTotalSize -> byMeasure), but I wasn't 100% sure how to approach this best (sorting types per chart type? per old chart type -> new chart type? just sorting types mapping?) and I thought that at this point this will be enough, as anyway we only have one mismatch (total size vs measure).

And for the testing actions it also makes prefect sense to have such "library" to perform common tasks. Let's try to implement this going forward 👌

PR #768, opened 2022-10-07
feat: Use generic configurator for maps

bprusinowski:
This PR tries to use generic configurator for maps:
- to be able to better manage changes to the maps configuration (and to prepare for handling of Ordinal Measures),
- to be more consistent with the way other encodings are handled (e.g. to enable segment field we have to select a dimension via a select element, while for maps we use checkbox to show layer).

Design changes approved by @AnninaWalker :)


app/utils/chart-config/versioning.ts
```diff
@@ -101,6 +101,94 @@ const migrations: Migration[] = [
...
+    areaLayer? {
+      - show
+    }
+    
+    symbolLayer? {
+      - show
+    }`,
+    from: "1.0.2",
+    to: "1.1.0",
+    up: (config: any) => {
```
ptbrowne, line 118: any ?

app/charts/chart-config-ui-options.ts
```diff
@@ -1,76 +1,78 @@
...
-  map: ChartSpec;
 }
 
-/**
- * @todo
- * - Differentiate sorting within chart vs. sorting legend/tooltip only
- */
-
-const SEGMENT_DIMENSION_TYPES: DimensionType[] = [
+const SEGMENT_COMPONENT_TYPES: ComponentType[] = [
```
ptbrowne, line 75: Why the renaming with `COMPONENT`, I think "DIMENSION" is clearer actually.

app/charts/chart-config-ui-options.ts
```diff
@@ -1,76 +1,78 @@
...
-  | "TemporalDimension";
+export type EncodingFieldType =
+  // Generic.
+  | "x"
+  | "y"
+  | "segment"
+  // Map.
+  | "baseLayer"
+  | "areaLayer"
+  | "symbolLayer";
```
ptbrowne, line 21: I think it would be good to have

```
type MapEncodings = 'baseLayer' | 'areaLayer' | 'symbolLayer'
type XYEncodings = 'x' | 'y' | 'segment'
type Encodings = MapEncodings | XYEncodings
```

This way we can use the granularity when we need it, and we do not need the comments.

app/charts/chart-config-ui-options.ts
```diff
@@ -229,12 +230,14 @@ export const chartConfigOptionsUISpec: ChartSpecs = {
...
         optional: true,
         componentTypes: ["GeoShapesDimension"],
+        allowOtherFieldComponents: true,
         filters: true,
       },
       {
         field: "symbolLayer",
         optional: true,
         componentTypes: ["GeoCoordinatesDimension", "GeoShapesDimension"],
+        allowOtherFieldComponents: true,
```
ptbrowne, line 257: I am not sure to understand clearly what this means, could you add a docstring on the field in the interface ?
What do you think of `exclusive: false` ?

```patch
- allowOtherFieldComponents?: boolean;
+ /** If false, using a dimension in this encoding will not prevent it to be used in an other encoding. Default: true */
+ exclusive?: boolean
```

app/charts/map/map.tsx
```diff
@@ -365,6 +355,14 @@ export const MapComponent = () => {
...
             setMap(e.target);
             currentBBox.current = e.target.getBounds().toArray() as BBox;
+
+            /**
+             * Redraw the map on load, when style data has been downloaded
+             * to solve an offset problem between the viz layer and the base layer
+             * happening in Edge under precise conditions (Windows 10, Edge 105,
+             * resolution 1440x900).
+             */
+            redrawMap();
```
ptbrowne, line 365: I hope you tested this under Edge :)

app/charts/chart-config-ui-options.ts
```diff
@@ -235,13 +241,29 @@ export const chartConfigOptionsUISpec: ChartSpecs = {
...
         componentTypes: ["GeoCoordinatesDimension", "GeoShapesDimension"],
         allowOtherFieldComponents: true,
         filters: true,
+        options: [
+          {
+            field: "color",
+            type: "component",
+            componentTypes: ["NumericalMeasure", "GeoShapesDimension"],
+            optional: true,
+          },
```
ptbrowne, line 265: Should/will there be a "size" field here ?

app/charts/chart-config-ui-options.ts
```diff
@@ -235,13 +241,29 @@ export const chartConfigOptionsUISpec: ChartSpecs = {
...
       {
         field: "symbolLayer",
         optional: true,
         componentTypes: ["GeoCoordinatesDimension", "GeoShapesDimension"],
         allowOtherFieldComponents: true,
         filters: true,
+        options: [
+          {
+            field: "color",
+            type: "component",
```
ptbrowne, line 262: "dimension" would be easier to understand I think.

ptbrowne (commented):
Wow, what a beast of a PR. Congrats on generalizing so that we do not have a specific map configurator.

When looking at the preview and clicking on Map Display I have a select with "None" and "(optional)", I do not think this is wanted. I guess we would need to support the checkbox type  from the encoding spec, and mark this field as required ?

app/charts/chart-config-ui-options.ts
bprusinowski: Yes 👌

app/charts/map/map.tsx
bprusinowski: 👁️

ptbrowne:
I do not see the former bugs that we had on maps with the canton borders being offseted on Edge 🎉

e2e/symbol-layer-colors.spec.ts
```diff
@@ -25,19 +27,27 @@ test("Selecting SymbolLayer colors> should be possible to select nominal dimensi
...
-    expect(await select.getAttribute("class")).toContain("Mui-disabled");
-  }
+  expect(count).toEqual(1);
+
+  const panelRight = await selectors.panels.right(ctx);
+  await (await within(panelRight).findByText("None")).click();
+  // Select options open in portal
+  await screen
+    .locator('li[role="option"]:has-text("production region")')
+    .click();
```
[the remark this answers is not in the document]
ptbrowne: `await actions.editor.selectEncodingDimension('Symbols', 'production region')` ?

e2e/symbol-layer-colors.spec.ts
```diff
@@ -25,19 +27,27 @@ test("Selecting SymbolLayer colors> should be possible to select nominal dimensi
...
 
-  await (await screen.findByText("Show layer")).click();
-  await (await screen.findByText("None")).click();
+  // allow select options to disappear to prevent re-selecting none
+  await sleep(3000);
```
ptbrowne, line 40: I think this should be done by another mean `await selectors.menu.closed()' ?

ptbrowne:
LGTM, 👏 , I've added suggestions to improve selectors/actions in test. Feel free to move this to another issue, if it cannot be tackled right now.

ptbrowne (approved):

PR #777, opened 2022-10-13
High filter value count

### LINKED ISSUE #717: oddly data loading error appears - only for this specific configuration
**Describe the bug**
oddly a data loading error appears - but only for this specific configuration...

**To Reproduce**
1. Go to https://test.visualize.admin.ch/create/new?from=fqheWcBWiQ3z&dataSource=Int
2. see error

or also reproducible by
1. Go to https://test.visualize.admin.ch/create/new?cube=https://environment.ld.admin.ch/foen/fab_hierarchy_test12_switzerland_canton_municipality/3
2. Goto 'color'
3. Select 'BFSIdentifier'

**Expected behavior**
bar chart colorized by 'BFSIdentifier'

**Screenshots**
image

 - Visualize environment and version: TEST 3.8.0
 - Browser and version [e.g. Chrome 95]

**Additional context**
bar chart (single color), map, table, etc. works fine.

cc @fabiancretton

### LINKED ISSUE #722: map doesn't load
**Describe the bug**
map doesn't load

**To Reproduce**
1. Go to https://test.visualize.admin.ch/v/SUlonFeYMavu?dataSource=Int
2. See map doesn't load

**Expected behavior**
Show map as for same dataset w/o hierarchy (https://test.visualize.admin.ch/v/oUxguf5fe1z5?dataSource=Int)

**Screenshots**
image

 - Visualize environment and version: TEST v3.8.0
 - Browser and version [e.g. Chrome 95]

ptbrowne:
When a filter had too much values in it, the HTTP query to Stardog
would fail due to a too large HTTPPayload (since the filter values
were inside the query).

Now if a filter has too much values, the filtering on our end. Too much here is > 100.

We will fetch more data from SPARQL, but I do not see another way.

Presumably, when we have more than 100 values in a filter, from a user
point of view, we want more values *in* than *out*, so hopefully the
number of values  from the SPARQL result that will not be used, will be low.

### How to test

- Use a dataset with a high number of values inside a filter (for example, this dataset with the population for each cantons of Zurich: `/en/browse/dataset/https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Ffab_hierarchy_test13_switzerland_canton_municipality%2F3?dataSource=Int`)
- See if map is correctly displayed

Fix #722 
Fix #717

### DIFF
--- a/app/rdf/queries.ts
+++ b/app/rdf/queries.ts
@@ -397,15 +397,13 @@ export const getCubeObservations = async ({
 }> => {
   const cubeView = View.fromCube(cube);
 
-  /**
-   * Add labels to named dimensions
-   */
-  const allCubeDimensions = await getCubeDimensions({
+  const allResolvedDimensions = await getCubeDimensions({
     cube,
     locale,
     sparqlClient,
   });
-  const cubeDimensions = allCubeDimensions.filter((d) =>
+
+  const cubeDimensions = allResolvedDimensions.filter((d) =>
     dimensions ? dimensions.includes(d.data.iri) : true
   );
 
--- a/app/rdf/queries.ts
+++ b/app/rdf/queries.ts
@@ -19,6 +19,7 @@ import { Filters } from "../configurator";
 import {
   DimensionValue,
   Observation,
+  ObservationValue,
   parseObservationValue,
   shouldValuesBeLoadedForResolvedDimension,
 } from "../domain/data";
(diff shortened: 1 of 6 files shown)


app/graphql/context.tsx
```diff
@@ -69,6 +69,18 @@ const createContextContent = async ({
...
+  //   (t, ...args: Parameters<typeof sparqlClientStream.query.select>) => {
+  //     console.log("select on sparql client stream", args[0]);
+  //     ctx.queries.push({
+  //       startTime: t.start,
+  //       endTime: t.end,
+  //       text: args[0],
+  //     });
+  //   }
+  // );
+
```
bprusinowski, line 83: 👓

bprusinowski (approved):
LGTM! I like the refactors on the RDF side 👌

ptbrowne:
Cool. For information: we had discussed about splitting the filters and making multiple SPARQL queries to fetch the results, but I thought that it would not work if several high cardinality filters were present since we would not know which "intersection" of those filters would return results. This is why I applied the filtering server side.

cc @l00mi, I am interested to know if you have any thoughts on this subject, and if you have a better idea for making high cardinality filters work.

PR #780, opened 2022-10-14
Add latest update date to chart footnotes

### LINKED ISSUE #759: Show the date of the latest data update on chart
- Design: See if it makes sense to have an info panel, with this information + values & dimension description
- Development: 1st add the information in the footnotes, then see if we need to make an info panel

ptbrowne:
Chart update in footnotes

Show when the data for the current chart has been updated.
Fix #759

### DIFF
--- a/app/components/chart-footnotes.tsx
+++ b/app/components/chart-footnotes.tsx
@@ -108,7 +108,7 @@ export const ChartFootnotes = ({
       dataCubeByIri.dateModified
         ? t({
             id: "dataset.metadata.date.footnotes.updated",
-            message: `, latest update: ${formatLocale.format("%x %X")(
+            message: `, latest update: ${formatLocale.format("%x %H:%M")(
               new Date(dataCubeByIri.dateModified)
             )}`,
           })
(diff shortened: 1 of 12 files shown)


app/components/chart-footnotes.tsx
```diff
@@ -95,28 +96,40 @@ export const ChartFootnotes = ({
...
   }, [locale, data?.dataCubeByIri]);
 
   if (data?.dataCubeByIri) {
     const { dataCubeByIri } = data;
 
+    const titleContent = `${dataCubeByIri.title}${
+      dataCubeByIri.dateModified
+        ? t({
+            id: "dataset.metadata.date.footnotes.updated",
+            message: `, latest update: ${formatLocale.format("%x %X")(
```
bprusinowski, line 111: TIL about `%x` time formats 📆

bprusinowski (approved):
LGTM :)

app/components/chart-footnotes.tsx
ptbrowne: I changed the %X to %H:%M otherwise we have seconds in the output and this does not seem helpful. There is no equivalent of %X without seconds https://github.com/d3/d3-time-format/issues/34.

PR #785, opened 2022-10-18
Bold footnotes label

ptbrowne:
- feat: Use bold for footnotes labels, and use swiss date format

I used a translation string for the colon since in French, there is a space before the colon.

Fixes https://github.com/orgs/visualize-admin/projects/1/views/18


bprusinowski (approved):
👌

PR #786, opened 2022-10-19
feat: Automatic sort based on identifier and position

ptbrowne:
Added a new sort option called "Automatic" for column / lines / area and pie chart. It uses either
the identifier or the position of the value to sort the values.
Legend is sorted according to this and also the tooltip content values.

Technically, I've tried to group the sorting logic that was dispersed
across chart to be able to more easily add the new sorting method.

- [ ] ~~Add alternate name as a sorting option~~ Will do that in another PR
- [x] Add E2E tests

## How to test

- Go to the preview deployment (See "Visit preview" link below)
- Go to the photovoltaik dataset *on int*: `/en/create/new?cube=https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/13&dataSource=Int`
- Select {areas,pie,column,lines} charts
- Go the segment and pick "Standort der Anlage" to color
- "Automatic" should be the default selected options, Zurich should come first, Jura should come last
- This should be visible in the chart, the legend, and the tooltip options
- You have the ability to revert the sort
- You should be able to select "Name" as a sorting option

Original issue : https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/119

Fixes https://github.com/visualize-admin/visualization-tool/issues/789

### DIFF
--- a/app/charts/column/columns-grouped-state.tsx
+++ b/app/charts/column/columns-grouped-state.tsx
@@ -141,7 +141,6 @@ const useGroupedColumnsState = (
 
   // Sort
   const xSorting = fields.x.sorting;
-  const xSortingOrder = fields.x.sorting?.sortingOrder;
 
   // Group by X
   const sortedData = useMemo(() => {
@@ -197,19 +196,24 @@ const useGroupedColumnsState = (
     const dimension = dimensions.find(
       (d) => d.iri === fields.segment?.componentIri
     );
+
+    const sorting = fields?.segment?.sorting;
     const sorters = makeDimensionValueSorters(dimension, {
-      sorting: fields?.segment?.sorting,
+      sorting,
       sumsBySegment,
     });
 
-    return orderBy(uniqueSegments, sorters, xSortingOrder || "asc");
+    return orderBy(
+      uniqueSegments,
+      sorters,
+      sorting?.sortingOrder === "desc" ? "desc" : "asc"
+    );
   }, [
     plottableSortedData,
     dimensions,
     fields.segment?.sorting,
     fields.segment?.componentIri,
     sumsBySegment,
-    xSortingOrder,
     getSegment,
   ]);
 
--- a/app/charts/column/columns-grouped-state.tsx
+++ b/app/charts/column/columns-grouped-state.tsx
@@ -266,10 +266,9 @@ const useGroupedColumnsState = (
     const xScaleIn = scaleBand().domain(segments).padding(PADDING_WITHIN);
 
     // x as time, needs to be memoized!
-    const xEntireDomainAsTime = extent(sortedData, (d) => getXAsDate(d)) as [
-      Date,
-      Date
-    ];
+    const xEntireDomainAsTime = extent(plottableSortedData, (d) =>
+      getXAsDate(d)
+    ) as [Date, Date];
     const xEntireScale = scaleTime().domain(xEntireDomainAsTime);
 
     // y
@@ -304,13 +303,13 @@ const useGroupedColumnsState = (
   }, [
     dimensions,
     fields.segment,
-    getX,
-    getXAsDate,
-    getY,
-    getYErrorRange,
-    sortedData,
     preparedData,
     segments,
+    plottableSortedData,
+    getX,
+    getXAsDate,
+    getYErrorRange,
+    getY,
   ]);
 
   const yMeasure = measures.find((d) => d.iri === fields.y.componentIri);
@@ -446,7 +445,7 @@ const useGroupedColumnsState = (
   return {
     chartType: "column",
     preparedData,
-    allData: sortedData,
+    allData: plottableSortedData,
     bounds,
     getX,
     getXAsDate,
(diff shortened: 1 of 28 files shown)


AnninaWalker:
Currently it seems that in the bar chart column the Sorting from A-Z and Z-A does not seem to work. This is not only for the "Automatic" Sorting but also for the "Name" Sorting

app/rdf/queries.ts
```diff
@@ -234,7 +234,7 @@ export const getCubeDimensionValues = async (
     return [];
   }
 
-  return await getCubeDimensionValuesWithLabels({
+  return await getCubeDimensionValuesWithMetadata({
```
bprusinowski, line 237: 👍

app/charts/column/columns-grouped-state.tsx
```diff
@@ -287,9 +263,10 @@ const useGroupedColumnsState = (
     const xScaleIn = scaleBand().domain(segments).padding(PADDING_WITHIN);
 
     // x as time, needs to be memoized!
-    const xEntireDomainAsTime = extent(plottableSortedData, (d) =>
-      getXAsDate(d)
-    ) as [Date, Date];
+    const xEntireDomainAsTime = extent(sortedData, (d) => getXAsDate(d)) as [
+      Date,
+      Date
+    ];
```
bprusinowski, line 272: Why `sortedData` and not `plottableSortedData` as before? The same comment for lines 308 and 446 😅

app/charts/pie/pie-state.tsx
```diff
@@ -100,51 +73,64 @@ const usePieState = (
...
+  const plottableData = usePlottableData({
+    data: data,
     plotters: [getY],
   });
 
   // Apply end-user-activated interactive filters to the stack
   const preparedData = usePreparedData({
     legendFilterActive: interactiveFiltersConfig?.legend.active,
-    sortedData: plottableSortedData,
+    sortedData: data,
```
bprusinowski, line 87: `plottableData`?

app/charts/pie/pie-state.tsx
```diff
@@ -100,51 +73,64 @@ const usePieState = (
...
   // Map ordered segments to colors
   const colors = useMemo(() => {
     const colors = scaleOrdinal<string, string>();
-    const segments = Array.from(
-      new Set(plottableSortedData.map((d) => getX(d)))
+    const measureBySegment = Object.fromEntries(
+      plottableData.map((d) => [getX(d), getY(d)])
+    );
+    const uniqueSegments = Object.entries(measureBySegment)
+      .filter((x) => x[1])
```
bprusinowski, line 117: `.filter(x[1] !== null && x[1] !== undefined)`? With the current logic it would filter out 0s right? Or was this intended, as anyway we can't see 0s on pie chart (but on the other hand we still we could see it in a legend)?

app/charts/pie/pie-state.tsx
```diff
@@ -158,8 +144,19 @@ const usePieState = (
       colors.domain(segments);
       colors.range(getPalette(fields.segment?.palette));
     }
+    // Do not let the scale be implicitly extended by children calling it
+    // on unknown values
+    colors.unknown(() => undefined);
```
bprusinowski, line 149: Do you know in which cases can this happen?

app/rdf/queries.ts
```diff
@@ -318,51 +318,39 @@ const getCubeDimensionValuesWithMetadata = async ({
...
+              ? schema.identifier
+              : null,
+          position: scaleType === "Ordinal" ? schema.position : null,
+          color: scaleType === "Ordinal" ? schema.color : null,
+        },
+      }),
+      dimensionIsVersioned(dimension)
+        ? loadUnversionedResources({ ids: namedNodes, sparqlClient })
+        : [],
+    ]);
```
bprusinowski, line 341: 👌

e2e/sorting.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    // Wait for chart to be loaded
+    await selectors.chart.loaded();
+    await selectors.edition.filtersLoaded();
+    await selectors.chart.colorLegend(undefined, { setTimeout: 5_000 });
+
+    const legendItems = await selectors.chart.colorLegendItems();
+    const legendTexts = await legendItems.allInnerTexts();
+    expect(legendTexts[0]).toEqual("Zurich");
+
+    if (chartType !== "Lines") {
```
bprusinowski, line 41: I think this condition also can be removed with the introduction of sorting to line chart?

bprusinowski (approved):
LGTM, nice improvements 🎉

app/charts/pie/pie-state.tsx
ptbrowne: Good point :+1:

app/charts/pie/pie-state.tsx
ptbrowne: Yep :)

app/charts/pie/pie-state.tsx
ptbrowne: Not really, I just noticed that we had extra entries in the legend 🤷‍♂️

app/charts/column/columns-grouped-state.tsx
ptbrowne: No reason, thanks for noticing.

e2e/sorting.spec.ts
ptbrowne: Done

bprusinowski (approved):
LGTM, thanks!

PR #787, opened 2022-10-19
fix: Broken color mapping for area layer's categorical color field

bprusinowski:
When implementing new color field for maps, I missed one place in the code which handled proper mapping of filters. This PR fixes this issue.

### How to test
1. Go to `/en/browse/dataset/https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Ftest_ordinal_507%2F7?dataSource=Int`
2. Create a map chart.
3. Activate areaLayer.
4. Select `ord1` as a color mapping dimension.
5. Open filter panel for colors and de-select one value (any).
6. Close the panel and see that the filters work.
7. Repeat the same process on TEST to see how they break currently.

### DIFF
--- a/app/configurator/configurator-state.tsx
+++ b/app/configurator/configurator-state.tsx
@@ -34,6 +34,7 @@ import {
   DataSource,
   ImputationType,
   isAreaConfig,
+  isColorFieldInConfig,
   isColumnConfig,
   isMapConfig,
   isSegmentColorMappingInConfig,
@@ -368,8 +369,10 @@ export const deriveFiltersFromFields = produce(
       const isField = (iri: string) => fieldDimensionIris.has(iri);
 
       // Apply hierarchical dimensions first
-      const sortedDimensions = sortBy(dimensions, d => isGeoDimension(d) ? -1 : 1, (d) =>
-        d.hierarchy ? -1 : 1
+      const sortedDimensions = sortBy(
+        dimensions,
+        (d) => (isGeoDimension(d) ? -1 : 1),
+        (d) => (d.hierarchy ? -1 : 1)
       );
       sortedDimensions.forEach((dimension) =>
         applyNonTableDimensionToFilters({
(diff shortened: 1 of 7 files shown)


app/configurator/configurator-state.tsx
```diff
@@ -636,15 +639,32 @@ export const canTransitionToPreviousStep = (_: ConfiguratorState): boolean => {
   return true;
 };
 
+// FIXME: should by handled better, as color is a subfield and not actual field.
+// Side effects in ui encodings?
```
ptbrowne, line 643: Why is it important if color is a subfield or an actual field ? From I what I understand, the goal here is to extract the needed filters based on the current IRIs used in the config, and we need special treatment for IRIs used in color mapping because they are not generic. Do I understand correctly ? In this case why does it matter if it's a field or a subfield ?

 Can you add a docstring to getFiltersByMappingStatus so that when we return to it later, it's easier to understand ?

app/configurator/configurator-state.tsx
```diff
@@ -636,15 +639,32 @@ export const canTransitionToPreviousStep = (_: ConfiguratorState): boolean => {
...
 export const getFiltersByMappingStatus = (chartConfig: ChartConfig) => {
   const genericFieldValues = Object.values(chartConfig.fields).map(
     (d) => d.componentIri
   );
-  const nonGenericFieldValues =
-    isMapConfig(chartConfig) &&
-    chartConfig.fields.symbolLayer?.color.type === "categorical"
-      ? [chartConfig.fields.symbolLayer.color.componentIri]
-      : [];
+  const nonGenericFieldValues = extractCategoricalColorIris(chartConfig);
```
ptbrowne, line 666: Nit: I would call the function extractNonGenericFields, I think it could handle other cases than only the categorical color. Here my brain does "why the left is called nonGenericField and the right is called categorical colors".

ptbrowne (approved):
Looks good, end to end testing ? I think now that we have the infrastructure, they are fairly easy to add ;)

app/configurator/configurator-state.tsx
bprusinowski: Good point 👍

app/configurator/configurator-state.tsx
ptbrowne: I see, thanks for the explanation, I'll try to keep an eye for this and see if I think of a better way :)

ptbrowne (approved):
👍 Thanks

PR #791, opened 2022-10-20
feat: Open up color formats

bprusinowski:
As we use D3's `color` function to parse color strings, we are free to use any CSS Color Module Level 3 compatible strings to specify colors.

In the test dataset provided by @Rdataflow ([test_ord_507](https://int.cube-creator.lindas.admin.ch/app/cube-projects/cube-project!!testcolor507d-znxzma1yb17/)) several different conventions are used and all of them work; however, when trying to edit the pre-defined colors in visualize, it's not possible to use the original conventions if it's not a HEX color.

This PR simplifies the color logic to remove the constraint of having to use HEX colors in the color picker.


ptbrowne (approved):
👍

PR #792, opened 2022-10-20
fix: Resetting color palette for color field with pre-defined colors

bprusinowski:
Refreshing of colors in multi filter panel wasn't working for map's area layer – this PR makes the handling of color mapping updates more generic (by using `field` + `colorPathField`).

It also introduces a distinction between reshuffling colors (default) and resetting colors (when a dimension has pre-defined colors coming from LINDAS).

### DIFF
--- a/app/configurator/components/filters.tsx
+++ b/app/configurator/components/filters.tsx
@@ -64,6 +64,7 @@ import SvgIcCheck from "@/icons/components/IcCheck";
 import SvgIcChevronRight from "@/icons/components/IcChevronRight";
 import SvgIcClose from "@/icons/components/IcClose";
 import SvgIcFormatting from "@/icons/components/IcFormatting";
+import SvgIcRefresh from "@/icons/components/IcRefresh";
 import SvgIcSearch from "@/icons/components/IcSearch";
 import { useLocale } from "@/locales/use-locale";
 import { dfs } from "@/utils/dfs";
@@ -355,9 +356,9 @@ const MultiFilterContent = ({
       dispatch({
         type: "CHART_CONFIG_UPDATE_COLOR_MAPPING",
         value: {
-        field: `${field}${
-          colorConfigPath !== undefined ? `.${colorConfigPath}` : ""
-        }`,
+          field: `${field}${
+            colorConfigPath !== undefined ? `.${colorConfigPath}` : ""
+          }`,
           dimensionIri,
           values: sortBy(colorComponent?.values, (d) =>
             usedValues.has(d.value) ? 0 : 1
@@ -376,6 +377,10 @@ const MultiFilterContent = ({
         : true)
     );
   }, [colorConfigPath, config, dimensionIri, colorComponent]);
+
+  const hasDefaultColors = useMemo(() => {
+    return colorComponent?.values?.[0].color !== undefined;
+  }, [colorComponent?.values]);
 
   return (
     <Box sx={{ position: "relative" }}>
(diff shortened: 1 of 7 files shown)


app/configurator/components/filters.tsx
```diff
@@ -353,6 +356,9 @@ const MultiFilterContent = ({
       dispatch({
         type: "CHART_CONFIG_UPDATE_COLOR_MAPPING",
         value: {
+          field: `${field}${
```
ptbrowne, line 359: handleResetOrRenewColorMapping ?

app/configurator/configurator-state.tsx
```diff
@@ -1107,30 +1109,26 @@ const reducer: Reducer<ConfiguratorState, ConfiguratorStateAction> = (
...
-        if (
-          isSegmentColorMappingInConfig(draft.chartConfig) &&
-          draft.chartConfig.fields.segment &&
-          draft.chartConfig.fields.segment.componentIri === dimensionIri
-        ) {
+        const { field, dimensionIri, values, random } = action.value;
+        const fieldValue = get(draft.chartConfig, `fields.${field}`) as
+          | GenericSegmentField
+          | CategoricalColorField
+          | undefined;
```
ptbrowne, line 1116: Can we try not to have this assertion ? If we had types or segment field later, this assertion will be false.

bprusinowski: Would a `isColorMappingInConfig` helper function work in this case (to check if there is a segment or color field with type of categorical)?

ptbrowne: Here we want to have a any type of field with a palette right ? What could be good would be to have a type `type ColorField = Extract<ChartFields, { palette: any }>` (example code, not working) ?

app/configurator/components/filters.tsx
```diff
@@ -345,21 +348,43 @@ const MultiFilterContent = ({
...
+          const usedValues = new Set(values.map((v) => v.value));
+          return dispatch({
+            type: "CHART_CONFIG_UPDATE_COLOR_MAPPING",
+            value: {
+              field: computedField,
+              dimensionIri,
+              values: sortBy(values, (d) => (usedValues.has(d.value) ? 0 : 1)),
+              random: true,
+            },
+          });
```
ptbrowne, line 386: I wonder here why we do not have two events ?

bprusinowski: Two `useEvents` or two action types of reducer?

app/configurator/components/filters.tsx
```diff
@@ -345,21 +348,43 @@ const MultiFilterContent = ({
...
-          ),
-          random,
-        },
-      });
+  /** Handles updates to color mapping.
+   *
+   *  An update can be either a _reset_, to go back to original colors or a _shuffle_
+   * to create a new, randomized color mapping.
+   *
+   * @param type _reset_ or _shuffle_
```
ptbrowne, line 356: This is redundant with the type definition.

app/configurator/components/filters.tsx
```diff
@@ -345,21 +348,43 @@ const MultiFilterContent = ({
...
+   */
+  const handleUpdateColorMapping = useEvent(
+    ({ type }: { type: "reset" | "shuffle" }) => {
```
ptbrowne, line 359: Nit: I think you can put directly docstrings on each type here

```
type: |
  /** Goes back to original colors from the dataset */
  'reset' |
  /** Shuffles colors to create a new ... */
  'shuffle'
```

ptbrowne (approved):
LGTM, I would just like a little test here, I feel otherwise we can break this part of the app.

app/configurator/components/filters.tsx
ptbrowne: two action types. This is a nit.

app/configurator/components/filters.tsx
```diff
@@ -357,17 +357,15 @@ const MultiFilterContent = ({
...
       const values = colorComponent?.values || [];
 
       switch (type) {
         case "reset":
           return dispatch({
             type: "CHART_CONFIG_UPDATE_COLOR_MAPPING",
             value: {
-              field: computedField,
+              field,
+              colorConfigPath,
```
ptbrowne, line 368: 👍

ptbrowne (approved):
Thanks 👍

PR #794, opened 2022-10-20
feat: Record/replay HAR for chart snapshots

ptbrowne:
Record and replay HAR network dumps for two tests that are long.

- For snapshots, we go from 100s to 30s for 10 configs tested (5 different charts, 2 viewports)
- For sorting we go from 26s to 13s


bprusinowski (approved):
Cool, LGTM! 👍

PR #796, opened 2022-10-21
Show contents of urql panel in PROD

bprusinowski:
- [x] Edit: this doesn't seem to do the trick, needs further investigation
- [x] gql flamegraph exchange is activated on production
- [x] The urql devtool was guarded on process.env.NODE_ENV === 'development' at two places, via the .dev / .prod technique and inside the the client creation. Now it's only done via the .dev / .prod technique. We prefer the .dev / .prod technique so that in production the urql devtools are not bundled.


ptbrowne (approved):

bprusinowski:
Thanks 🎉

PR #797, opened 2022-10-21
fix: Map tooltip duplicated values in symbol layer + missing legend colors

bprusinowski:
Tooltip row for radius was being shown when area layer color was using the same dimension, which resulted in duplicated rows showing the same information – this PR fixes that (and also makes it easier to see the logic for displaying tooltip rows).

Also, it fixes the issue of missing color in map color legend.

### DIFF
--- a/app/charts/map/map-tooltip.tsx
+++ b/app/charts/map/map-tooltip.tsx
@@ -67,31 +67,37 @@ export const MapTooltip = () => {
   });
 
   const areaTooltipState = useMemo(() => {
-    if (areaLayer) {
-      const obs = interaction.d;
+    const obs = interaction.d;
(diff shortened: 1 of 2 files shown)


app/charts/map/map-tooltip.tsx
```diff
@@ -69,29 +69,38 @@ export const MapTooltip = () => {
   const areaTooltipState = useMemo(() => {
     if (areaLayer) {
       const obs = interaction.d;
-      const { colors } = areaLayer;
+      if (obs) {
```
ptbrowne, line 72: Nit, I think returning early if not areaLayer or obs would be a bit clearer here.

ptbrowne (approved):

Issue #690, opened 2022-09-08
Full precision must be dispayed with Decimal Datatype


ortnever:
## Context
We would like to see the full precision of decimal/float datapoints within visualize expressed in scientific notation – either within the charts directly or when hovering over a datapoint.

## Current Behavior
Currently, the behaviour of the application is such that the full precision of the data coming from LINDAS is not displayed. Small values, in particular, are problematic, as they can be truncated to 0.

## Expected Behavior
The application should provide a means for the user to see the full precision of a datapoint – either in the chart display or, at least, in the tooltip for the datapoint.

Full precision must be dispayed for decimal numbers. It concerns table chart, mouse over, axis labels, other places? For export, already ok.

## Datasets
The dataset "Greenhouse gas emissions by sector and by gas (hierarchie Gas)" can be used as an example/ for testing. Select the Gas "Synthetic gas" and the sector "Sectors CO2 Ordinance". . - - 
- https://int.visualize.admin.ch/en/v/mDYTVQkRhvvU
- https://int.visualize.admin.ch/en/v/_vkg3ZVUnAZU

### References

- See also https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/361#note_8490 
- Corresponding GitLab Isssue: https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/408

sosiology:
Decision: Show full percision expressed as scientific notation. 
Example: https://www.covid19.admin.ch/en/epidemiologic/waste-water/d/development?wasteWaterFacility=427100&rel=abs
Also see comment in Gitlab. https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/408#note_8894

PR #799, opened 2022-10-21
Scientific notation for decimal

ptbrowne:
Fix : https://github.com/visualize-admin/visualization-tool/issues/690

For dimensions with decimal datatype, we use a scientific notation formatter, when the value is > 999 999 999 or value < 0.0001

## How to test

* See the greenhouse gas emissions (hierarchie) dataset

`/en/browse/dataset/https://environment.ld.admin.ch/foen/ubd000502test/3`

* The column "Greenhouse gas emissions in CO2eq(Mt) should be in scientific notation
* When looking at a chart, the tooltips should show number formatter in scientific notation


app/formatters.ts
```diff
@@ -112,10 +113,26 @@ const dateFormatterFromDimension = (
...
   return x !== DIMENSION_VALUE_UNDEFINED ? `${x}` : "–";
 };
 
+const decimalFormatter = (dim: NumericalMeasure, formatNumber: Formatter) => {
+  const res = dim.resolution;
+  const hasResolution = typeof res === "number";
+  const expFormatter = format(`${hasResolution ? `.${res}` : ""}~e`);
+  return (v: string) => {
+    if (v === "0" || v === "0.0") {
+      return formatNumber(v);
```
bprusinowski, line 128: In general, I am not a fan of displaying scientific notation when `e = 0`. Do you think there could be another rule introduced, to check if the number of leading zeros equals to 0 (and then apply just regular `formatNumber`?

So then we would be able to avoid cases like below, where e+0 doesn't add any information and just pollutes the UI?

Screenshot 2022-10-24 at 09 42 03

bprusinowski (approved):
Overall LGTM! 👍 

But I am not convinced about applying the scientific notation basing just on a data type (`isDecimal`). I think scientific notation makes sense for really small / large numbers (e.g. when `|e| >= 3`), otherwise it should be fine to show full numbers such as 0.002. I would be in favor of applying it specifically to only such numbers; if you don't agree, then I would try to remove it at least for cases where `e = 0`.

ptbrowne:
Yes I also thought about this, I should have clearly laid it out in the PR description.

I thought about doing checks on > 9999 or < 0.001 for example but then I thought that without ifs, the method would stay coherent and it would be easier to compare numbers (especially in a table) since they would all have the same formatting method.

It would be easy to add a rule. I'd like to have @rdataflow or @ortnerv thoughts on this ?

bprusinowski:
Yes, in cases like table it makes sense, in places where we don't compare multiple numbers, I would still stick to more user-friendly display 😅

ptbrowne:
If you brush over bars and see tooltips, I think having the display content jumping back and forth between scientific notation and "normal" notation could be distracting. In case of stacked bar chart, there is multiple values in the tooltip content, some falling into the "normal" range and some falling into the "scientific notation" range.

ptbrowne:
For me the real problem here is the number of digits in the displayed result, for example we show "5.187271e0Mt", here I think that having "5.18e0Mt" would convey the necessary information for someone viewing the chart, and I am not sure that we should show all these digits. @ortnever, could you explain what is the usecase here ?

ortnever:
Sorry for my late reaction, I'm back from holiday today. 

I agree with the proposition to define a value range for which scientific notification is not applied. It is difficult to define the limits. For the moment I would propose 0.00001 to 9'999'999. I guess we can change these limits later if desired?
The cases where we have values outside this range should be exceptions as I assume the data owner will adapt the unit if the numbers are very big or very small. 

I agree with you Patrick, that it would be nicer to display only two or three decimal but we decided, together with Thomas, that it is better to always display the full precision delivered in the data. It is up to the data provider to decide if he wants to (or can) round the values.

It also makes it possible to quickly get an exact value without having to download the data.

PR #804, opened 2022-10-25
fix: Color field filters not cleared when switching dimensions

### LINKED ISSUE #803: Color filters not cleared when switching dimensions
I couldn't pinpoint the problem exactly (yet), but somewhere when switching from ord1 to ord2 and back to ord1, it seems the filter states get out of sync with what is displayed on the map – at the latest, this is the case at about 0:50 in the video.

https://user-images.githubusercontent.com/6533245/197601655-db66cbe8-071b-4f1f-b64e-9580eab56a70.mp4

Initially, I tested on DEV (test.visualize), but similar behaviour can also be reproduced on INT. In addition, deselecting all filter values for ord2 produced this error message at one point:

Screenshot 2022-10-24 at 20 52 17

_Originally posted by @adintegra in https://github.com/visualize-admin/visualization-tool/issues/734#issuecomment-1289450980_

bprusinowski:
Fixes #803.

The filters created by a new color field weren't cleared when switching to a new color dimension. This PR fixes that.

Additionally, it fixes another bug that appeared when no dimension values were selected (undefined color assessor) + it adds fetching of un-filtered dimensions and measures to be used as `dataSetMetadata` (as in general, we need unfiltered values in configurator state).

### DIFF
--- a/app/configurator/components/chart-options-selector.tsx
+++ b/app/configurator/components/chart-options-selector.tsx
@@ -63,6 +63,7 @@ import {
 } from "@/domain/data";
 import {
   DimensionMetadataFragment,
+  useDataCubeMetadataWithComponentValuesQuery,
   useDataCubeObservationsQuery,
 } from "@/graphql/query-hooks";
 import { DataCubeMetadata } from "@/graphql/types";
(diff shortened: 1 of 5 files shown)


ptbrowne (commented):
LGTM, can we a small unit tests for this ? Making sure selecting the component as segment clears the filter.

ptbrowne (approved):
🙇

PR #807, opened 2022-10-26
fix: Crashing of charts when selecting component "too fast"

### LINKED ISSUE #788: Crash of the map when activating symbol layer
@ptbrowne the map now loads as expected in both the editor and published view. However I noticed that if I add the symbol layer an application error is shown. Below the screenshot from the browser console

Image

_Originally posted by @sosiology in https://github.com/visualize-admin/visualization-tool/issues/722#issuecomment-1280900855_

bprusinowski:
Fixes #788.

In some rare scenarios, user could re-select (or select another dimension) of a field that was still being computed.

In such cases, we would trigger an "update" of a field, which by default cleared all the contents of a field, keeping only `componentIri`, which led to errors with the map charts, as in other places we were expecting `(area | symbol)Layer.color` to be defined.

This PR fixes this issue by populating the color field for map charts while the field is updated.

I am not 100% happy with this solution, as ideally we should set the select value immediately and fetch the hierarchies afterwards (as the part that makes the selection slow is [this part](https://github.com/visualize-admin/visualization-tool/blob/main/app/configurator/config-form.tsx#L79-L95), which fetches the hierarchy before sending update event). I tried to make it work by first sending an `CHART_FIELD_CHANGED` event with `selectedValues: []`, then querying hierarchies with subsequent event sent after they are fetched, but there were also some problems with this approach and I still think it should be done better 😅 

Let me know if you have better ideas on how to handle this – maybe it needs a bigger refactor?

Edit: I've additionally disabled select element when hierarchies are being fetched.

### DIFF
--- a/app/configurator/components/field.tsx
+++ b/app/configurator/components/field.tsx
@@ -1,4 +1,6 @@
 import { t } from "@lingui/macro";
+import { CircularProgress, Theme } from "@mui/material";
+import { makeStyles } from "@mui/styles";
 import { extent, timeFormat, TimeLocaleObject, timeParse } from "d3";
 import get from "lodash/get";
 import { ChangeEvent, ReactNode, useCallback, useMemo, useState } from "react";
@@ -52,6 +54,15 @@ import { DimensionMetadataFragment, TimeUnit } from "@/graphql/query-hooks";
 import { IconName } from "@/icons";
 import { useLocale } from "@/locales/use-locale";
 import { getPalette } from "@/palettes";
+
+const useStyles = makeStyles<Theme>((theme) => ({
+  loadingIndicator: {
+    color: theme.palette.grey[700],
+    display: "inline-block",
+    marginTop: theme.spacing(1),
+    marginLeft: theme.spacing(2),
+  },
+}));
 
 export const ControlTabField = ({
   component,
@@ -625,7 +636,8 @@ export const ChartFieldField = ({
   optional?: boolean;
   disabled?: boolean;
 }) => {
-  const fieldProps = useChartFieldField({ field });
+  const classes = useStyles();
+  const { fetching, ...fieldProps } = useChartFieldField({ field });
 
   const noneLabel = t({
     id: "controls.dimension.none",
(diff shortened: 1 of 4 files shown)


app/configurator/components/field.tsx
```diff
@@ -53,6 +55,15 @@ import { IconName } from "@/icons";
 import { useLocale } from "@/locales/use-locale";
 import { getPalette } from "@/palettes";
 
+const useStyles = makeStyles<Theme>((theme) => ({
+  loadingIndicator: {
+    color: theme.palette.grey[700],
+    display: "inline-block",
+    marginTop: 3,
+    marginLeft: 8,
```
ptbrowne, line 63: Nit: Shouldn't those margins be defined as function of theme.spacing ?

app/configurator/components/field.tsx
```diff
@@ -625,7 +636,10 @@ export const ChartFieldField = ({
   optional?: boolean;
   disabled?: boolean;
 }) => {
-  const fieldProps = useChartFieldField({ field });
+  const classes = useStyles();
+  const { fetchingHierarchy, ...fieldProps } = useChartFieldField({
```
ptbrowne, line 640: I am thinking that fetchingHierarchy is maybe too narrow as a usecase to have it as a standalone property. I would propose to have only `fetching` here.

app/configurator/config-form.tsx
```diff
@@ -87,22 +102,29 @@ export const useChartFieldField = ({
...
+                selectedValues: leaves,
+              },
+            });
+          }
+        })
+        .finally(() => {
+          if (!unmountedRef.current) {
+            setFetchingHierarchy(false);
+          }
+        });
```
ptbrowne, line 127: Why don't we use a try/finally here ?

app/configurator/components/field.tsx
bprusinowski: Yes, thanks for noticing this 👍

app/configurator/components/field.tsx
bprusinowski: Done :)

app/configurator/config-form.tsx
bprusinowski: Not sure why I've added finally there, I have moved this logic up inside .then call as anyway it is not supposed to fail, even when user switching to another chart in the meantime :)

ptbrowne: Could we use an `await` then ?

bprusinowski: Yes, I've changed the code to use `await` :)

ptbrowne (approved):

ptbrowne:
Thanks !

PR #809, opened 2022-10-27
feat: Refinements to table column styles

### LINKED ISSUE #808: Enhancements to table view for ordinal scales
https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/430

* currently visualize allows `column_style in (bar chart, heat map, text)` [demo](https://test.visualize.admin.ch/en/v/d8_zoaoAjGFJ?dataSource=Int)
* in a next step the following should be supported: 
  * disallow bar chart as it makes no sense for nominal nor ordinal scales
  * allow heat map using schema:color as in case from [forest fire danger level](https://www.waldbrandgefahr.ch/en/current-hazard-situation)

bprusinowski:
Closes #808.

Previously we incorrectly enabled bar and heat map column style options for some dimension types; this PR improves that by enabling them only for numerical measures and allowing to use categories for other dimension types. It also moves a placeholder text "Select a dimension" to top, instead of being the last option.

Additionally, there is now a "dimension" color palette, which exists when there are pre-defined colors attached to dimension values; we are then able to select such palette from ColorPalette picker.

### How to test
1. Go to `/en/browse/dataset/https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Ftest_ordinal_507%2F7?dataSource=Int`
2. Select `ord1` dimension in Columns section.
3. Change column style to `Categories`.
4. See that the values are colored according to custom dimension colors.
5. See that it's possible to change the default color palette to different palettes.

### DIFF
--- a/app/charts/index.ts
+++ b/app/charts/index.ts
@@ -27,9 +27,9 @@ import {
   TableColumn,
   TableFields,
 } from "@/configurator/config-types";
-import { DEFAULT_PALETTE } from "@/configurator/configurator-state";
 import { FIELD_VALUE_NONE } from "@/configurator/constants";
 import { HierarchyValue } from "@/graphql/resolver-types";
+import { getDefaultCategoricalPaletteName } from "@/palettes";
 import { visitHierarchy } from "@/rdf/tree-utils";
 import { CHART_CONFIG_VERSION } from "@/utils/chart-config/versioning";
 
@@ -160,6 +160,8 @@ export const getInitialAreaLayer = ({
   // color
   measure: NumericalMeasure | OrdinalMeasure;
 }): MapAreaLayer => {
+  const palette = getDefaultCategoricalPaletteName(measure);
+
   return {
     componentIri: component.iri,
     color: isNumericalMeasure(measure)
@@ -173,9 +175,9 @@ export const getInitialAreaLayer = ({
       : {
           type: "categorical",
           componentIri: measure.iri,
-          palette: "oranges",
+          palette,
           colorMapping: mapValueIrisToColor({
-            palette: DEFAULT_PALETTE,
+            palette,
             dimensionValues: measure.values,
           }),
         },
@@ -271,6 +273,9 @@ export const getInitialConfig = ({
       const scatterplotSegmentComponent =
         getCategoricalDimensions(dimensions)[0] ||
         getGeoDimensions(dimensions)[0];
+      const scatterplotPalette = getDefaultCategoricalPaletteName(
+        scatterplotSegmentComponent
+      );
 
       return {
         version: CHART_CONFIG_VERSION,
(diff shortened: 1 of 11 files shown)


app/charts/table/table-state.tsx
```diff
@@ -281,7 +281,7 @@ const useTableState = ({
           return common as TextColumnMeta;
         } else if (columnStyleType === "category") {
           const { colorMapping } = columnStyle as ColumnStyleCategory;
-          const dimension = dimensions.find(
+          const dimension = [...dimensions, ...measures].find(
```
ptbrowne, line 284: Nit: Seems like we have a allColumnsByIri that can be used for this purpose.

app/charts/index.ts
```diff
@@ -173,9 +176,9 @@ export const getInitialAreaLayer = ({
       : {
           type: "categorical",
           componentIri: measure.iri,
-          palette: "oranges",
+          palette: hasDefaultColors ? "default" : DEFAULT_PALETTE,
           colorMapping: mapValueIrisToColor({
-            palette: DEFAULT_PALETTE,
+            palette: hasDefaultColors ? "default" : DEFAULT_PALETTE,
```
ptbrowne, line 181: Nit: I think "dimension" here would be clearer.

app/charts/shared/colors.tsx
```diff
@@ -18,3 +20,9 @@ export const rgbArrayToHex = (rgbArray: number[]): string => {
       );
   }
 };
+
+export const hasDimensionDefaultColors = (
+  d?: DimensionMetadataFragment
+): d is DimensionMetadataFragment => {
+  return d?.values?.[0]?.color !== undefined;
```
ptbrowne, line 27: If by any chance, the 1st color has no color, we have no palette. I think it would be good to be a bit more robust here and do a `some(x => x.color)`.

There is a small chance that the problem appears but it would be somewhat hard to debug I guess if it happens.

The problem would be for dimensions with a lot of values, where this check should be memoized.

app/charts/index.ts
```diff
@@ -1153,17 +1202,18 @@ const convertTableFieldsToSegmentField = ({
 
   if (component) {
     const { componentIri } = component;
+    const actualComponent = [...dimensions, ...measures].find(
+      (d) => d.iri === componentIri
+    ) as DimensionMetadataFragment;
+    const hasDefaultColors = hasDimensionDefaultColors(actualComponent);
+    const palette = hasDefaultColors ? "default" : DEFAULT_PALETTE;
```
ptbrowne, line 1209: Nit: Maybe a getDefaultPalette(component) function would be helpful there as the check + assignment is repeated for every chart.

ptbrowne (approved):
Cool change, some remarks but overall LGTM :)

app/charts/shared/colors.tsx
bprusinowski: Yes, that's a really good point! I will change this behavior as suggested 🙇

ptbrowne (approved):

Issue #806, opened 2022-10-25
sort order (identifier) reversed in edit mode


Rdataflow:
**Describe the bug**
sort order (identifier) counterintuitively reversed in edit mode (only for some datasets for some unkwown reason - maybe hierarchy?)

**To Reproduce**
1. Go to https://test.visualize.admin.ch/en/v/MiUdeVIMBOgq?dataSource=Int
2. Click on copy&edit
3. See reversed sort order (DESC) in legend below the chart

**Expected behavior**
Same sort order as on published chart

**Screenshots**
![image](https://user-images.githubusercontent.com/26090254/197818041-72d6b88d-0707-4927-bf5d-7bb8a8bd6911.png)

**Environment (please complete the following information):**
 - Visualize environment and version: TEST v3.11.0 ([0703f16](https://github.com/visualize-admin/visualization-tool/commit/0703f169f11d32259384c75317d0cb4e15d4bc34))

PR #810, opened 2022-10-27
Sort legend items in case of hierarchy

ptbrowne:
In case of hierarchy, we would not sort legend items and instead rely
on hierarchy order, which is not what we want.

Fixes : https://github.com/visualize-admin/visualization-tool/issues/806

### How to test

- Use the NFI dataset `/en/create/new?cube=https://environment.ld.admin.ch/foen/nfi/49-19-44/cube/1`
- Color by production region
- Verify that legend items are sorted automatically and can sorted in reverse, and by name

### DIFF
--- a/app/charts/shared/legend-color.tsx
+++ b/app/charts/shared/legend-color.tsx
@@ -11,6 +11,7 @@ import { useInteractiveFilters } from "@/charts/shared/use-interactive-filters";
 import Flex from "@/components/flex";
 import { Checkbox } from "@/components/form";
 import {
+  DataSource,
   isSegmentInConfig,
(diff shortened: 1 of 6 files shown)


bprusinowski:
LGTM 💯

bprusinowski (approved):

ptbrowne:
I changed the logic and I base myself on the labels received. This is because for some sorters we need the measures (byMeasure), and here we do not have it. Since the labels come already sorted, I figured we could remove the need for sorters here, and resort the hierarchy group entries according to the original order.

bprusinowski (approved):
Cool, looking good! 😎

PR #812, opened 2022-10-27
Default search order

ptbrowne:
Fixes https://github.com/visualize-admin/visualization-tool/issues/811

- feat: Default sort order is newest, when a search is made, it switches to relevance

### DIFF
--- a/app/graphql/resolvers/rdf.ts
+++ b/app/graphql/resolvers/rdf.ts
@@ -14,7 +14,6 @@ import {
   QueryResolvers,
   Resolvers,
 } from "@/graphql/resolver-types";
-import { ResolvedDataCube } from "@/graphql/shared-types";
 import { parseLocaleString } from "@/locales/locales";
 import { createSource } from "@/rdf/create-source";
 import {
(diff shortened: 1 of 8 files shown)


app/graphql/resolvers/rdf.ts
```diff
@@ -33,7 +32,37 @@ import {
...
+    case undefined:
+    case null:
+      results.sort((a, b) => {
+        const ra = getCube(a).datePublished || "0";
+        const rb = getCube(b).datePublished || "0";
+        return descending(ra, rb);
+      });
+      break;
+    case DataCubeResultOrder.Score:
+      break;
```
bprusinowski, line 59: Results are sorted be score by default, right?

bprusinowski (commented):
Overall LGTM, but the behaviour is inconsistent when removing the search query and clicking enter and when clicking "x" to clear the search query – in the first case it sorts by Newest, while in second it sorts by Title. I feel like they should behave in the same way, what do you think?

bprusinowski:
Also, not related to this PR, but do you know how it's possible that some cubes do not have a created date?

app/graphql/resolvers/rdf.ts
ptbrowne: Yes, when they enter this function, they are already sorted.

ptbrowne:
No, I don't know :/ @l00mi ?

ptbrowne:
>  I feel like they should behave in the same way, what do you think?

Agreed, I improved the behavior, and fixed the test ;)

bprusinowski (approved):
Thanks, works great now 🎉

l00mi:
> No, I don't know :/ @l00mi ?

It should not happen, but if cubes are manually created, it might always happen, that this gets forgot, or is "not yet" available. So it is up to you to either ignore these cubes, or handle the fact that they have no created date.

ptbrowne:
OK we handle this case. Do you have documentation on which SHACL rules are applied when ? It seems like the ICV guard mode from Stardog could be useful to apply validation, is this something that could be activated to improve data integrity ?

l00mi:
Lets discuss this in one of the next dev meetings.

Issue #802, opened 2022-10-24
Tech Exploration: Animated Charts / Time Slider with Play Button


ptbrowne:
In a first step we will  invest 2 days in exploration 

Input: First data set where this feature should be implemented (BAFU)
Output: Initial proof of concept (IXT), defining a scope for further development in a first iteration.

References: https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/78

bprusinowski:
Might be useful to explore: https://observablehq.com/@mbostock/scrubber

bprusinowski:
Still todo (related to this task):
- [ ] Update design of time slider
- [ ] Make chart container size fixed so it doesn't jump like now
- [ ] Add an ability to run an animation (requestAnimationFrame?)
- [ ] Enable time slider for more chart types
- [ ] Add transitions to charts (visx?)
- [ ] Apply interactive filters to table preview
- [ ] Enable interactive filters for maps (and extend CommonChartState with preparedData and allData)

PR #815, opened 2022-10-28
refactor: Set up the ground for interactive time slider

bprusinowski:
Contributes to #802.

I think it makes the most sense to implement the time slider / animated charts feature by adding additional interactive filter options (named e.g. time slider).

The user could enable the time slider and select given temporal dimension (that was not mapped to any field in configuration step); that action would remove the single filter for this dimension and trigger a new load of dataset, which now can be sliced by this temporal dimension.

This PR sets the ground for this implementation by cleaning stuff related to interactive filters & renaming current `time` filter into `timeRange` filter.


ptbrowne (approved):
💯

PR #816, opened 2022-10-28
feat: Add time slider to interactive filters

bprusinowski:
Contributes to #802.

This PR introduces a basic time slider which can be used to interactively filter a chart based on time dimension.

### How to test
1. Go to `/en/browse/dataset/https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fubd0104prod%2F5?dataSource=Int&flag__timeslider=true` (notice the `flag__timeslider`, feature is not available by default).
2. Create a column chart.
3. Change X variable to `Parameter`.
4. Go to Publishing step and enable Time Slider.
5. See that a slider appeared below the chart.
6. Use the slider to filter the data.
7. Change X dimension back to Date and see that a time slider disappeared.


app/configurator/configurator-state.spec.tsx
```diff
@@ -646,7 +646,7 @@ describe("retainChartConfigWhenSwitchingChartType", () => {
   };
 
   it("should retain appropriate x & y fields and discard the others", () => {
-    runChecks(covid19ColumnChartConfig as ColumnConfig, xyChecks);
+    runChecks(covid19ColumnChartConfig as unknown as ColumnConfig, xyChecks);
```
ptbrowne, line 649: I think this should be fixed before merge ?

app/configurator/interactive-filters/interactive-filters-config-state.tsx
```diff
@@ -16,26 +16,21 @@ export const useInteractiveFiltersToggle = ({
...
 }) => {
   const [state, dispatch] = useConfiguratorState(isDescribing);
-  const onChange = useCallback<(e: ChangeEvent<HTMLInputElement>) => void>(
-    (e) => {
-      const newIFConfig = produce(
-        state.chartConfig.interactiveFiltersConfig,
-        (draft) => {
-          if (draft?.[path]) {
-            draft[path].active = e.currentTarget.checked;
-          }
```
ptbrowne, line 26: I do not get why is this part (the .active = .. checked) removed ?

app/configurator/interactive-filters/interactive-filters-config-state.tsx
```diff
@@ -16,26 +16,21 @@ export const useInteractiveFiltersToggle = ({
...
-        value: newIFConfig,
-      });
-    },
-    [dispatch, path, state]
-  );
+    dispatch({
+      type: "INTERACTIVE_FILTER_CHANGED",
+      value: newConfig,
+    });
+  }, [dispatch, state]);
```
ptbrowne, line 36: Not part of this PR, but useEvent could be used.

app/configurator/components/chart-annotations-selector.tsx
```diff
@@ -20,54 +21,65 @@ export const ChartAnnotationsSelector = ({
...
+  const isInteractiveFilterField = useMemo(() => {
+    switch (activeField as InteractiveFilterType) {
+      case "dataFilters":
+      case "legend":
+      case "timeRange":
+      case "timeSlider":
+        return true;
+
+      default:
+        return false;
```
ptbrowne, line 42: Why could we have the default case ?

app/configurator/interactive-filters/interactive-filters-config-options.tsx
```diff
@@ -216,6 +243,90 @@ const InteractiveTimeRangeFilterOptions = ({
...
+
+    const xComponentIri = getFieldComponentIri(chartConfig.fields, "x");
+    const xComponent = allComponents.find((d) => d.iri === xComponentIri);
+
+    const timeSliderDimensions = getTimeSliderFilterDimensions({
+      allComponents,
+      excludedComponent: isTemporalDimension(xComponent)
+        ? xComponent
+        : undefined,
+    });
```
ptbrowne, line 277: Nit: I think the getTimeSliderFilterDimensions helper should directly take the dataCubeByIri and the chart config. I think this way we do not have to "invent" a new API in the form of "allComponents" and "excludedComponent". This way it would also be easier to have this helper run directly on the redux + urql client cache if needed.

app/charts/area/areas-state.tsx
```diff
@@ -29,7 +29,7 @@ import {
   stackOffsetDivergingPositiveZeros,
   useOptionalNumericVariable,
   usePlottableData,
-  usePreparedData,
+  useDataAfterInteractiveFilters,
```
ptbrowne, line 32: 💯

app/charts/shared/chart-state.ts
```diff
@@ -0,0 +1,8 @@
+import { ChartType } from "@/configurator";
+
+import { Bounds } from "./use-width";
+
+export interface CommonChartState {
+  chartType: ChartType;
+  bounds: Bounds;
+}
```
ptbrowne, line 8: Could allData and preparedData be here ?

app/charts/area/areas-state.tsx
```diff
@@ -160,9 +160,13 @@ const useAreasState = (
     [dataGroupedByX, xKey, getY, getSegment]
   );
 
+  const plotters = useMemo(() => {
+    return [getX, getY];
+  }, [getX, getY]);
```
ptbrowne, line 165: It's a bit inconvenient to memo here, maybe usePlottableData could have the plotters directly as the dependency ?

in chart-helpers.tsx

```patch
  const isPlottable = useCallback(
    (d: Observation) => {
      for (let p of plotters) {
        const v = p(d);
        if (v === undefined || v === null) {
          return false;
        }
      }
      return true;
    },
-    [plotters]
+    plotters
  );
```

app/configurator/interactive-filters/time-slider.tsx
```diff
@@ -0,0 +1,61 @@
...
+    if (componentIri && chartState.chartType !== "map") {
+      const rawRange = [
+        ...new Set(
+          chartState.allData.map((d) => d[componentIri]).filter(Boolean)
+        ),
+      ].sort() as string[];
+      const range = rawRange.map(parseDate);
+      const [min, max] = [range[0], range[range.length - 1]];
+      const scale = scaleLinear().range([min.getTime(), max.getTime()]);
+      const reversedRange = range.reverse();
```
ptbrowne, line 27: The computation of the available values for the range can be done once outside of the onChange. I think d3 extent could be used here to get the min/max without sorting.

app/configurator/interactive-filters/time-slider.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+      return (e: ChangeEvent<HTMLInputElement>) => {
+        const t = +e.target.value;
+        setT(t);
+
+        const tDate = new Date(scale(t));
+        const IFDate = reversedRange.find((d) => d <= tDate) as Date;
```
ptbrowne, line 34: Since we have a sorted list here, I think we could have d3 bisect here to search for the nearest date. I think it is important to be fast here as we could have a lot of dates and the user could slide quickly.

ptbrowne (commented):
I was curious and did a pre-review :) Looks great.

ptbrowne:
I think we could have 

- Enable time slider for more chart types
- Add transitions to charts (visx?)

in a next PR, and we would hide the interactive filter behind a flag. The PR is already getting big ;) What do you think >

app/configurator/configurator-state.spec.tsx
bprusinowski: Yes, will remember about that!

app/configurator/interactive-filters/interactive-filters-config-state.tsx
bprusinowski: It was removed because timeSlider do not use a "Show time slider" checkbox like all other interactive filters. I thought it was inconvenient to need to first toggle a checkbox and then select a dimension afterwards – but anyway we don't have design yet so this might change 😅

bprusinowski: Hmm, or maybe not, I will check :)

app/configurator/components/chart-annotations-selector.tsx
bprusinowski: I'll also need to probably improve this, the case here is that activeField could be a lot of things – areaLayer, segment, x, y, but also interactive filter and here I wanted to catch these cases in more clear way, for all other strings the default case if triggered. Do you have an idea how to handle this better?

app/configurator/interactive-filters/interactive-filters-config-options.tsx
bprusinowski: Good point! I'll change that :)

app/charts/shared/chart-state.ts
bprusinowski: Yes, that was my goal 😅 But as always, maps are different here and there we have data for areas and data for symbols... That's why allData and preparedData are not here yet, I need to think about how to handle this.

app/charts/area/areas-state.tsx
bprusinowski: Wow, that's a great hint, I totally missed that 😅 I'll adjust :)

app/configurator/interactive-filters/time-slider.tsx
bprusinowski: Yup, also for the time slider it would be good to only trigger dispatch when new date is different than current date, as right now we update unnecessarily – I'll modify this part, thanks for suggestions 🙇

bprusinowski:
Thanks @ptbrowne for review! Yes, it's getting big, I'll try to clean up the stuff and I even think it would make sense to leave things like chart container size, animation, design of the slider also for another PR and just fix things here + hide the feature behind a feature flag. I think it would make sense to merge before new UI is implemented so we can avoid lots of conflicts 😨

ptbrowne:
Yep, makes sense 💯

app/configurator/interactive-filters/interactive-filters-configurator.tsx
```diff
@@ -22,6 +22,8 @@ import { useLocale } from "@/locales/use-locale";
 
 import { getTimeSliderFilterDimensions } from "./helpers";
 
+const ENABLE_TIME_SLIDER = false;
```
ptbrowne, line 25: You can use `app/configurator/components/flag.tsx` to have flags that can be dynamically set via the URL :)

app/charts/shared/use-sync-interactive-filters.tsx
```diff
@@ -44,14 +49,35 @@ const useSyncInteractiveFilters = (chartConfig: ChartConfig) => {
...
-  }, [IFstate.timeSlider.value, timeSliderFilter?.componentIri, dispatch]);
+
+    if (
+      xComponentIri !== undefined &&
+      xComponentIri === timeSliderFilter?.componentIri
+    ) {
+      dispatchConfigurator({
+        type: "INTERACTIVE_FILTER_CHANGED",
+        value: {
+          ...interactiveFiltersConfig,
```
ptbrowne, line 68: Nit: Maybe there could be a "EMPTY_TIME_SLIDER_FILTER" action type, to be coherent with what's above ? I do not like that we have to pass the interactiveFiltersConfig here, I would prefer that this logic is in the reducer.

ptbrowne (approved):
LGTM 🙇

app/configurator/interactive-filters/interactive-filters-configurator.tsx
bprusinowski: That's a really neat way, thanks 💯

PR #819, opened 2022-11-02
Use a cube loader

ptbrowne:
While investigating queries, I noticed that we were loading the cubes a lot,
since the dataCubeByIri resolver is high and needs to load the cube shape.
Since the cube can be shared across resolvers, I used a loader.
Since the cube does not change a lot, I added also an application level
cache through the cachedWithTTL helper. This means that even across requests,
within a 60s timespan, the cube will be resolved through a cache.

## How to test

Every interaction with cubes in the cube configurator should be faster.

### DIFF
--- a/app/configurator/components/chart-configurator.tsx
+++ b/app/configurator/components/chart-configurator.tsx
@@ -7,6 +7,7 @@ import {
   Menu,
   MenuItem,
   Theme,
+  Typography,
 } from "@mui/material";
 import { makeStyles } from "@mui/styles";
 import { ascending } from "d3";
@@ -229,6 +230,7 @@ const useEnsurePossibleFilters = ({
         .toPromise();
       if (error || !data) {
         setError(error);
+        setFetching(false);
         console.warn("Could not fetch possible filters", error);
         return;
       }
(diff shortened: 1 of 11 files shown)


app/rdf/query-literals.ts
```diff
@@ -20,10 +21,11 @@ const buildResourceLiteralQuery = ({
   values: NamedNode<string>[];
   predicates: PredicateOption;
 }) => {
+  const uniqueValues = uniqBy(values, (x) => x.value);
   const q = SELECT.DISTINCT`?iri ${Object.keys(predicates).map((x) => `?${x}`)}`
     .WHERE`
       values ?iri {
-        ${values}
+        ${uniqueValues}
```
bprusinowski, line 28: 💯

app/configurator/components/chart-configurator.tsx
```diff
@@ -577,6 +580,11 @@ export const ChartConfigurator = ({
             aria-labelledby="controls-data"
             data-testid="configurator-filters"
           >
+            {possibleFiltersError ? (
+              <Typography variant="body2" color="error">
+                {possibleFiltersError.message}
+              </Typography>
+            ) : null}
```
bprusinowski, line 587: 👍

bprusinowski (approved):
Cool, really nice change 👏

Issue #820, opened 2022-11-03
Ordinal dimensions, Bar Chart: Values are not classified according to their schema:position


ortnever:
Values are not classified according to their schema:position in Bar chart although it was the case before (see #580)

open and edit : https://test.visualize.admin.ch/en/v/4E2vn95fel3H

or 

Open int.visualize.admin.ch, source : int
Search "Red list" and inculde draft Datasets
Select "Red list (latest survey with%)"
Start a visualization
Select Dimension "Status" as color

The order of stati is correct in the filters (on the right side and on the left side) but not in the chart and Legend

Expected behavior :
Values (in this case Stati) have to be classified according to their schema:position in all Charts. Please, could you do the correction for all chart's types at  same time (see #580)

Screenshots

![image](https://user-images.githubusercontent.com/79834090/199688257-6d957489-c524-479e-ace5-feb526db7812.png)

![image](https://user-images.githubusercontent.com/79834090/199686601-20c426b4-fbd2-4083-8288-72040576647d.png)

Environment (please complete the following information):

Visualize environment and version: test.visualize.admin.ch v3.11.0 ([a94a977]. Same Problem on INT v3.11.0 ([76d8044]
Browser : Edge

ptbrowne:
@ortnever I was surprised because when developing this sorting, I tested this dataset and it worked correctly. I think it's because the order of the sorting was "identifier" then "position" instead of "position" then "identifier". It should be fixed in https://visualization-tool-git-fix-order-position-identifier-ixt1.vercel.app/en?dataSource=Int. Can you confirm ?

ortnever:
@ptbrowne Thanks Patrick. It's correct now and also for Pie Chart. That's great.

Issue #580, opened 2022-06-09
Ordinal dimensions, Pie Chart: Stati are not classified according to their schema:position


ortnever:
Stati are not classified according to their schema:position in Pie Chart. In Bar chart is ok.

1. Open test.visualize.admin.ch
2. Search "Red list" and inculde draft Datasets
3. Select "Red list (latest survey with%)"
4. Start a visualization
5. Select Dimension "Status" as color
6. Unselect a category to load the data (because of the bug #297)
7. Select Pie Chart
8. The order of stati is correct in the legend (on the right side) but not in the chart

**Expected behavior**
 Stati have to be classified according to their schema:position in all Charts inclusiv Pie Chart

**Screenshots**
![image](https://user-images.githubusercontent.com/79834090/172802036-1cb2e262-b113-42d1-8862-644d0e66ca94.png)

**Environment (please complete the following information):**
 - Visualize environment and version: v3.7.4 ([9fa6b1f]
 - Browser : Edge

PR #821, opened 2022-11-03
fix: Order for sorting is position / identifier

ptbrowne:
When ordering segment, "position" has priority over "identifier".

fix https://github.com/visualize-admin/visualization-tool/issues/580
fix https://github.com/visualize-admin/visualization-tool/issues/820


bprusinowski (approved):
LGTM! Cool catch related to the "asc" & "desc" array filling ;)

PR #823, opened 2022-11-04
fix: Use dimension colors for segments

bprusinowski:
Recently we enabled fetching of default colors attached to dimensions, but it turned out they weren't enabled for segment fields – this PR fixes that. It also corrects display of reset / refresh based on whether a palette is `dimension` or not (and not as previously, where we just checked if dimension has custom colors).


ptbrowne (approved):
👍

PR #829, opened 2022-11-08
UI configurator enhancements

ptbrowne:
- Remove right panel, and move its content to a drawer that slides in
- Allow to change title / subtitle on click
- Add warning icon when title / description have not been added

I will implement changes to the interactive filters in a subsequent PR.


app/configurator/components/empty-right-panel.tsx
```diff
@@ -2,15 +2,12 @@ import { Trans } from "@lingui/macro";
 import { Typography } from "@mui/material";
```
bprusinowski, line 2: Here the whole `empty-right-panel.tsx` could be removed 😀

bprusinowski (approved):
Wow, works like a charm, I really like the changes 💯

There is only one thing with collapse animation in Safari, I'm attaching a video below (works fine on Chrome).

https://user-images.githubusercontent.com/52032047/200574028-db246bf8-3ba4-4603-a11a-835d222d001a.mov

app/configurator/components/empty-right-panel.tsx
ptbrowne: Done 👍

AnninaWalker:
Great Work! I'll have to look into it in more detail but so far it looks good to me and I'm ok with all the changes that have been made in regards to design as mentioned in the description. Well done! 👍 

@bprusinowski I Had some feedback regarding the hover on the Data search page. Comment is here: #817

AnninaWalker:
I was wondering if per default only the chart type and the chart options should be expanded and the other sections collapsed

ptbrowne:
I will optimize the Safari animation in a subsequent PR.

PR #831, opened 2022-11-09
feat: Show tree in MultiFilters

### LINKED ISSUE #800: Hierarchies UI Refinements
Implement changes from https://github.com/visualize-admin/visualization-tool-private/issues/18

bprusinowski:
Closes #800.

[Design specs](https://www.notion.so/interactivethings/V04-Indent-and-new-UI-844455727a2b421ca00e0309c3a25957)

- Filter values in drawer are displayed as tree
- Search in the tree keeps tree structure (e.g. when a child node is "found", all of its parents will also be displayed)

### How to test

Look for a dataset with hierarchies (e.g. Bathing water quality (with Hierarchie) or Öffentliche Ausgaben TEST 3) and without hierarchies (e.g. Bathing water quality) and create a chart with color segmentation; afterwards edit filters to see if it works correctly.


app/configurator/components/filters.tsx
```diff
@@ -568,16 +570,49 @@ const TreeAccordion = ({
...
+      return children
+        .map((d) =>
+          areChildrenSelected({ children: d.children, selectedValues })
+        )
+        .some((d) => d === true);
+    }
+  } else {
+    return false;
+  }
+};
```
ptbrowne, line 620: I think we could compute a tree of state, in one go we can compute all this tree, here we are doing multiple the same work (to compute the state of the parent, we must go over their children/grand-children, and then we process again the grand children to compute the state.

The easy way to have the same behavior would be to memoize.

```
const areChildrenSelectedMemoized = useMemo(() => {
	const areChildrenSelected = memoize((d) => {
	        const { children } = d
		if (Array.isArray(children) && children.length > 0) {
		  const values = children.map((d) => d.value);
		  if (selectedValues.some((d) => values.includes(d.value))) {
		    return true;
		  } else {
		    return children
		      .map((d) =>
		        areChildrenSelected(d)
		      )
		      .some((d) => d === true);
		  }
		} else {
		  return false;
		}
	})
       return do
}, [selectedValues])
```

bprusinowski: Yes, that's a good point, I also wasn't happy with re-computations – I'll introduce this change 🙇

ptbrowne: I guess here you'd need to have a second parameter to the memoize, `(d) => d.value`.

app/configurator/components/filters.tsx
```diff
@@ -689,6 +698,35 @@ const Tree = ({
...
+      if (children.length > 0) {
+        result.push({ ...d, children });
+      }
+    }
+
+    return result;
+  };
+
+  return tree.reduce(filter, []);
+};
```
ptbrowne, line 744: 1. Nit: Seems like it is similar to pruneTree, maybe it could be reused ?
2. Nit: I think it'd be clearer to use Array.filter instead of Array.reduce ?

app/configurator/components/filters.tsx
```diff
@@ -701,11 +739,16 @@ const DrawerContent = forwardRef<
...
   const { onClose, values, options, pendingValuesRef, hasColorMapping } = props;
   const classes = useStyles();
+  const [textInput, setTextInput] = useState("");
   const [pendingValues, setPendingValues] = useState<HierarchyValue[]>(
     () => values
   );
   pendingValuesRef.current = pendingValues;
 
+  const filteredOptions = useMemo(() => {
+    return filterTree({ text: textInput, tree: options });
```
ptbrowne, line 773: `return pruneTree(options, x => x.label.toLowerCase().includes(lowercaseText))` ?

app/configurator/components/filters.tsx
```diff
@@ -782,6 +815,12 @@ const DrawerContent = forwardRef<
           }
           sx={{ typography: "body2" }}
         />
+        <FilterControls
+          selectAll={() => setPendingValues(flatOptions)}
+          selectNone={() => setPendingValues([])}
+          allKeysLength={flatOptions.length}
```
ptbrowne, line 789: I guess there should be a uniqueBy value here otherwise we can count twice values that are multiple times in the tree.

ptbrowne (commented):
LGTM 👍 Thanks, I have some remarks but I think they could also be tackled in a further PR.

app/rdf/query-hierarchies.ts
```diff
@@ -95,17 +135,45 @@ export const queryHierarchy = async (
...
+      depth: -1,
+      children: [],
+      dimensionIri,
+      hasValue: true,
+    }));
+
+  return sortTree([...prunedTree, ...additionalTreeValues], (a, b) =>
+    a.depth === -1 && b.depth === -1
+      ? defaultSort(a, b)
+      : a.depth === -1 || b.depth === -1
```
ptbrowne, line 174: I do not think it works correctly since you check both a and b. Here I think only `a` should be checked.

```
> // INCORRECT
> [{d:-1, a:'hey'}, {d:1}, { d: -1 }].sort((a, b) => a.d === -1 || b.d === -1 ? 1 : 0)
Array(3) [ {…}, {…}, {…} ]
0: Object { d: -1 }
1: Object { d: 1 }
2: Object { d: -1, a: "hey" }
> // CORRECT
> [{d:-1, a:'hey'}, {d:1}, { d: -1 }].sort((a, b) => a.d === -1 ? 1 : 0)
> Array(3) [ {…}, {…}, {…} ]
0: Object { d: 1 }
1: Object { d: -1 }
2: Object { d: -1, a: "hey" }
```

bprusinowski: Yes, I don't know how I created this 😅 I switched to using `d3.ascending` & `d3.descending` here.

bprusinowski:
Thanks for the review @ptbrowne. For the first point, we include the color checkbox for all values, it's then visible or hidden based on `hasValue` property. I corrected this by checking if all values for a given depth level do not have a value; if so, the color box is not rendered. Otherwise, if there is at least one option with a value for a given depth level, it will be shown.

The second point should also be ok after the changes.

ptbrowne (approved):
Not sure why the hierarchy values were not correctly handled before the final commit ?

LGTM 👏

app/rdf/queries.ts
```diff
@@ -459,6 +459,8 @@ export const getCubeObservations = async ({
     disableDistinct: !!(!filters || Object.keys(filters).length === 0),
   });
 
+  // console.log({ query, observationsRaw });
+
```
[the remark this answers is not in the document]
bprusinowski: yes 👍

bprusinowski:
@ptbrowne I guess it's related to versioning / unversioning  of observations as for some datasets values didn't match and this resulted in many duplicated values e.g. in some cases there were "https://abc.org" in "proper" dimension values, and when using the previous query they had a format of "https://abc-test.org". I think it was the case for Ausgaben dataset.

Issue #822, opened 2022-11-04
Hierarchies list question/bug - edition mode - left filter


FabianCretton:
**Describe the bug**
I am working on a new hierarchy and see odd things in the display.

As this hierarchy is generated by a script, I am trying here to understand if:
- my generated hierarchy has problems (RDF)
- I do not fully understand how Visualize is using the hierarchy
- There is a display bug in Visualize

Cube: "fab Foreign Trade test 2" 
Environment TEST
In visualize, [here](https://test.visualize.admin.ch/en/browse/dataset/https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Ffab_foreign_trade_test_2%2F3?previous=%7B%22includeDrafts%22%3Atrue%2C%22order%22%3A%22SCORE%22%2C%22search%22%3A%22fab%22%7D&dataSource=Test)

Here is an example: the node "WK '1' ROUND WOOD (WOOD IN THE ROUGH)" has a child "List of Zolltarifs", which has 43 children.
But the filter on the left "Tariff or Basket ID" displays only 4 children:
![image](https://user-images.githubusercontent.com/6167669/199920180-8887eea9-90dc-4d4b-87db-513f09c479a9.png)

The filter on the right, when choosing a color, does display all children:
![image](https://user-images.githubusercontent.com/6167669/199920232-e32007fc-cc4e-4a6a-b59b-ab3616ff2679.png)

Those 43 children are visible when [displaying the hierarchy in the Cube Creator](https://test.cube-creator.lindas.admin.ch/app/hierarchies/dimension!!hierarchy!!hierarchy-swiss-foreign-trade-baskets+zolltarif-(on-bafu-helpers)). I suppose that the RDF is thus correct.

I see "missing" nodes in different parts of the hierarchy, I could give more examples if needed.

FabianCretton:
@ptbrowne salut Patrick, aurais-tu quelques infos à ce sujet ? Est-ce une incompréhension de ma part, une erreur dans les données, un bug de Visualize ? Merci d'avance

ptbrowne:
Hi @FabianCretton , I'll have a look tomorrow, it seems we are effectively not returning all the nodes but I do not know exactly why and right now I get 503 from test.lindas.admin.ch.

<details>
  <summary>Technical details</summary>

  ```
query DimensionHierarchy($sourceType: String!, $sourceUrl: String!, $locale: String!, $cubeIri: String!, $dimensionIri: String!) {
[...]

PR #832, opened 2022-11-09
Fix unique values hierarchy

ptbrowne:
- fix: Do not uniqueBy hierarchy values
- feat: Add debug method for graphviz

Fix https://github.com/visualize-admin/visualization-tool/issues/822#issuecomment-1308752418

### DIFF
--- a/app/configurator/components/use-hierarchy-parents.tsx
+++ b/app/configurator/components/use-hierarchy-parents.tsx
@@ -61,4 +61,30 @@ const useHierarchyParents = (
   }, [dimension.values, hierarchy]);
 };
 
+/**
+ * Can be used for debugging, pass a hierarchy, and copy the output
+ * to graphviz.
+ *
+ * @see https://dreampuf.github.io/GraphvizOnline/
+ */
+export const hierarchyToGraphviz = (
+  hierarchy: DimensionHierarchyQueryHierarchy
+) => {
+  const lines = [] as string[];
+  dfs(hierarchy, (node, { depth, parents }) => {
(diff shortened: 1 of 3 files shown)


app/configurator/components/use-hierarchy-parents.tsx
```diff
@@ -61,4 +61,30 @@ const useHierarchyParents = (
...
+  });
+  return `
+    digraph G {
+      rankdir=LR
+
+      ${lines.join("\n")}
+    }
+  `;
+};
+
```
bprusinowski, line 89: 👌

bprusinowski (commented):
When mapping `Tariff or Basket ID` as color field, selecting the `Zolltarifs` and `List of Zolltarifs` values doesn't do anything (no data is filtered + no color appears in the option row). I guess it's related to multiple values shared across parent, but is this an expected behavior?

app/configurator/components/use-hierarchy-parents.tsx
FabianCretton: I can give an input here, and this might be an "special" situation in my data: most of the nodes are related to values in the Data, except the "list of Zolltarif" node itself -> this is kind of a "fake" node, just to regroup and display the list of Zolltarifs that are included in a basket (whithout having to look at all tarifs in the children, and also given that a tarif could be directly at that level and not coming from a child).
As discussed with Patrick, there could be a better way to show this, we can talk about it. This was just a "possible" way that I used for that POC.
Thus, for me it makes sense if the node "list of Zolltarfis" can not be selected. It is a nice handling of the situation by Visualize.

bprusinowski (approved):
👌

PR #839, opened 2022-11-10
feat: Open search result only on title

ptbrowne:
Open search results only on title

Hover effect on all the search result is misleading since the chips for themes / organisations can
be clicked. Here only the title is clickable, and it gets an underline on hover.


bprusinowski (approved):
LGTM, but the cursor shouldn't be a pointer if clicking on a whole tile is not doing anything, I think it should only be visible when hovering over the title 👁️

PR #860, opened 2022-11-17
fix: Lines tooltip value colors

### LINKED ISSUE #859: Different colour in the mouse over than in the graphic and the legend
If I select a dimension as a colour and then I eliminate all categories except one (via the filter), blue will be used by default in the mouse over, while the category is a different colour.

This is not very important as it is normally rare to go through the colour axis if you want to display only one category. At most, this would be the case if you want to display a colour other than blue for this category.

**Expected behavior**
Mouse over should always use the same color as the selected color for the catagorie.

**Screenshots or video**
image

- Visualize environment and version: INT v3.12.0
- Browser and version : Edge

**Additional context**

bprusinowski:
Fixes #859.

Broken E2E test will be fixed by #855.

### DIFF
--- a/app/charts/line/lines-state.tsx
+++ b/app/charts/line/lines-state.tsx
@@ -41,7 +41,6 @@ import {
   useTimeFormatUnit,
 } from "@/formatters";
 import { getPalette } from "@/palettes";
-import { useTheme } from "@/themes";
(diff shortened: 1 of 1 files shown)


ptbrowne (approved):
👍

PR #862, opened 2022-11-17
fix: Top level hierarchy values in color legend

### LINKED ISSUE #858: Order in Legend by dataset "Greenhouse Gas Emission"
The order in the graph and the lengend are reversed in dataset "Greenhouse gas emission". In other datasets (with categories Production Regions or cantons) the order is ok.

**To Reproduce**
- Copy and edit this visualization : https://int.visualize.admin.ch/en/v/z3272nClUz2J
- The categories are ordered by ID in the graph (1. Sectors of emission 2. International aviation 3. Balance of land use) but the order in the legend is reversed

**Expected behavior**
graph and legend have to be ordered by schema:identifier by default (Auto)
See #199 https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/119

**Screenshots or video**
image

- Visualize environment and version: [e.g. INT v3.12.0]
- Browser : Edge

**Additional context**

bprusinowski:
Fixes #858.

The `useLegendGroups` function was appending empty arrays for top-level values (without parents) – this PR fixes that by creating an empty array outside of the `dfs` function to use the same object as parent for such values.

### DIFF
--- a/app/charts/shared/legend-color-helpers.ts
+++ b/app/charts/shared/legend-color-helpers.ts
@@ -1,0 +1,46 @@
+import { ascending } from "d3";
+
+import { HierarchyValue } from "@/graphql/resolver-types";
+import { dfs } from "@/utils/dfs";
+
+export const getLegendGroups = ({
+  title,
+  labels,
+  hierarchy,
+  sort,
+}: {
+  title?: string;
+  labels: string[];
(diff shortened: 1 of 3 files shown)


ptbrowne (approved):
💯 Thanks !

PR #863, opened 2022-11-17
Use trees in selects when hierarchy is present

ptbrowne:
## How to test

- Use bathing water with hierarchies (check draft when searching)
- In the data filters, the "observation location" select should display the hierarchy as a tree
- When activating interactive filters, the interactive filter also displays the hierarchy as a tree

## TODO

- [x] When you first open the select, it loads the hierarchy and then closes the select, and you have to open it again

### DIFF
--- a/app/charts/shared/chart-data-filters.tsx
+++ b/app/charts/shared/chart-data-filters.tsx
@@ -7,6 +7,7 @@ import { ChartFiltersList } from "@/components/chart-filters-list";
 import Flex from "@/components/flex";
 import { Select } from "@/components/form";
 import { Loading } from "@/components/hint";
+import SelectTree from "@/components/select-tree";
 import {
   ChartConfig,
   DataSource,
@@ -14,24 +15,24 @@ import {
   OptionGroup,
   Option,
 } from "@/configurator";
-import { TimeInput } from "@/configurator/components/field";
+import { hierarchyToOptions, TimeInput } from "@/configurator/components/field";
 import {
   getTimeIntervalFormattedSelectOptions,
   getTimeIntervalWithProps,
 } from "@/configurator/components/ui-helpers";
-import useHierarchyParents from "@/configurator/components/use-hierarchy-parents";
 import { FIELD_VALUE_NONE } from "@/configurator/constants";
 import { isTemporalDimension } from "@/domain/data";
 import { useTimeFormatLocale } from "@/formatters";
 import {
   Dimension,
+  HierarchyValue,
   TemporalDimension,
   TimeUnit,
+  useDimensionHierarchyQuery,
   useDimensionValuesQuery,
 } from "@/graphql/query-hooks";
 import { Icon } from "@/icons";
 import { useLocale } from "@/locales/use-locale";
-import { makeOptionGroups } from "@/utils/hierarchy";
 
 export const ChartDataFilters = ({
   dataSet,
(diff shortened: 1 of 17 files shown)


app/components/select-tree.tsx
```diff
@@ -0,0 +1,351 @@
...
+    setTimeout(() => {
+      menuRef.current?.updatePosition();
+    }, 500);
+    setTimeout(() => {
+      menuRef.current?.updatePosition();
+    }, 1000);
+    setTimeout(() => {
+      menuRef.current?.updatePosition();
+    }, 2000);
+  };
```
bprusinowski, line 278: Why such combo? Wouldn't just first timeout do the job? Or is the transition duration dependent on the number of children in a node?

app/components/select-tree.tsx
```diff
@@ -0,0 +1,351 @@
...
+        TransitionProps={transitionProps}
+      >
+        <TreeView
+          defaultSelected={value}
+          defaultExpanded={defaultExpanded}
+          onNodeSelect={handleNodeSelect}
+          onNodeToggle={handleNodeToggle}
+          defaultCollapseIcon={<ExpandMoreIcon />}
+          defaultExpandIcon={<ChevronRightIcon />}
+          // onNodeSelect={handleNodeSelect}
```
bprusinowski, line 341: 👁️

app/docs/form.docs.tsx
```diff
@@ -28,6 +29,49 @@ const SwitchExample = ({ initialChecked }: { initialChecked?: boolean }) => {
...
+          label: "Numbers",
+          value: "numbers",
+          selectable: false,
+          children: numberOptions,
+        },
+      ]}
+    />
+  );
+};
+
```
bprusinowski, line 74: 👌

app/charts/shared/chart-data-filters.tsx
```diff
@@ -255,14 +269,63 @@ const DataFilterBaseDimension = ({
...
+    }
+    if (!isKeyDimension) {
+      opts.push({
+        value: FIELD_VALUE_NONE,
+        label: noneLabel,
+        isNoneValue: true,
+      });
+    }
+    return opts;
+  }, [hierarchy, isKeyDimension, dimensionValues, noneLabel]);
```
bprusinowski, line 316: Did not check this as I'm not aware of optional hierarchy in any of the datasets, but wouldn't this append the "None" option at the end of the array (I think it should be at top)?

app/components/select-tree.tsx
ptbrowne: Refactored away by using the onEntered / onExited transition props of the tree item transitions.

app/charts/shared/chart-data-filters.tsx
ptbrowne: Ah yes ! I replaced push by unshift :)

ptbrowne:
1. Yes I need to see how to still lazy load the hierarchies (we had gained a significant boost by postponing the downloading of hierarchies), but it should be better done, I need to do refactors here to correctly provide the loading to the select.
2. Fixed, the method may seem hacky (via the --depth CSS variable), but I tried first to use calc(--depth) + 1, but it does not seem to work, it seems we can't have recursion in CSS variables :/
3. I tried to add an indication, I think it's better now, what do you think ? (cc @AnninaWalker )
4. Since I changed the way of doing the padding for the hover effect, this has changed I think. I personally think this looks good already ? 

image

bprusinowski (approved):
Yes, much better now 💯

PR #866, opened 2022-11-18
feat: Sort filter values & hierarchy tree

### LINKED ISSUE #857: Categories in the filter are not ordered by the IDs
The categories in the filter are not ordered by the IDs

**To Reproduce**

1. Go to 'int.visualize.admin.ch' and search "Rétributions uniques pour installations photovoltaïques"
2. Open the filter "Standort der Anlage"
3. The list of cantons in the left-hand filter is displayed in alphabetical order. 
4. Select "Show Datafilter" for "Standort der Anlage", Publish the graph and show filters
5. The list of cantons in interactiv filter is displayed in alphabetical order. 
6. Copy and edit the visualization
7. Choose "Standort der Anlage" as color.
8. The list of cantons in the left-hand filter is also displayed by alphabetical order.

**Expected behavior**
Categories in filters should be sorted by default by ID (Gitlab Issue #405).  
- It must work in all filters
- It must work for all dimensions, including the shared dimension with hierarchy (at least for level 1). Try with : https://int.visualize.admin.ch/en/v/OM6rpkJ7K9bT
The order of "Emission sources" according to the id is : 1. Sectors Ordinance CO2 / 2. International aviation (not included in total) / 3. Balance of land use
- it must work in edit mode and in publish mode with interactive filters. 

**Screenshots**
image

image

image

 - Visualize environment and version: v3.12.0 ([c7820ce]
- Browser : Edge
- 
**Additional context**

bprusinowski:
Fixes #857.


app/configurator/components/field.tsx
```diff
@@ -171,25 +171,33 @@ export const DataFilterSelect = ({
...
 
-  const allOptions = useMemo(() => {
+  const sortedValues = useMemo(() => {
+    const sorters = makeDimensionValueSorters(dimension);
+    const sortedValues = orderBy(dimension.values, sorters);
+
+    return sortedValues;
+  }, [dimension]);
+
+  const allValues = useMemo(() => {
```
bprusinowski, line 181: I wonder if here we shouldn't apply the same, fixed approach as in hierarchy tree sorting ["position", "identifier", "label"]. Not sure if values in filters should be affected by chart sorting, e.g. by measure.

ptbrowne (approved):
🔥

app/configurator/components/field.tsx
ptbrowne: Yes it makes sense that it should not be dependent on the legend display, as as a chart publisher you'd want to always see the same order 👍 Isn't it the case here since we do not pass a sorting to makeDimensionValueSorters ? Seems to me that in this case we should default to the defaultSorters  ?

```
  const defaultSorters = [getPosition, getIdentifier, getLabel];
```

bprusinowski: Yes, I just realised that 👍

PR #874, opened 2022-11-21
feat: Adjustments to select tree

ptbrowne:
- Add check icon if value is selected
- Anchor select tree popover to the bottom of the anchor
- Fix popover position when select is opened programmatically


bprusinowski (approved):
👍

PR #879, opened 2022-11-21
Deactivate decimal formatting

ptbrowne:
- fix: Click behavior is no longer different depending on where you click on the row
- feat: Disable decimal formatting


ptbrowne:
This PR is now outdated.

Issue #850, opened 2022-11-15
Interactive Filters: Show Datafilters available even if there is no filter


AnninaWalker:
**Describe the bug**
In the interactive Filter Panel the Show Data Filter Checkbox can be activated although no filter is available.

**To Reproduce**
1. Go to 'test.visualize.ch'
2. Click on '[Photovoltaikanlagen Dataset](https://test.visualize.admin.ch/en/browse/dataset/https%3A%2F%2Fenergy.ld.admin.ch%2Fsfoe%2Fbfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen%2F5?previous=%7B%22includeDrafts%22%3Afalse%7D&dataSource=Prod) '
3. Add Kanton as Color Segementation
4. Scroll Down to Datafilters and activate Show data filters

**Expected behavior**
Data Filters shouldn't be available in the Interactive Filters sections if there are none to select. 

**Screenshots**

https://user-images.githubusercontent.com/59613691/201955254-97ab99c0-b7d7-4910-8a7f-22a6e7032ed8.mov

**Environment (please complete the following information):**
 - Visualize environment and version: [e.g. PROD v3.12.0]
 - Browser and version [e.g. Chrome 95]

ptbrowne:
What I see we could do:

* If the user cannot select any data filters, we show in the panel "All dimensions have been encoded onto the chart, no dimensions are available to server as interactive data filters", and we disable the checkbox
* When the user has activated data filters, and we have selected one dimension as data filters (for ex "Kanton"), and then we encode this dimension into the chart, we deactivate the data filters if the resulting number of data filter dimensions is 0.
<img width="323" alt="image" src="https://user-images.githubusercontent.com/465582/203294849-b9b9a015-ef9e-4aec-8729-ee6d06ee83c0.png">

What do you think ?

AnninaWalker:
Hmm, shouldn't we just diplay the data filters at all if none are available? Initially the concept within the Interactive Filters was, that we only show what can actually be enabled. 
Thinking of the new UI I guess it will solve itself as we would simply not show the toggle button.

AnninaWalker:
Ok, lets do it like this for now. As I mentioned above, in the new UI it will be different in any case

ptbrowne:
Fixed in 50ea5f813bde7854b0a0e72bfe2c1b4d46f39afb

PR #895, opened 2022-11-22
feat: Handle unavailable data filters (when all dimensions have been encoded)

ptbrowne:
Fix https://github.com/visualize-admin/visualization-tool/issues/850
See https://github.com/visualize-admin/visualization-tool/issues/850#issuecomment-1323474821

### DIFF
--- a/app/components/form.tsx
+++ b/app/components/form.tsx
@@ -195,7 +195,7 @@ export const Checkbox = ({
   smaller,
   indeterminate,
 }: {
-  label: string;
+  label: string | React.ReactNode;
   disabled?: boolean;
   color?: string;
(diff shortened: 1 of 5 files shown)


bprusinowski (approved):
I think that based on the comments in #850, the Data Filters section should show up even if there are no filters to be used there, and it is not the case when we have a temporal dimension not mapped (see a screenshot below from Bathing water quality (with Hierarchie) dataset, where the Data Filters section disappeared).

Screenshot 2022-11-22 at 14 51 11

Otherwise looks good 👌

ptbrowne:
Superseded by https://github.com/visualize-admin/visualization-tool/pull/902

PR #902, opened 2022-11-22
In place interactive filters

ptbrowne:
This moves the checkbox for "interactive" from the dedicated "interactive filters" section to the data filters and segment settings. The date filters changes will be in another PR.

## How to test

- Go to any dataset
- "Interactive" switches should be inside the "Filters" section on the left
- "Interactive" switch is also available for segments


bprusinowski (approved):
Really nice, the change of keeping interactive filters enabling next to the actual filters makes it much more smooth, I really like this! 👏 

I've noticed several things after some testing (using Bathing water quality [with Hierarchie]):
- it's possible to make X field interactive, but nothing happens then (when Date is mapped here),
- when enabling color mapping (Observation Location), it's automatically set as interactive – I am not sure if this is the best approach, taking into consideration the fact that then legend displays as one giant column 😅
- there is still an "Interactive filters" section at the bottom, shouldn't this be removed?

PR #906, opened 2022-11-23
feat: Metadata panel v1

bprusinowski:
TODO:
- [ ] add expanding of long dimension descriptions + autocomplete
- [ ] add a way of opening the panel by clicking on values in chart
- [ ] add shrinking of chart after the panel is opened in published / embed mode
- [ ] add tests
- [ ] do not show `Data` tab when no metadata is attached to dimensions
- [ ] add animations
- [ ] dimension values?


app/components/metadata-panel.tsx
```diff
@@ -0,0 +1,121 @@
...
+  children: React.ReactNode;
+}) => {
+  const [open, setOpen] = React.useState(false);
+
+  const ctx = React.useMemo(() => {
+    return {
+      open,
+      toggle: () => setOpen(!open),
+    };
+  }, [open, setOpen]);
```
ptbrowne: useDisclosure is helpful for those usecases.

app/icons/index.tsx
```diff
@@ -44,3 +53,23 @@ export const getChartIcon = (chartType: ChartType): IconName => {
...
+      return "chartPie";
+    case "NumericalMeasure":
+      return "numerical";
+    case "OrdinalDimension":
+      return "sort";
+    case "OrdinalMeasure":
+      return "sort";
+    case "TemporalDimension":
+      return "pointInTime";
+  }
```
ptbrowne, line 74: could we have a exhaustive check here ?

I think `return dimensionType as never` should work.

ptbrowne (approved):
👏  I think we could merge if we used a feature flag to activate / deactivate the button to open the metadata panel ?

app/components/metadata-panel.tsx
```diff
@@ -0,0 +1,121 @@
...
+    <ContextProvider>
+      <PanelInner />
+    </ContextProvider>
+  );
+};
+
+const ToggleButton = ({ onClick }: { onClick: () => void }) => {
+  return (
+    <Button component="a" variant="text" size="small" onClick={onClick}>
+      <Typography variant="body2">Metadata</Typography>
```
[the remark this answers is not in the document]
bprusinowski: I think we could maybe make it bold when the panel is open?

cc @AnninaWalker – it's about the button below :)

Screenshot 2022-11-28 at 09 36 09

ptbrowne (approved):

PR #913, opened 2022-11-28
feat: Metadata panel v2

bprusinowski:
This PR includes enhancements to the metadata panel.

### TODO
- [x] Tab switching animations
- [x] Expandable dimension description
- [x] Autocomplete (jump to)
- [x] Way of opening the panel by clicking on values in chart (and remove tooltips)
- [x] Do not share state across one store
- [x] Shrinking of chart after the panel is opened in published / embed mode
  - [x] Only shrink when it makes sense
- [ ] Fix focus of panels on the homepage when metadata panel is opened in another chart
- [ ] Do not show `Data` tab when no metadata is attached to dimensions
- [ ] Optional: Fetching of dimension values' descriptions
- [ ] Optional: Tests

### How to test
TBD


app/components/chart-filters-list.tsx
```diff
@@ -18,9 +40,26 @@ export const ChartFiltersList = ({
...
+        setSelectedDimension: state.setSelectedDimension,
+      }),
+      shallow
+    );
+
+  const handleClick = useEvent((d: DimensionMetadataFragment) => {
+    setOpen(true);
+    setActiveSection("data");
+    setSelectedDimension(d);
+  });
```
ptbrowne, line 61: I wonder if we should not have a method openDimension in the store that would take care of the internals of activeSection, open-ness, and selected dimension ? This way it's not the component job to know about this.

app/components/metadata-panel.tsx
```diff
@@ -158,6 +177,37 @@ const animationProps: Transition = {
...
+  children,
+}: {
+  dim: DimensionMetadataFragment;
+  svg?: boolean;
+  children: React.ReactNode;
+}) => {
+  const classes = useOtherStyles();
+  const openDimenion = useMetadataPanelStore((state) => state.openDimension);
+  const handleClick = useEvent(() => {
+    openDimenion(dim);
```
ptbrowne, line 201: typo: openDimension

ptbrowne (approved):

app/components/metadata-panel.tsx
```diff
@@ -49,32 +49,37 @@ type MetadataPanelState = {
...
+      set({ open: true, activeSection: "data", selectedDimension: d });
+    },
+    reset: () => {
+      set({ activeSection: "general", selectedDimension: undefined });
+    },
+  }));
+
+const defaultStore = createMetadataPanelStore();
+
+export const MetadataPanelStoreContext = React.createContext(defaultStore);
```
ptbrowne, line 82: We could have `useMetadataPanelStore` here right ? This would keep the same API as before for consumers.

```
const useMetadataPanelStore = (selector) => {
	const store = useContext(MetadataPanelStoreContext)
	return useStore(store, selector)
}
```

app/components/metadata-panel.tsx
```diff
@@ -196,9 +201,12 @@ export const OpenMetadataPanelWrapper = ({
   children: React.ReactNode;
 }) => {
   const classes = useOtherStyles();
-  const openDimenion = useMetadataPanelStore((state) => state.openDimension);
+  const store = React.useContext(MetadataPanelStoreContext);
+  const { openDimension } = useStore(store, (state) => ({
+    openDimension: state.openDimension,
+  }));
```
ptbrowne, line 207: We could only have the openDimension returned here, here since we always return a new object from the state, we will get unnecessary updates. See https://tkdodo.eu/blog/working-with-zustand#prefer-atomic-selectors

ptbrowne (approved):
👍 LGTM ! Thanks !

ptbrowne (approved):

ptbrowne (approved):
👏

PR #922, opened 2022-12-02
feat: Metadata panel enhancements

### LINKED ISSUE #917: Metadata panel – enhancements
- [ ] Add fetching of dimension's values descriptions
- [ ] Fix focus issue on the homepage, when a panel is open in the first chart, and the user tries to open another panel in the second chart
- [ ] Add links to open metadata panel from the tables (but keep them for dataset preview table)
- [ ] Add E2E tests

bprusinowski:
Closes #917.

- Table headers now point to metadata panel, the tooltip is only present for dataset preview before user starts creating a chart
- Fixed a focus issue on homepage (focus is not enforced to last opened panel)
- E2E test for opening the panel when clicking on elements inside the editor

cc: @AnninaWalker

### DIFF
--- a/app/components/metadata-panel.tsx
+++ b/app/components/metadata-panel.tsx
@@ -294,6 +294,7 @@ export const MetadataPanel = ({
         hideBackdrop
         ModalProps={{ container }}
         PaperProps={{ style: { position: "absolute" } }}
+        disableEnforceFocus
       >
         <Header onClose={handleToggle} />
 
--- a/app/components/metadata-panel.tsx
+++ b/app/components/metadata-panel.tsx
@@ -139,6 +139,7 @@ const useOtherStyles = makeStyles<Theme>((theme) => {
     },
     tabPanel: {
       padding: 0,
+      paddingBottom: theme.spacing(3),
     },
     tabPanelContent: {
       display: "flex",
@@ -155,11 +156,10 @@ const useOtherStyles = makeStyles<Theme>((theme) => {
         marginLeft: 0,
       },
     },
-    dimensionIcon: {
-      display: "inline",
-      marginLeft: -2,
-      marginTop: -3,
-      marginRight: 2,
+    dimensionValues: {
+      flexDirection: "column",
+      gap: theme.spacing(2),
+      marginTop: theme.spacing(3),
     },
     search: {
       "& .MuiAutocomplete-inputRoot": {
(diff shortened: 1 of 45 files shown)


codegen.yml
```diff
@@ -13,6 +13,13 @@ generates:
       withComponent: false
       withHooks: true
       contextType: ./context#GraphQLContext
+      scalars:
+        Observation: "../domain/data#Observation"
+        ObservationValue: "../domain/data#ObservationValue"
+        DimensionValue: "../domain/data#DimensionValue"
+        RawObservation: "../domain/data#RawObservation"
+        Filters: "../configurator#QueryFilters"
+        GeoShape: "../domain/data#GeoShape"
```
ptbrowne, line 22: Can you explain to me what is this change for ?

app/components/metadata-panel.tsx
```diff
@@ -514,37 +503,70 @@ const TabPanelDataDimension = ({
...
+                  )}
+                </Typography>
+                {d.description ? (
+                  <Typography variant="caption">{d.description}</Typography>
+                ) : null}
+              </React.Fragment>
+            ))}
+          </MotionBox>
+        )}
+      </AnimatePresence>
```
ptbrowne, line 565: I feel like it'd be interesting to have the dimension values or for example the min/max as a gray text under the dimension name in the panel describing all the dimensions (so that you do not have to open the panel in most cases, and also to give a taste to the user of what she will find in the details). This is not for the PR, but something to keep in mind. @AnninaWalker

e2e/metadata-panel.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+    "Prod"
+  );
+
+  const checkKantonDescription = async () => {
+    const kantonDimensionDescription = await selectors.panels
+      .metadata()
+      .within()
+      .findByText("Kanton, in welchem die geförderten Anlagen stehen");
+
+    expect(kantonDimensionDescription).toBeDefined();
```
ptbrowne, line 18: Nit: I am not sure it is necessary, since the find will throw otherwise. At least it's explicit. Just a thought, not something to fix.

ptbrowne (approved):
Looking good :) I did not really understand the changes around graphql scalars though, if you could explain what it is for, I would be interested ;)

PR #925, opened 2022-12-05
feat: Proof of concept of mobx

ptbrowne:
This is a proof of concept of how mobx could help with state management and computed values.

A cube class is created and made observable. It is responsible for holding and loading the data.
It is observable so components decorated with observer will automatically react to changes.

See the _mobx page.


app/pages/_mobx.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+    try {
+      this.loading = true;
+      const res: OperationResult<DataCubeMetadataWithComponentValuesQuery> =
+        yield this.client
+          .query<DataCubeMetadataWithComponentValuesQuery>(
+            DataCubeMetadataWithComponentValuesDocument,
+            vars
+          )
+          .toPromise();
```
bprusinowski, line 91: `*load()` + yielding a promise to avoid doing `await Cube.load()`?

bprusinowski (approved):
Looks fantastic, I really like this approach 🎉 We should try to incorporate this in the app in the future, as mentioned in PR's review 👌

app/pages/_mobx.tsx
ptbrowne: This is unfortunately the way to do [async `actions`](https://mobx.js.org/actions.html#using-flow-instead-of-async--await-) in mobx

PR #934, opened 2023-01-17
feat: Time slider continuation

bprusinowski:
This PR continues the work on the `TimeSlider`,
- aligning it with the [design](https://www.notion.so/interactivethings/V02-Timeslider-85eec9497bed42c0a8b762d1e699a59f),
- making it animatable,
- enabling it for more chart types (grouped / stacked column, pie, scatterplot),
- fixing chart scale domains in place.

UI of slider's settings, stepped mode & introduction to maps will be tackled in subsequent PRs.

### Testing
1. Go to `/en/create/new?cube=https://environment.ld.admin.ch/foen/ubd0104/6&flag__timeslider=true`
2. Change X field from `Date` to `Parameter`
3. Enable Time Slider (Interactive Filters)
4. See that the slider appeared and it's possible to animate the chart

### TODO
- [x] Add data option to observable Timeline
- [x] Add tests


app/charts/column/columns-grouped-state.tsx
```diff
@@ -267,23 +266,19 @@ const useGroupedColumnsState = (
...
-      ) as number,
+        getYErrorRange ? getYErrorRange(d)[1] : getY(d)
+      ) ?? 0,
       0
     );
 
-    const yScale = scaleLinear()
-      .domain([mkNumber(minValue), mkNumber(maxValue)])
-      .nice();
+    const yScale = scaleLinear().domain([minValue, maxValue]).nice();
```
ptbrowne, line 281: It is because were are sure that values in preparedData are numbers and not strings ?

app/charts/shared/chart-helpers.tsx
```diff
@@ -101,17 +101,22 @@ export const usePlottableData = ({
...
-/** Prepares the data to be used in charts.
+/** Prepares the data to be used in charts, taking interactive filters into account.
  *
- * Different than the full dataset, because interactive filters might be applied.
+ * @returns `preparedData`: data to be used in the chart,
+ * `scalesData`: data to be used to compute the scales.
+ *
+ * They are different when a time slider is present, since the scales should be computed using all the data,
+ * to prevent them from changing when the time slider is moved, while the chart should only show the data
+ * corresponding to the selected time.
```
ptbrowne, line 111: 👍

app/charts/column/columns-grouped-state.tsx
bprusinowski: It just wasn't necessary to use the function, as above it we already made sure `minValue` & `maxValue` are numbers (`... || 0`). I changed the logic slightly to use `?? 0` :)

app/themes/federal.tsx
```diff
@@ -613,6 +613,17 @@ theme.components = {
             display: "none",
           },
         },
+        "& .MuiSlider-valueLabel": {
+          background: "white",
```
ptbrowne, line 617: theme.palette.background.paper

bprusinowski:
That's a fantastic suggestion, I've never really used this pattern but I agree it makes this a lot clearer. I've updated the code, thanks 💯

app/configurator/interactive-filters/time-slider.tsx
```diff
@@ -12,57 +12,29 @@ import { useTimeFormatUnit } from "@/formatters";
...
+    const callback = () => setS((s) => s + 1);
+    timeline?.subscribe(callback);
 
-  if (!ctx) {
-    throw new Error(
-      "You need to wrap your component in <TimeSliderStateProvider /> to useTimeSliderContext()"
-    );
-  }
+    return () => timeline?.unsubscribe(callback);
+  }, [timeline]);
```
ptbrowne, line 28: I think it could be a bit improved by not tying it to the timeline.

Maybe have a useObservable(observable: Observable) and then a useTimeline that uses this and tying it to the context.

app/configurator/interactive-filters/time-slider.tsx
```diff
@@ -226,7 +140,7 @@ const PlayButton = () => {
         borderRadius: "50%",
       }}
     >
-      <Icon name={animating ? "pause" : "play"} size={16} />
+      <Icon name={timeline?.playing ? "pause" : "play"} size={16} />
```
ptbrowne, line 137: To prevent this optional operator, I suggest that useTimeline throws if not inside the context, this way we can reflect in the type that the result of useTimeline is Timeline and not Timeline | undefined.

app/configurator/interactive-filters/time-slider.tsx
```diff
@@ -12,57 +12,29 @@ import { useTimeFormatUnit } from "@/formatters";
...
 
-const TimeSliderStateContext = React.createContext<TimeSliderState>({
-  progress: 0,
-  setProgress: () => {},
-  animating: false,
-  setAnimating: () => {},
-});
+const useTimeline = () => {
+  // This is a "hack" to force re-rendering of the component using the hook.
+  const [_, setS] = React.useState(0);
```
ptbrowne, line 20: I think on React 18 we could remove the need for the stack hack through useSyncExternalStore.

https://beta.reactjs.org/reference/react/useSyncExternalStore#usesyncexternalstore

```
class Timeline extends Observable {
  getSnapshot() {
   // hopefully returning the timeline itself is OK for a snapshot
    return this
  }
}

const useTimeline = () => {
  const timeline = React.useContext(TimelineContext);
  if (!timeline) {
    throw new Error("useTimeline must be called inside a TimelineContext.Provider")
  }
  useSyncExternalStore(timeline.subscribe, timeline.getSnapshot)
  ...
}
```

ptbrowne (approved):
Really nice, thank you !

app/configurator/interactive-filters/time-slider.tsx
ptbrowne: It seems a package offers a polyfill for older versions of React, I think it's worth it trying it out.

bprusinowski: Cool, another nice suggestion :) I've modified the code to use this approach, thanks!

ptbrowne (approved):
💯

PR #935, opened 2023-01-18
Fixes

### LINKED ISSUE #931: "Cascading filter" behaviour by hierarchical dimensions
The "Cascading filter" behaviour does not work correctly on filters containing hierarchical dimensions.

**To Reproduce**

1. Open the Draft dataset "Greenhouse gas emissions by sector and by gas" (include Draft) and start a visualization
2. Open the Filter "Emission sources" and select "Synthetic gases" (Sectors Ordinance CO2 -> Other)

image

5. Open the second Filter "Greenhouse gas"

7. See error
-> all gases are displayed, while there are only values for synthetic gases
-> When I select a gas for which there are no values for the selected emission source, e.g. Methane, another gas is selected, in our example "Perfluorocarbons" without the user being warned.

image

**Expected behavior**

Only available values should be displayed in the filters (in our example: synthetic gases).

image

- Visualize environment and version: PROD v3.16.0 ([5fccbf0]
- Browser and version: Edge

**Additional context**

### LINKED ISSUE #933: An error message did appear in the user interface instead of the filter selection
**To Reproduce**
I do not know why, but I guess there should be an interception of this error message

**Screenshots or video**
image

- Visualize environment and version: INT v3.18.0 ([91ae9f5]
- Browser and version: Firefox 108.0.2

ptbrowne:
- fix: Dev server would fail to run without the flag. I do not know why this problem started.
- feat: Better error handling for possible filters, fixes #933 
- fix: Hierarchical filters only show possible values, fixes #931

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -5,7 +5,9 @@ let
 in pkgs.mkShell {
   buildInputs = [
     pkgs.darwin.apple_sdk.frameworks.CoreServices
-    pkgs.yarn
     nodejs
+    (pkgs.yarn.override {
+      nodejs = null;
+    })
   ];
 }
--- a/app/configurator/components/chart-configurator.tsx
+++ b/app/configurator/components/chart-configurator.tsx
@@ -223,7 +223,7 @@ const useEnsurePossibleFilters = ({
       if (error || !data) {
         setError(error);
         setFetching(false);
-        console.warn("Could not fetch possible filters", error);
+        console.error("Could not fetch possible filters", error);
         return;
       }
       setError(undefined);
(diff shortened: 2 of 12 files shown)


shell.nix
```diff
@@ -5,7 +5,9 @@ let
 in pkgs.mkShell {
   buildInputs = [
     pkgs.darwin.apple_sdk.frameworks.CoreServices
-    pkgs.yarn
     nodejs
+    (pkgs.yarn.override {
+      nodejs = null;
```
ptbrowne, line 10: Make it so that yarn uses the same node version as the node version we specify.

bprusinowski (approved):
LGTM, thanks for the explanation 🙇

PR #940, opened 2023-01-20
feat: Add Animation for Column, Pie and Scatter charts

### LINKED ISSUE #834: First Iteration of Animated Charts / Timeslider
**DoD**
- Implement Design of time slider
-  Fix chart container size so it doesn't jump 
- Feature will be applied to the following chart types ONLY: Scatter, Bar, Pie (Maps, lines, areas, tables are OUT OF SCOPE for first iteration) 

**Nice to Have**  
- Add the ability to run an animation with a play and pause button (constant speed: no options to speed up or slow down)

Gitlab: https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/444

bprusinowski:
Closes #834.

## Goals / Scope

This PR introduces Animation field which is possible to use with Column, Pie and Scatter charts (more charts to be implemented in the future) and removes `timeSlider` from interactive filters.

## Description

Additionally, I included a proposed change to use D3.js to animate columns in non-grouped and non-stacked column charts to greatly improve the interactions with the animation. I am curious what you think about this approach @ptbrowne, if you think it's fine, then I can expand this to other charts (and extract the logic to be reusable).

## Comments

I've extracted the comments from below discussion into two new issues to be tackled separately from this PR (#1029 #1030).

I think it would make sense to have some sort of UI feedback when starting stepped animation (I've thought about a progress circle around the play button that would do one lap in between the animation jumps). But I think it could also be tackled in another PR, possibly with some design support :)

## How to test

1. Go to [link](https://visualization-tool-git-feat-time-slider-continuation-2-ixt1.vercel.app/en/create/new?cube=https://environment.ld.admin.ch/foen/ubd0104/6).
2. Change horizontal axis to `Parameter`.
3. Select `Date` and Animation encoding.
4. See that it's possible to run the animation in two modes, change the duration and hide the play button.


bprusinowski:
@AnninaWalker if you'd like to test the current version of the time slider, you can do so [here](https://visualization-tool-git-feat-time-slider-continuation-2-ixt1.vercel.app/en?dataSource=Prod&flag__timeslider=true) (should work for columns, pies and scatterplots) :)

AnninaWalker:
Hey @bprusinowski Following my feedback:
Per default the Time Dimension is already selected in the Horizontal Axis section. (for ex. Bathing Water dataset). If it is selected there I've noticed that when I click on the Animation section, that I can't select the Time-Dimension, as it is disabled in the Dropdown-Menu.
To avoid a dead end to the User I'd see following two solutions: Either, if there is no selectable dimension available, the user can't click on the edit icon. Instead he sees an info icon with a tooltip (No dimension available) -> Check layout in Skecth. I've created this screen already.

Or another scenario that I could see, that the user can select the time dimension in the Animation section and that the Selection of the Horizontal Axis Dimension changes automatically to another dimension.
Screenshot 2023-01-24 at 09 41 07
Screenshot 2023-01-24 at 09 42 18

From a user flow I'd probably prefer the latter scenario. What do you think?

AnninaWalker:
A minor detail regarding UI. The Slider could be directly displayed below the dropdown without the filter heading. As in the layout. See below.
Screenshot 2023-01-24 at 09 53 27

AnninaWalker:
Is it impossible to improve the layout of the pie chart. On my big screen the pie chart is huge. But the animation slider is unfortunately out of my viewport. I think this could be optimized.
Screenshot 2023-01-24 at 09 55 38

AnninaWalker:
Same applies to Scatterplot. They are also huge and you cant see the full graph at once

AnninaWalker:
Screenshot 2023-01-24 at 10 00 18

AnninaWalker:
I think it would be helpful to add tooltips to the slider element in the left side panel too. Or otherwise some inputfields, as I had them in the design initially. Otherwise, currrently when I'm draging the slider, I have no clue what year I'm going to select . 
Screenshot 2023-01-24 at 10 01 04

ptbrowne:
> Is it impossible to improve the layout of the pie chart. On my big screen the pie chart is huge. But the animation slider is unfortunately out of my viewport. I think this could be optimized.

For now charts, take the whole screen horizontally, and to keep the aspect ratio, they also grow vertically. I think this layout problem affects all the charts. The solution I see would be to constrain the height of the charts to a maximum value (so event on bigger screen, charts would not get over 400px big). @AnninaWalker What do you think ?

ptbrowne:
> Or another scenario that I could see, that the user can select the time dimension in the Animation section and that the Selection of the Horizontal Axis Dimension changes automatically to another dimension.

I agree that I prefer the flow where the user is not blocked. Otherwise, the user would have to select another field in horiztonal axis, then select animation field.

1. Do you think we should apply this behavior to other fields ? For now, we can only select dimensions that have not been mapped. I think if we make this decision for the animation field, then it should be done across other fields too.
2. If we put already-mapped fields inside the selects, I think they should be marked with the dimension they are mapped to, so that the user can better understand the behavior. We could for example add the icon of the dimension (horizontal axis, vertical axis, animation field) to the right of the option, to indicate to the user that the dimension has already been mapped. Then if the user selects it, I think it would be more understandable why the horizontal axis then changes automatically.

AnninaWalker:
Hmm what do you think @ptbrowne Shall we restrict the height? I personally think the charts are often too big and its a pity when the Metainfo below is out of the Viewport.

AnninaWalker:
@ptbrowne Good input. You are right. If we apply it here it would make sense to apply this behaviour anywhere. And I also like the idea to indicate in case it is already mapped somewhere else that the user understand that this will switch. 👍

Rdataflow:
@bprusinowski @ptbrowne this feature is awesome! 
can you ensure animation is also possible for NFI cubes on the dimension `Inventory`?
i.e. https://visualization-tool-git-feat-time-slider-continuation-2-ixt1.vercel.app/en/browse?dataset=https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fnfi%2FC-96%2Fcube%2F1&dataSource=Int 

FYI: currently this key is nominal - but we tend to switch it to ordinal soonish and in the future it might become dataKind temporal - once visualize tolerates temporal dataKind w/o unit (currently ordinal dataKind time w/o unit fails with `[GraphQL] Cannot return null for non-nullable field TemporalDimension.timeUnit.`)

cc @kronmar

bprusinowski:
Hi @Rdataflow, thanks for the feedback :) I need to think about this, because I see some potential challenges:
- if we allow animation for nominal / ordinal variables it might be confusing for the users, as generally we should only be allowed to animate time dimensions,
- if we allow temporal dimensions without units, we would have problems across the whole application, as we use this to correctly format dates,
- if we create an artificial time unit (probably in Cube Creator?), such as "flexible time frame" or something similar, how would we sort values? In the case of "Inventory" dimensions, we have the convenience of values named "N1", "N2", which would sort the values correctly, but right now for the time slider we use min / max value to position the values on track slider track. If we allowed to have "nominal values", we would need to have some logic to sort them correctly – that might be possible to do with `order`, but I am not sure if this is the best approach,
- maybe it would be possible to create additional dimension, which would be "correctly" temporal (and be based on Inventory)? This seems like the best solution to avoid some edge cases.

cc @ptbrowne

Rdataflow:
@bprusinowski I share certain of your thoughts:
* nominal: this would be difficult as by definition nominal has no order - OTOH ordinal is fine
* temporal w/o units: while most temporal dimensions have units - others don't by design see [userstory](https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/78#note_8710)
* sort order of ordinal time: ordinal key dimensions by schema definition have their order in [schema:position](https://cube.link/#qudt-scaletype)

therefore let's take into account also ordinal time w/o units for animations.

app/configurator/components/chart-options-selector.tsx
```diff
@@ -475,29 +475,60 @@ const ChartFieldAnimation = ({ field }: { field: AnimationField }) => {
...
-                field="animation"
-                path="duration"
-                value={60}
+              <Flex sx={{ justifyContent: "flex-start" }}>
+                <ChartOptionRadioField
+                  label="30s"
+                  field="animation"
+                  path="duration"
+                  value={30}
+                />
```
ptbrowne, line 493: I think we could add 10s also;

app/charts/column/columns-simple.tsx
```diff
@@ -41,31 +43,92 @@ export const ErrorWhiskers = () => {
...
+  React.useEffect(() => {
+    if (ref.current) {
+      const g = select(ref.current);
+      g.selectAll<SVGRectElement, ChartDatum>("rect")
+        .data(chartData, (d) => d.x)
+        .join(
+          (enter) =>
+            enter
+              .append("rect")
+              .attr("data-index", (_, i) => i)
```
ptbrowne, line 101: What is the purpose of this data-index attribute ? I am not seeing it used in the update/exit transitions.

ptbrowne: Ah ok I see it is used in a test.

app/charts/shared/chart-helpers.tsx
```diff
@@ -37,17 +37,13 @@ export const prepareQueryFilters = (
...
     queryFilters = { ...queryFilters, ...dataFilters };
   }
 
-  queryFilters = omitBy(queryFilters, (v, k) => {
-    return (
-      (v.type === "single" && v.value === FIELD_VALUE_NONE) ||
-      k === timeSlider?.componentIri
-    );
+  queryFilters = omitBy(queryFilters, (v) => {
+    return v.type === "single" && v.value === FIELD_VALUE_NONE;
```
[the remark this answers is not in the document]
bprusinowski: There was a need to remove it as we ditched time slider from interactiveFiltersConfig and added Animation field instead. We do not include the iri from the Animation field in filters, so it's not filtered on the query level but dynamically when preparing the data 👍

app/charts/column/chart-column.tsx
```diff
@@ -111,9 +111,9 @@ export const ChartColumns = memo(
               }
             />
 
-            {interactiveFiltersConfig?.timeSlider.componentIri && (
+            {fields.animation && (
```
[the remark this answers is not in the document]
bprusinowski: Mainly due to design, as it was supposed to be a field and not an interactive filter, which are located in a different place in the UI 👀

app/charts/column/columns-simple.tsx
bprusinowski: 👍

PR #945, opened 2023-01-24
Save draft config

ptbrowne:
Instead of saving the chart locally while it has not been published, we save it to the database.
This lets us share the creation link before the chart is published.
While a chart is not published, it is in draft

- [x] Ability to update a chart
- [x] Chart is auto saved when it changes 
- [x] Share button next to publish to get either a URL bringing the person to a copy of chart, or to the chart with edit rights
- [x] Chart in draft should not be viewable or embeddable
- [x] When changing the chart title and re-opening the page, some letters are removed
- [ ] Clean up of draft charts that have not been updated in a while

### DIFF
--- a/app/components/chart-selection-tabs.tsx
+++ b/app/components/chart-selection-tabs.tsx
@@ -1,5 +1,13 @@
-import { Trans } from "@lingui/macro";
-import { Box, Popover, Tab, Tabs, Theme, Button } from "@mui/material";
+import { t, Trans } from "@lingui/macro";
+import {
+  Box,
+  Popover,
+  Tab,
+  Tabs,
+  Theme,
+  Button,
+  Typography,
+} from "@mui/material";
 import { makeStyles } from "@mui/styles";
 import React, {
   createContext,
@@ -7,6 +15,7 @@ import React, {
   ReactNode,
   useContext,
   useEffect,
+  useRef,
   useState,
 } from "react";
 
@@ -18,11 +27,14 @@ import {
 } from "@/configurator";
 import { ChartTypeSelector } from "@/configurator/components/chart-type-selector";
 import { getIconName } from "@/configurator/components/ui-helpers";
+import useDisclosure from "@/configurator/components/use-disclosure";
 import { useDataCubeMetadataWithComponentValuesQuery } from "@/graphql/query-hooks";
 import { Icon, IconName } from "@/icons";
 import { useLocale } from "@/src";
 import useEvent from "@/utils/use-event";
 
+import { Boldify } from "./boldify";
+import { CopyToClipboardTextInput } from "./copy-to-clipboard-text-input";
 import Flex from "./flex";
 
 type TabsState = {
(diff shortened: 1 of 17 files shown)


app/pages/api/config.ts
```diff
@@ -1,34 +1,17 @@
-import { NextApiRequest, NextApiResponse } from "next";
-
 import { createConfig } from "@/db/config";
```
bprusinowski, line 1: Nit: `createConfig as createConfigInDatabase`? When reading the code below initially I asked myself "why the config in being created, when we need to sent it to a db?" - maybe it could be worth to mention the db in this context?

app/server/nextkit.ts
```diff
@@ -0,0 +1,10 @@
+import createAPI from "nextkit";
+
+export const api = createAPI({
+  async onError() {
+    return {
+      status: 500,
+      message: "Something went wrong.",
+    };
+  },
+});
```
bprusinowski, line 10: Cool, much cleaner!

app/db/config.ts
```diff
@@ -22,15 +22,44 @@ type PublishedConfig = Omit<ConfigType, "activeField">;
...
+};
+
+/**
+ * Updates a config in the DB.
+ *
+ * @param data Data to be stored as configuration
+ */
+export const updateConfig = async (
+  key: string,
+  data: Exclude<Prisma.ConfigUpdateInput["data"], undefined>,
```
bprusinowski, line 50: Nit: `NonNullable`?

app/utils/chart-config/api.ts
```diff
@@ -10,6 +13,7 @@ export const createConfig = async (state: ConfiguratorStatePublishing) => {
   return apiFetch<InferAPIResponse<typeof apiConfigs, "POST">>("/api/config", {
     method: "POST",
     data: {
+      isDraft: true,
```
bprusinowski, line 18: Why? Shouldn't this depend on whether we publish or not?

app/configurator/configurator-state.tsx
```diff
@@ -1604,12 +1612,14 @@ const ConfiguratorStateProviderInternal = ({
...
+        }
+      } catch (e) {
+        console.error(e);
       }
-    } catch (e) {
-      console.error(e);
-    }
+    };
+    run();
   }, [state, dispatch, chartId, push, asPath, locale, query.from, replace]);
```
bprusinowski, line 1623: Nit: I think we do not need to update the chart in the db when locale changes.

app/components/chart-selection-tabs.tsx
```diff
@@ -184,6 +196,114 @@ const PublishChartButton = () => {
...
+            <Trans id="share-creation.popover.title">Share creation URL</Trans>
+          </Typography>
+          <div>
+            <Typography variant="h5" gutterBottom>
+              <Trans id="share-creation.popover.share-copy.title">
+                Share a copy
+              </Trans>
+            </Typography>
+            <CopyToClipboardTextInput
+              content={`${window.location.origin}/${locale}/create/${key}`}
```
bprusinowski, line 271: It seems that `key` can be undefined?

bprusinowski (dismissed):
Nice, a fantastic change 👏 💯 

I wasn't able to fully test the feature, as it doesn't work on preview environment (which is expected I guess, as we don't connect to a DB there), but overall looks good!

Which option below matches copying a URL directly from the browser? Will this work or should the menu be used for sharing the chart?

Screenshot 2023-01-24 at 17 45 36

app/server/nextkit.ts
ptbrowne: Most importantly, we get types client-side 🤤

ptbrowne:
> Which option below matches copying a URL directly from the browser? Will this work or should the menu be used for sharing the chart?

When copying the URL manually, you get editing rights. I think your question shows the need to expand/improve the help text like so "The other person will be able to edit your chart. This is the same URL as when you copy manually from the adress bar.". @AnninaWalker Any thoughts on this or ways to make it shorter ?

app/configurator/configurator-state.tsx
ptbrowne: True, fixed in 350744ca46f1ae350e47c8fa47049f34ae3befe3, the dependencies were not up to date.

app/utils/chart-config/api.ts
ptbrowne: Create is the first time. It's update that will be used for publish.

app/pages/api/config.ts
ptbrowne: I thought that the import is coming from a path with "db" in it was suffcient ?

bprusinowski: I meant that when looking at the logic, not the imports, it wasn't immediately apparent that it creates a config in db, and not just on a client side, but this is a minor thing and the "/db" path should be enough 👍

ptbrowne: OK, next time maybe I'll write `import * as db from '@/db/config'` so that it's clearer from the call site :)

app/configurator/configurator-state.tsx
```diff
@@ -1616,7 +1616,7 @@ const ConfiguratorStateProviderInternal = ({
       }
     };
     run();
-  }, [state, dispatch, chartId, push, asPath, locale, query.from, replace]);
+  }, [dispatch, push, replace, state]);
```
bprusinowski: 🎉

sosiology:
this can be closed now, correct?

KerstinFaye:
@ptbrowne As discussed in the meeting:
- Switch Draft & "Published Visualizations"
- Rename list "My Visualizations" to "Published Visualizations" (in the future not in two different tabs but in the same tab as sections underneath each other. I'm ok to leave it for now in two tabs until we have "My Favourite Datasets"- tab)
- Add action "Unpublish" to list of Published Visualizations within the more options button. Icon for this is here:
Icons_16dp_ic_unpublish
- [Agreed that tooltip instead of alert banner in the UI.](https://www.figma.com/file/Hn9wvuEYLUmdJtg4KwZYUx/visualize.admin.ch---design?type=design&node-id=1534-139839&mode=design&t=EgHkJQJfWHwiiceU-4)
- [Improvement for UI Design: Renaming overlay with Tabs (to rename multi charts in tab layout), add "save" and "cancel" button.](https://www.figma.com/file/Hn9wvuEYLUmdJtg4KwZYUx/visualize.admin.ch---design?type=design&node-id=3168-337228&mode=design&t=EgHkJQJfWHwiiceU-4)

One more question:
- When I go to "preview" a draft I don't have the action to edit directly from there. Is this possible? It only allows me to copy and edit?

ptbrowne:
Wow, I did not know that and had never notice this in macOS 👍 Thanks for the explanation and the screenshot :)

ptbrowne:
OK, just read a bit about that and it seems it differs from companies to companies across the web, and also depending on the country, the US being more prone to that. The rule seems to be "be consistent". Here is the article I read about this: https://uxloc.medium.com/when-to-use-sentence-case-and-title-case-161768eaab6e.

Looking at [admin.ch](https://www.admin.ch/gov/en/start.html), they do not use that much Title case. Do they have a rule about that in the official guidelines ?

ptbrowne:
> When I go to "preview" a draft I don't have the action to edit directly from there. Is this possible? It only allows me to copy and edit? 

If you are the author, of the chart you should see an Edit button in the warning banner, do you see it ? Did you expect to see it a the bottom of the chart ?

image

ptbrowne:
(I am going to apply Title Case now, but I think it is good to make sure that it is a visualize.admin guideline)

ptbrowne:
Yes it would be possible to differentiate the two views.

PR #947, opened 2023-01-27
Embed options

### LINKED ISSUE #943: Choice of displaying the links below the embedded graph
**User Story:** 
As an editor I would like to have the choice to embed my Visualization in a web page (iFrame or external application) without all the links below the chart and without links to the Metadata.
Information to be removed for the "Minimal View" (see also screen shot)

- Dataset
- Latest update
- Download data
- switch to table view
- Learn more about the dataset
- run SparQl query
- Metadata Button + all Links to Metadata

The link "created with Visualize" must be kept, but the label has to be replaced with "more Information" or something like that.
I do not have a concrete idee for the labels in the box “Embed Code”. Maybe you could have : “Embed full view” vs “Embed minimal view”.

The current Embed codes are of course still available for editors who wish to implement the full view.

Image

Image

Image

ptbrowne:
Adds an option in the embed popover to choose either minimal or standard view for the chart when it's embedded.

Technically, it adds a query parameter "embedOptions" that is then picked up by the view to hide some elements under the chart.

Fix #943 

How to test:

- Open on https://visualization-tool-git-feat-embed-options-ixt1.vercel.app/en/v/sO8qHplebSER?dataSource=Prod
- Click the embed options
- Choose either standard or minimal options.
- The page should refresh to show how the chart would be embedded (right now, on smaller screen, the refresh triggers a layout change, which makes the popover move a little, a possible solution could be to have an [alternate layout for published view](https://github.com/visualize-admin/visualization-tool/pull/948) but I am keeping the dicussion into the other PR)
- The codes should reflect the options and can be copy pasted into jsbin.com for example for testing. The resulting embedded chart should reflect the selected options

### DIFF
--- a/app/components/publish-actions.tsx
+++ b/app/components/publish-actions.tsx
@@ -103,7 +103,6 @@ export const Share = ({ configKey, locale }: EmbedShareProps) => {
               setAnchorEl(ev.target as HTMLElement);
             }}
             startIcon={<Icon name="linkExternal" size={16} />}
-            variant="outlined"
           >
             <Trans id="button.share">Share</Trans>
           </Button>
@@ -286,11 +285,11 @@ export const Embed = ({ configKey, locale }: EmbedShareProps) => {
                         Standard
                       </Trans>
                     </Typography>
-                    {/* <Typography variant="caption" display="block">
+                    <Typography variant="caption" display="block">
                       <Trans id="publication.embed.style.standard.caption">
-                        Chart, download the data links, attribution etc...
+                        Provides metadata and download links for the dataset
                       </Trans>
-                    </Typography> */}
+                    </Typography>
                   </div>
                 }
                 disableTypography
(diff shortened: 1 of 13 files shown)


app/components/publish-actions.tsx
```diff
@@ -193,13 +193,16 @@ export const Embed = ({ configKey, locale }: EmbedShareProps) => {
   const [embedIframeUrl, setEmbedIframeUrl] = useState("");
   const [embedAEMUrl, setEmbedAEMUrl] = useState("");
   const [embedOptions, setEmbedOptions] = useEmbedOptions();
+  console.log([embedOptions, setEmbedOptions]);
   const handleChange: RadioGroupProps["onChange"] = (_ev, value) => {
+    console.log(setEmbedOptions);
```
bprusinowski: 👁️

bprusinowski (approved):
LGTM! 👏

PR #955, opened 2023-02-03
Performance improvements

ptbrowne:
Introduce LRU caches for some SPARQL queries that could take a long time (>1000ms).
Due to the nature of LRU caches, queries that are not used are removed from the cache.
If a dataset has not been opened yet, it takes time since the cache is not primed but then all interactions with the dataset are snappier.

### DIFF
--- a/app/rdf/query-cache.ts
+++ b/app/rdf/query-cache.ts
@@ -1,0 +1,33 @@
+import StreamClient from "sparql-http-client";
+import { ParsingClient } from "sparql-http-client/ParsingClient";
+import { LRUCache, LRUCacheOptions } from "typescript-lru-cache";
+
+type SparqlClient = StreamClient | ParsingClient;
+type SparqlClientQuery = {
+  execute: (query: SparqlClient["query"], options: any) => Promise<unknown>;
+  build: () => string;
+};
+export const makeExecuteWithCache = <T>({
+  parse,
+  cacheOptions,
+}: {
+  parse: (v: any) => T;
+  cacheOptions: LRUCacheOptions<string, T>;
+}) => {
+  const cache = new LRUCache(cacheOptions);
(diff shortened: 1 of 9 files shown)


app/rdf/query-dimension-values.ts
```diff
@@ -215,6 +216,10 @@ export const loadMinMaxDimensionValues = async ({
       `Failed to fetch min max dimension values for ${datasetIri}, ${dimensionIri}.`
     );
   } finally {
-    return result[0];
+    const { minValue, maxValue } = result[0];
+    const min = parseObservationValue({ value: minValue }) ?? 0;
+    const max = parseObservationValue({ value: maxValue }) ?? 0;
+
+    return [min, max];
```
bprusinowski: 👍

app/rdf/query-cache.ts
```diff
@@ -0,0 +1,33 @@
...
+}: {
+  parse: (v: any) => T;
+  cacheOptions: LRUCacheOptions<string, T>;
+}) => {
+  const cache = new LRUCache(cacheOptions);
+  return async (sparqlClient: SparqlClient, query: SparqlClientQuery) => {
+    const key = `${sparqlClient.query.endpoint.endpointUrl} - ${query.build()}`;
+    const cached = cache.get(key);
+    if (cached) {
+      console.log("Found in cache", key);
```
bprusinowski: 👓

Rdataflow:
this PR brings up an interesting idea. Will there be a way to disable this cache? (i.e. url param `cache=false` which could be applied for E2E tests scenarios)

ptbrowne:
If the need arise, it will be quite straightforward to support this kind of option.

Rdataflow:
@ptbrowne fine, so let me pronounce the need for it. we need it to have representative comparable performance tests. see also #961

Issue #963, opened 2023-02-14
search on TEST broken


Rdataflow:
**Describe the bug**
![image](https://user-images.githubusercontent.com/26090254/218789811-46dd3135-3052-40b7-b073-d392644705e2.png)

**To Reproduce**

1. Go to https://test.visualize.admin.ch/browse?includeDrafts=true&order=SCORE&search=someimpossibleword&dataSource=Test

**Expected behavior**
empty list

**Environment (please complete the following information):**

- Visualize environment and version: [e.g. TEST v3.19.1] v3.19.1 ([fb87d10](https://github.com/visualize-admin/visualization-tool/commit/fb87d10567cc986922b9f227a28e597b90c4b0a8))

Issue #959, opened 2023-02-08
poor performance loading list of datasets


Rdataflow:
**Describe the bug**
while loading the list of datasets poor performance occurs.
the query to list the datasets behaves (incl. drafts) takes extraorinary long

**To Reproduce**

1. Go to https://int.visualize.admin.ch/browse?includeDrafts=true&dataSource=Int
2. observe extraordinary slow graphql query (50secs and more - up to 152secs for 332 datasets)

**Expected behavior**
quick load time

**Additional context**
don't know the exact reason. possibly this is due to multilingual building up cartesian?
```
curl 'https://test.visualize.admin.ch/api/graphql' \
  --data-raw $'{"query":"query DataCubes($sourceType: String\u0021, $sourceUrl: String\u0021, $locale: String\u0021, $query: String, $order: DataCubeResultOrder, $includeDrafts: Boolean, $filters: [DataCubeSearchFilter\u0021]) {\\n  dataCubes(\\n    sourceType: $sourceType\\n    sourceUrl: $sourceUrl\\n    locale: $locale\\n    query: $query\\n    order: $order\\n    includeDrafts: $includeDrafts\\n    filters: $filters\\n  ) {\\n    highlightedTitle\\n    highlightedDescription\\n    dataCube {\\n      iri\\n      title\\n      workExamples\\n      creator {\\n        iri\\n        label\\n        __typename\\n      }\\n      description\\n      publicationStatus\\n      datePublished\\n      themes {\\n        iri\\n        label\\n        __typename\\n      }\\n      __typename\\n    }\\n    __typename\\n  }\\n}\\n","operationName":"DataCubes","variables":{"sourceType":"sparql","sourceUrl":"https://int.lindas.admin.ch/query","locale":"de","includeDrafts":true,"filters":[]}}'
```
are languages COALESCEd or combined cartesian?

Rdataflow:
part of the problem seem to be in the query https://s.zazuko.com/HWrCrD returning 140'000 entries for the 332 datasets 👀

Rdataflow:
@ptbrowne just to inform you. the same query is executed twice. any reason for that?
![image](https://user-images.githubusercontent.com/26090254/219083562-e18e5ce4-10ce-4b35-b3fc-79dc81ffcf4b.png)

```sparql
#pragma describe.strategy cbd
#pragma join.hash off

PREFIX schema: <http://schema.org/>
PREFIX dcterms: <http://purl.org/dc/terms/>
[...]

PR #964, opened 2023-02-15
fix: Search performance and empty results

ptbrowne:
Fixes on search

- Empty list of IRIs was not taken into account
- Different languages lead to combinatorial explosion and poor performance as the dataset list grew 
Previously the table returned in the results would be like so

`cubeIRI` | `cube` | `theme` | `creator` 
--- | --- | --- | ---
1 | Snowfalls | Snow | WSL
1 | Schneefall | Schnee | WSL
1 | Snowfalls | Schnee | WSL
1 | Schneefall | Snow | WSL

So for a single cube we would have 4  (more generally N_LANGUAGE*N_FIELDS_WITH_LANGUAGE) different rows instead of 2 (one row for each language). The added equality filter ensures each row is in the same language.

Fix https://github.com/visualize-admin/visualization-tool/issues/963
Fix https://github.com/visualize-admin/visualization-tool/issues/959

### DIFF
--- a/app/rdf/query-search.ts
+++ b/app/rdf/query-search.ts
@@ -4,6 +4,7 @@ import clownface from "clownface";
 import { descending } from "d3";
 import { Cube } from "rdf-cube-view-query";
 import rdf from "rdf-ext";
+import { Quad, Stream } from "rdf-js";
 import StreamClient from "sparql-http-client";
 import ParsingClient from "sparql-http-client/ParsingClient";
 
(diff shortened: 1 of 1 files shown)


app/rdf/query-search.ts
```diff
@@ -168,6 +168,10 @@ export const searchCubes = async ({
     ${makeInFilter("theme", themeValues)}
     ${makeInFilter("creator", creatorValues)}
 
+    FILTER(LANG(?name) = LANG(?description))
+    FILTER(LANG(?name) = LANG(?creatorLabel))
+    FILTER(LANG(?name) = LANG(?themeName))
```
bprusinowski, line 173: How does this work?

bprusinowski (approved):
LGTM!

app/rdf/query-search.ts
ptbrowne: I edited the description, please tell me if it's clearer.

PR #983, opened 2023-02-24
Sorting values

ptbrowne:
1. Dimension values were not parsed when fetched in batch
2. The sorters from makeDimensionValueSorters return sorters that must receive dimension labels, not dimension values

### DIFF
--- a/app/configurator/components/field.tsx
+++ b/app/configurator/components/field.tsx
@@ -745,11 +745,7 @@ export const ChartFieldField = ({
       key={`select-${field}-dimension`}
       id={field}
       label={
-        <FieldLabel
-          isOptional={optional || false}
-          isFetching={fetching}
-          label={label}
-        />
+        <FieldLabel isOptional={optional} isFetching={fetching} label={label} />
       }
       disabled={disabled || fetching}
       options={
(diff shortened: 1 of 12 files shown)


app/configurator/components/field.tsx
```diff
@@ -702,36 +733,23 @@ export const ChartFieldField = ({
...
-            </span>
-          ) : (
-            <span>{label}</span>
-          )}
-          {fetching ? (
-            <CircularProgress size={12} className={classes.loadingIndicator} />
-          ) : null}
-        </>
+        <FieldLabel
+          isOptional={optional || false}
```
bprusinowski, line 749: Nit: not sure if we need `|| false`?

bprusinowski (approved):
LGTM 👏

app/configurator/components/field.tsx
ptbrowne: Ah yes you are right :)

app/locales/de/messages.po
```diff
@@ -13,24 +13,24 @@ msgstr ""
 "Language-Team: \n"
 "Plural-Forms: \n"
 
-#: app/configurator/components/chart-configurator.tsx:701
+#: app/configurator/components/chart-configurator.tsx:713
 msgid "Add filter"
 msgstr "Filter hinzufügen"
```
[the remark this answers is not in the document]
ptbrowne: Done in 91f376e0

Issue #986, opened 2023-03-01
schema:position not used to order chart legends


FabianCretton:
**Describe the bug**
An ordinal dimension is used to display a map, the schema:position should be used to order the legend under the chart

**To Reproduce**
1. Visualize INT or TEST
2. Dataset "waldbrand" which is a draft
3. Choose the map view
4. The labels under the map are ordered alphabetically, and not according to schema:position
5. Seems the same for the table view

**Screenshots or video**
![image](https://user-images.githubusercontent.com/6167669/222086723-aadd9be4-d991-48e5-9be4-a4fc2d319ad9.png)

![image](https://user-images.githubusercontent.com/6167669/222087036-02499dd7-1764-4e8d-b0b4-80894a264cc9.png)

![image](https://user-images.githubusercontent.com/6167669/222087217-a63fdd4b-c5ae-4055-ad85-b7ef0f79c38b.png)

**Environment (please complete the following information):**

- Visualize environment and version: INT v3.19.7
- Browser and version: FireFox 110

Rdataflow:
@ptbrowne table preview ordered by ordinal measure still behaves as screenshot nr 3

ptbrowne:
I extracted in a new issue: https://github.com/visualize-admin/visualization-tool/issues/988

PR #987, opened 2023-03-01
Sorting of map legend items

ptbrowne:
The categorical values in map legend were not sorted by position. Fixes Fixes https://github.com/visualize-admin/visualization-tool/issues/986

- refactor: Use generic so the caller does not have to do as
- feat: Sort map categorical legend values by position
- fix: Prevent error flash on map


bprusinowski (approved):
LGTM 🎉

Issue #988, opened 2023-03-01
Table sorting by ordinal measure


ptbrowne:
**Describe the bug**
When an ordinal dimension is used to order the table view of a map view, the schema:position should be used to order the values, as opposed to the label.

**To Reproduce**
1. Visualize INT or TEST
2. Dataset "waldbrand" which is a draft
3. Choose the map view, switch to map view
4. Sort by ordinal dimension
5. The values are ordered alphabetically, and not according to schema:position

![image](https://user-images.githubusercontent.com/6167669/222087217-a63fdd4b-c5ae-4055-ad85-b7ef0f79c38b.png)

**Environment (please complete the following information):**

- Visualize environment and version: INT v3.19.7
- Browser and version: FireFox 110

_Originally posted by @fabiancretton in https://github.com/visualize-admin/visualization-tool/issues/986#issuecomment-1449885853_

Rdataflow:
@ptbrowne to be precise: its not the table view  but the map view switched to the table view 😉

ptbrowne:
Thanks, edited :)

Rdataflow:
@ptbrowne yet another table view with the same Order issue is here: https://test.visualize.admin.ch/browse?previous=%7B%22order%22%3A%22SCORE%22%2C%22search%22%3A%22Waldbrand%22%2C%22includeDrafts%22%3Atrue%7D&dataset=https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fgefahren-waldbrand-warnung%2F1&dataSource=Int

PR #989, opened 2023-03-01
fix: Sorting table by ordinal value works

ptbrowne:
Sorting table by ordinal value works.

Fix https://github.com/visualize-admin/visualization-tool/issues/988


Rdataflow:
👍️

bprusinowski (approved):
💯

Issue #991, opened 2023-03-02
"Use abbreviations" option -> the languages are mixed-up 


FabianCretton:
**Describe the bug**
When "Use abbreviations" is on, the displayed labels (schema:alternateName) have mixed-up languages
This happens at least:
- under the map
- In metadata display

**To Reproduce**

1. Open the "Forest fire prevention measures" dataset on INT (draft)
https://int.visualize.admin.ch/en/create/diqYrvtc56uT?dataSource=Int
2. Map chart
3. Click on "Chart Options" -> Areas "Warning Region"
4. Under 'Color "Measure"' -> check Use abbreviations

And also:
1. Click on 'metadata', then 'data' tab, then 'measure' dimension

**Screenshots or video**
See the mixed-up languages:
![image](https://user-images.githubusercontent.com/6167669/222355271-9e71cb57-f3ce-46c5-8ca9-3fc5ad160346.png)

![image](https://user-images.githubusercontent.com/6167669/222355305-2f927f21-d586-4768-ac8f-07ae8ae36748.png)

**Environment (please complete the following information):**

- Visualize environment and version: INT v3.19.7 ([df9cafa]

PR #994, opened 2023-03-02
Abbreviations

ptbrowne:
Since abbreviations were not filtered by locale, in cases were abbreviations were localized, we could have multiple different locales competing.

Fix https://github.com/visualize-admin/visualization-tool/issues/991


bprusinowski (approved):
💯

PR #999, opened 2023-03-14
feat: When computing bbox, do not consider points without observations

ptbrowne:
Draft.

Questions:

Should the behavior be always true ?
It seems it should only be there when opening the map since after that
it can be confusing to always have the map moving.


bprusinowski (approved):
LGTM!

Issue #978, opened 2023-02-24
Strange behaviour when sorting by ID on the x-axis


ortnever:
I add another example related to the issue #976 .
In this example, sorting by ID is correct in the filter but not in the chart. I can't find any logic in the sorting in the chart and strangely it works when using the abbreviations.

- Edit this chart https://int.visualize.admin.ch/en/v/Si4Rjizp0B3n (it takes more than 1 minute until the chart is displayed)
- Edit the horizontal axis
- Select Dimension “BFSIdentifier”
- Select Sort by “Automatic”
- the order of cantons in the chart does not correspond to the ID. Strangely, when abbreviations are used, the sorting is correct. 

![image](https://user-images.githubusercontent.com/79834090/221173460-fb2cc5ad-f842-4219-ade5-7fa8734c7799.png)

![image](https://user-images.githubusercontent.com/79834090/221173503-d410745e-5c6e-426e-9cb2-d6d9ba3b07fe.png)

Environment : Visualize INT, source : INT, v3.19.5 ([0481cfa]
Browser : Edge

ptbrowne:
Fixed by https://github.com/visualize-admin/visualization-tool/commit/7eca90ea1260a727aad2fbfb415bd760d3e8b046, it's deploying on int

ortnever:
Hi @RdataFlow, the example above works well now on INT but it seems to me that there are some problems by Sorting by ID for Dimension with several parallel levels like Dimension "Region" in Datset  "LFI: Alle Zielgrössen nach Hoch-/Tieflagen".

https://int.visualize.admin.ch/de/v/rOJRafv5qRFd 

Could you have a look to your dataset.  The filter options for the dimension are quite complex.
![image](https://user-images.githubusercontent.com/79834090/224667154-7ea7b67e-efd0-4b60-9b42-9f15464c2e8b.png)

Rdataflow:
@ortnever it's really odd.
same here without switzerland https://int.visualize.admin.ch/de/v/w-SsK1P9T1s7 

@ptbrowne can you reopen this issue and fix it?
cc @zellersabine

ortnever:
Salut @ptbrowne : Could you also check the sorting by ID (Automatic sorting, by default) at the 2nd hierarchical level (municipalities). I don't understand the logic used. In any case it doesn't seem to be by ID.

ptbrowne:
[...]

PR #1006, opened 2023-03-30
Fix: sort values with equally labelled values inside the hierarchy

ptbrowne:
Previously, when sorting values, we have to build a dimension value
index, indexed by label, since the data that we receive does not contain
the value iris, but it contains the value labels.

The problem occurs
when two dimension values have the same label, but not the same IRI,
for example in the case of hierarchy, where two different regions
are named the same but are located in two different parts of the tree.

Previously, the sorting would end up wrong, since we would make a mistake
when getting the dimension value (finding one dimension value from the
index, when we should have found the other).

To remediate the problem, when there is a filter on the dimension, we
pass it to the function creating the sorting, so that only values
present in the filter are used. This way, we lower the risk of using
two values with the same label.

The problem would still be there if the chart creator selects two
values with the same label, to be displayed at the same time on the
same chart. For example if someone decides to display at the same time
Bern the canton, and Bern the city. Then we would have trouble indexing
correctly. In this case, the person reading the chart would also have
trouble though.

Fix https://github.com/visualize-admin/visualization-tool/issues/978

### DIFF
--- a/app/.gitignore
+++ b/app/.gitignore
@@ -2,3 +2,5 @@ node_modules/
 /test-results/
 /playwright-report/
 /playwright/.cache/
+
+**/tsconfig.tsbuildinfo
--- a/app/charts/area/areas-state.tsx
+++ b/app/charts/area/areas-state.tsx
@@ -203,6 +203,9 @@ const useAreasState = (
   const segmentSortingType = segmentSorting?.sortingType;
   const segmentSortingOrder = segmentSorting?.sortingOrder;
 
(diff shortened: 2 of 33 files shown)


bprusinowski (approved):
Code looks good! But I wasn't able to test the change on Vercel, as there seems to be some problem with the deployment 👀

ptbrowne:
👍 I did not have time to fix the build before the German course 😅 It's fixed now, thanks !

bprusinowski:
LGTM 👌

PR #1013, opened 2023-04-04
fix: Color mapping for tables

### LINKED ISSUE #872: Attribution of colours by categories for a dimension "Year"
The attribution of colours in the example below is strange. How does it work?

1. Copy and Edit  " https://int.visualize.admin.ch/en/v/Lnk_MB0Bx75c
2. Edit "Jahr der Vergütung"
3. Only 2 colors from the palette are used, while the values range from 2014 to 2021

**Expected behavior**
As many colours should be used as there are different years, as long as colours are still available in the palette. 

**Screenshots or video**
If applicable, add screenshots or a short video to help put your problem into a context.

- Visualize environment and version: INT v3.12.0
- Browser and version: Edge

**Additional context**

bprusinowski:
Fixes #872.

There was an issue with subsetting the chart config (lodash's `get`) when using strings for table fields, which are iris and not "regular" strings, like `segment` or `x`. This PR fixes that by using an array instead of string.

Also, there was a problem with refreshing color mappings for table, as we were expecting a `segment` field to be passed to `updateColorMapping`, while we also can pass table fields.

### To test
1. Go to https://visualization-tool-7acz0ufkc-ixt1.vercel.app
2. Look for Einmalvergütung für Photovoltaikanlagen dataset (INT).
3. Create a chart.
4. Switch to a table chart.
5. Change Standort der Anlage's column type to Category.
6. See that it's possible to map the colors individually to each value.
7. See that it's possible to refresh the colors.
8. Switch to Jahr column.
9. Change type to category.
10. See that more than two colors are mapped.

~~We also fetch all values for `TemporalDimensions` with resolution of Year, so the color mapping inside a table for such dimensions won't just map two colors. I feel like it's not an ideal solution – other way would be to add a `forceLoadAll` property to GQL query for dimension values and load them just inside the panel where we modify column options for table.~~

But then the question still remains, should we be able to modify each color individually like for multi-filter panel? Should the year be colored only with sequential color palettes, as it's kinda numerical? If we would like to go with the "color-each-value-individually" approach, there would be a need to modify much more logic, as right now multi-filter panel is only supposed to work with dimensions that can be multi-filtered, while for year we have different logic (slider).

What do you think @ptbrowne?

### DIFF
--- a/app/configurator/components/filters.tsx
+++ b/app/configurator/components/filters.tsx
@@ -1,20 +1,20 @@
 import { t, Trans } from "@lingui/macro";
 import {
+  Accordion,
+  AccordionDetails,
+  AccordionSummary,
   Box,
   Button,
   ClickAwayListener,
-  Typography,
   Divider,
-  Theme,
-  IconButton,
-  Tooltip,
-  Accordion,
-  AccordionSummary,
-  AccordionDetails,
-  InputAdornment,
-  Input,
   FormControlLabel,
+  IconButton,
+  Input,
+  InputAdornment,
   Switch,
+  Theme,
+  Tooltip,
+  Typography,
 } from "@mui/material";
 import { styled } from "@mui/material/styles";
 import { makeStyles } from "@mui/styles";
@@ -25,7 +25,7 @@ import keyBy from "lodash/keyBy";
 import orderBy from "lodash/orderBy";
 import sortBy from "lodash/sortBy";
 import uniqBy from "lodash/uniqBy";
-import React, {
+import {
   forwardRef,
   MouseEventHandler,
   MutableRefObject,
@@ -165,9 +165,9 @@ const getColorConfig = (
     return;
   }
 
-  const path = `${config.activeField}${
-    colorConfigPath ? `.${colorConfigPath}` : ""
-  }`;
+  const path = colorConfigPath
+    ? [config.activeField, colorConfigPath]
+    : [config.activeField];
 
   return get(config.chartConfig.fields, path) as
     | GenericSegmentField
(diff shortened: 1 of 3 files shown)


ptbrowne:
LGTM for the technical fixes concerning lodash gets, and updateColorMapping. Thanks for the explanation 💯 

Concerning the functionality:

- I agree that sequential palette would be more meaningful, but it would also be harder to read as shades would be hard to distinguish. What I like here is that it is very easy to distinguish each year in the table. It is only useful though if the dimension has a low cardinality (no more than 10 values), which is potentially not the case for temporal yearly values.
- What I wonder now if coloring a numerical value like this in a table, is helpful to the end user. @ortnever Will you produce tables with years formatted in color like that ? Then would you like to be able to control the colors ?
- I think here we would need more thoughts and potential usecases. In the meantime, I think coloring a numerical value as a "category" should not be offered to the user if the dimension is a temporal one. Only if it is a nominal or ordinal dimension.

cc @adintegra What do you think ?

ortnever:
Thank you Patrick and Bartosz for your feed-back concerning this issue. 
I opened this issue because I did'nt understand the logic used to apply the colors to this temporal dimension but I realize that the expected behavior I described in the issue #872 doesn't make sense for years or dates.
@Rdataflow : What do you think? Do you agree that coloring a numerical value as a "category" should not be offered to the user if the dimension is a temporal one?  Or do you have a use case ?

ortnever:
I discussed this issue with Thomas Bettler. He also doesn't have a use case where colors should be assigned to temporal values. You can keep it as it is.

ptbrowne:
So @bprusinowski, here I guess the next steps are to separate the loading of all the values from the rest of the PR, and add the "disabling" of the "categories" option for temporal values ?

bprusinowski:
@ptbrowne yes, I removed fetching all values for temporal dimensions with Year resolution and disabled category column style for such dimensions 👍

ptbrowne (approved):

PR #1014, opened 2023-04-04
fix: Redirect to versioned cube

### LINKED ISSUE #939: useRedirectToVersionedCube uses wrong dataSource on page load
It seems like useRedirectToVersionedCube uses the default dataSource prod on int.lindas.admin.ch when reaching https://int.visualize.admin.ch/en/browse?dataset=https%3A%2F%2Fpolitics.ld.admin.ch%2Ffch%2Fpolitical-correspondence&dataSource=Int instead of picking the Int dataSource indicated in the URL.

It is visible from the network tab where we can see that the request to Lindas is made to lindas.admin.ch

It leads to the failure of loading the cube.

I cannot yet reproduce on localhost.

bprusinowski:
Fixes #939.

We were initializing configurator state with default data source – this PR fixes that by correctly initializing the state with correct data source.

### To test
1. Go to `/en/browse?dataset=https%3A%2F%2Fenergy.ld.admin.ch%2Fsfoe%2Fbfe_ogd102_fuellstand_gasspeicher&dataSource=Int `(see that no cube version is specified in the URL).
2. See that the cube loads correctly (version 6).

### DIFF
--- a/app/configurator/components/use-redirect-to-versioned-cube.tsx
+++ b/app/configurator/components/use-redirect-to-versioned-cube.tsx
@@ -50,6 +50,7 @@ export const useRedirectToVersionedCube = ({
         router.replace({
           pathname: `/${locale}/browse`,
           query: {
+            ...router.query,
             dataset: resp.iri,
           },
         });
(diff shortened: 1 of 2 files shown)


ptbrowne (approved):

ptbrowne:
Good find !

Issue #1002, opened 2023-03-21
poor performance to load data (graphql)


Rdataflow:
**Describe the bug**
the data loading performance with graphql is really slow (typically 12secs, in some cases up to 90 secs) compared to the sparql query retrieving the same table (<0.1s)

A clear and concise description of what the bug is. If it seems connected to some data problem (missing values, wrong parsing) please first check the cube in Cube Checker ([int](https://int.visualize.admin.ch/_cube-checker), [prod](https://visualize.admin.ch/_cube-checker)) to see if everything is fine there.

**To Reproduce**

1. Go to https://test.visualize.admin.ch/browse?dataset=https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Ffab_foreign_trade_test_1%2F1&dataSource=Test
2. See loading takes long
3. Open SPARQL-Query
4. Compare loading is really quick

**Expected behavior**
Load the preview or chart equally quick

**Screenshots or video**
loading times in graphql
![image](https://user-images.githubusercontent.com/26090254/226612520-2d943d2c-ccb8-4135-83dd-c6e9f7baef19.png)

loading times in sparql
![image](https://user-images.githubusercontent.com/26090254/226613115-644d51c9-905d-42cd-abe6-febd83ad52f6.png)

**Environment (please complete the following information):**

- Visualize environment and version: [e.g. TEST v3.19.12]

**Additional context**
- graphql query 
```shell
curl 'https://test.visualize.admin.ch/api/graphql' \
  -H 'Referer: https://test.visualize.admin.ch/browse?previous=%7B%22includeDrafts%22%3Atrue%2C%22order%22%3A%22SCORE%22%2C%22search%22%3A%22test%22%7D&dataset=https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Ffab_foreign_trade_test_1%2F1&dataSource=Test' \
  -H 'content-type: application/json' \
  -H 'x-visualize-cache-control: no-cache' \
[...]

PR #1016, opened 2023-04-05
perf: Only load hierarchies where necessary

bprusinowski:
Contributes to fixing #1002.

Previously, we loaded hierarchies every time we depended on `dimensionMetadata` fragment. However, I believe it's not necessary to load hierarchies everywhere, as for example, they don't seem to be used in places that use `DataCubeObservations` query. When we don't fetch them there, the loading times of the chart inside the editor & published mode can be improved for datasets that use hierarchies (in the example in #1002, the published chart loads in 1.5s instead of ~10s).

This PR excludes loading of hierarchies across multiple queries if they are not required further down in the app.

### How to test
1. Go to [this published chart](https://visualization-tool-git-perf-published-chart-ixt1.vercel.app/en/v/RF4kn1KqqdDN?dataSource=Test) from this PR's deployment preview
2. Go to the [same chart](https://test.visualize.admin.ch/en/v/tVsVHOmiGU3h?dataSource=Test) but made using test.visualize.admin.ch and see the time difference in data loading.

### DIFF
--- a/app/charts/column/columns-state.tsx
+++ b/app/charts/column/columns-state.tsx
@@ -48,8 +48,8 @@ import {
 } from "@/configurator/components/ui-helpers";
 import { isTemporalDimension, Observation } from "@/domain/data";
 import {
-  useFormatNumber,
   formatNumberWithUnit,
+  useFormatNumber,
   useTimeFormatUnit,
 } from "@/formatters";
 import {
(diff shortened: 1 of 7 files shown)


ptbrowne (approved):
LGTM !

PR #1025, opened 2023-04-28
fix: Temporal values in X axis in column charts

### LINKED ISSUE #1024: Temporal values not being displayed correctly on x-axis on INT
When displaying temporal values on the x-axis on INT the values are (at least for one value) not shown correctly. Instead of the the expected values (e.g. year) _**0NaN**_ is shown. Also for some cubes not all the possible temporal values are shown in the bar chart and neither are only the valid values shown when in this case year is selected as a filter option (all years show up for selection instead).

[Heavy Metal Soil Contamination](https://int.visualize.admin.ch/create/QvP0Mlp8zd8H?dataSource=Int)
Tempaxis_bug_20230425

[Another example for the airpollution along the A2 and A13](https://int.visualize.admin.ch/fr/v/yjzYpOegXJIJ)

On PROD I did not encounter this issue

bprusinowski:
Fixes #1024.

We were creating the lookups in `useMaybeAbbreviations` by using dimension values. While it's not a problem for most dimensions, we only fetch min and max values for temporal dimensions which leads to problems described in #1024.

This PR fixes that by passing "fake" dimension values (based on the ones that appear in the data) for column charts (X axis).

### DIFF
--- a/app/charts/shared/chart-helpers.tsx
+++ b/app/charts/shared/chart-helpers.tsx
@@ -23,11 +23,7 @@ import {
   QueryFilters,
 } from "@/configurator/config-types";
 import { FIELD_VALUE_NONE } from "@/configurator/constants";
-import {
-  DimensionValue,
-  isTemporalDimension,
-  Observation,
-} from "@/domain/data";
+import { isTemporalDimension, Observation } from "@/domain/data";
 import { truthy } from "@/domain/types";
 import { DimensionMetadataFragment } from "@/graphql/query-hooks";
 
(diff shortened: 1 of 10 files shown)


app/charts/shared/chart-helpers.tsx
```diff
@@ -411,3 +416,23 @@ export const useImputationNeeded = ({
...
+    if (isTemporalDimension(dimension)) {
+      const dates = data.map((d) => d[dimension.iri] as string);
+      return uniq(dates).map((d) => ({ label: d, value: d }));
+    } else {
+      return dimension.values;
+    }
+  }, [dimension, data]);
+
+  return maybeTemporalDimensionValues;
+};
```
ptbrowne, line 438: I think it would be better to separate the getMaybeTemporalDimensionValues from the useMemo. Then it is easier to test (no need for renderHook, only a normal function test) and concerns (react memoizing and business logic) are not mixed. From the calling point, it is also simpler to grok what's happening.

ptbrowne (approved):
Nit: I think the useMaybeTemporalValues was coded as a hook to be similar to useMaybeAbbreviations but I think actually here that using a normal business function would be easier. 

LGTM !

Issue #1028, opened 2023-05-09
Slow performance on loading published charts


sosiology:
_Issue Shared by Thomas per Email on 09.05.2023_

**Describe the bug**
Slow performance on loading published charts with LFI dataset
 
**To Reproduce**
 
1. Open https://int.visualize.admin.ch/v/hKkUyBpGpWPl?dataSource=Int
2. Observe long loading times (up to 12s)
 
**Expected behavior**
Shorter loading time (preferably <1s) given that the filter values are available from slug (postgres) and
SPARQL query is really quick (i.e. 0.2s see https://s.zazuko.com/PCAa4k)
 
**Environment (please complete the following information):**
INT v3.19.14 ([e804d90](https://github.com/visualize-admin/visualization-tool/commit/e804d90b07922d0ffda7fd2ebedb50bcf4038b78))
 
**Additional context**
![image](https://github.com/visualize-admin/visualization-tool/assets/26090254/628739dc-14f0-4f66-b959-f37c644c786d)

sosiology:
Hi @bprusinowski 
This is a new Issue from Thomas, that has the highest prio at the moment. Could you have a look at it this week please?
cc: @adintegra

adintegra:
For reference: (Initial) retrieval of the test URL now returns in ~3.5s

<img width="865" alt="image" src="https://github.com/visualize-admin/visualization-tool/assets/6533245/51c1313c-4cc8-427a-897f-d3d316d7a6d5">

PR #1031, opened 2023-05-10
perf: Only fetch necessary components

bprusinowski:
Contributes towards fixing #1028.

## Goals / Scope

This PR aims to improve data loading performance for published charts.

## Description

The main change from this PR is about fetching only the dimensions that are actually used in the chart (for the published chart) to reduce the need to fetch values for them, which can lead to unnecessary fetch of dozens of dimensions and their values for some datasets. It means that we only show the dimensions that are actually used in chart in the metadata panel, table preview and data download. I think that from the end user perspective it makes sense, as what we see on the chart is reflected in all the other places. Let me know about your opinions here @RDataflow @ptbrowne.

## Comments

Additionally to the change described above, I also tried to understand what causes the time difference described in #1028. I ran 20 tests focusing on observations fetching using the chart from https://int.visualize.admin.ch/v/hKkUyBpGpWPl?dataSource=Int. [The function to fetch the observations](https://github.com/visualize-admin/visualization-tool/blob/e231b68e7b0e075ee3d1daa4eb173590bdd98e6d/app/rdf/queries.ts#L462-L566) consists of five main parts, which were measured and visualised below. The average time to run the function was 1.045s.

Screenshot 2023-05-10 at 17 06 44

As can be seen, the fetching of the observations took only 0.145s on average, compared to 0.930s to build the query itself (filters, dimensions, etc). I hope that this at least partially explains why it takes longer to fetch data from LINDAS using Visualize than direct calls to the API.

### DIFF
--- a/app/charts/area/chart-area.tsx
+++ b/app/charts/area/chart-area.tsx
@@ -2,9 +2,11 @@ import { memo } from "react";
 
 import { Areas } from "@/charts/area/areas";
 import { AreaChart } from "@/charts/area/areas-state";
+import { ChartLoadingWrapper } from "@/charts/chart-loading-wrapper";
 import { AxisHeightLinear } from "@/charts/shared/axis-height-linear";
 import { AxisTime, AxisTimeDomain } from "@/charts/shared/axis-width-time";
 import { BrushTime } from "@/charts/shared/brush";
+import { getChartConfigComponentIris } from "@/charts/shared/chart-helpers";
 import { ChartContainer, ChartSvg } from "@/charts/shared/containers";
 import { Ruler } from "@/charts/shared/interaction/ruler";
 import { Tooltip } from "@/charts/shared/interaction/tooltip";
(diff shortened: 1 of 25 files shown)


bprusinowski:
Thanks for review @ptbrowne! I'll add the test.

It's true that we might miss the cache for some observations queries, but only in edit mode, in published mode we should improve the loading times for a little bit. For the performance of creating the View, I discovered that it took so long because the dimensions are sorted each time it's created and I am not convinced they need to be sorted; I've opened an issue to [clarify / fix this](https://github.com/zazuko/rdf-cube-view-query/issues/95).

@Rdataflow the benchmark focuses only on the logic to fetch the observations – it's a part of bigger story, as we also need to fetch dimensions and in some cases, hierarchies. Nonetheless, the performance improvement for the particular chart that was used for testing would improve for 0.5s just by disabling sorting of the dimensions.

The benefit of changes in this PR should also reduce the time of loading observations by ~0.5s. I would say that in total, with disabling the sorting we should gain around one second of improved performance for the chart used for testing.

I'll dig into the performance of other data-fetching function in other PR 👍

PR #1041, opened 2023-05-15
perf: Do not sort dimensions in View

bprusinowski:
Until the appropriate change is implemented in rdf-cube-view-query library, we manually override sorting of dimensions to improve performance.

For more context see the description and discussion in #1031.

cc: @Rdataflow


ptbrowne (approved):
Thanks for the investigation ! 👏

PR #1042, opened 2023-05-16
feat: Hide Animation behind feature flag & fix column animations

bprusinowski:
- Animation field is now hidden behind a feature flag
- Columns now animate correctly during a resize event


bprusinowski:
@ptbrowne I am going to merge before the release to INT, feel free to add comments and I will tackle them afterwards 👍

PR #1049, opened 2023-06-01
chore: Update rdf-cube-view-query

bprusinowski:
As the [new version of rdf-cube-view-query library](https://github.com/zazuko/rdf-cube-view-query/releases/tag/v1.12.0) has been released, there is no need to manually disable sorting of the columns anymore.


ptbrowne (approved):
🎉

PR #1055, opened 2023-06-05
fix: Sorting issues

### LINKED ISSUE #1054: Sorting in reverse alphabetical order does not affect the legend in lines plot
Sorting in reverse alphabetical order does not affect the legend in lines plot. The same works in the bar plot without any issue. Just the button press from **A -> Z** to **Z -> A** does not affect the legend to display Jura first instead of Zurich (which would be index number 1)

[Photovoltaiv](https://int.visualize.admin.ch/create/KBkqret2PzUU?dataSource=Prod)
Sorting Bug Lines

bprusinowski:
Closes #1054.

### DIFF
--- a/app/charts/line/lines-state.tsx
+++ b/app/charts/line/lines-state.tsx
@@ -44,7 +44,10 @@ import { DimensionMetadataFragment } from "@/graphql/query-hooks";
 import { getPalette } from "@/palettes";
 import { sortByIndex } from "@/utils/array";
 import { estimateTextWidth } from "@/utils/estimate-text-width";
-import { makeDimensionValueSorters } from "@/utils/sorting-values";
+import {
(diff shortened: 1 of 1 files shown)


ptbrowne (approved):
Code changes all look OK to me ! Thanks @bprusinowski ! Do you think we could add a test for the nominal values sorting ?

bprusinowski:
@ptbrowne I've added two tests for nominal and measure dimensions 👍

sosiology:
> > @ptbrowne do you remember why such filtering was introduced in the first place?
> 
> Sadly, I do not remember why there was a sorting on identifier or position in the first place.

Maybe this @bprusinowski https://github.com/visualize-admin/visualization-tool/pull/821

PR #1059, opened 2023-06-06
fix: Interactive color legend

### LINKED ISSUE #1058: Interactive Filter Error
**Describe the bug**
Nach Auswahl von Interaktiven Filter unter "FARBE" in Gestapeltem Modus. Können diese Nach der Veröffentlichung nicht korrekt an und abgewählt werden. Bei nicht Gruppierten Säulen funktioniert das ganze.

Error:
https://int.visualize.admin.ch/v/kp_yVMjroNFA?dataSource=Prod

Expected: 
https://int.visualize.admin.ch/v/0r-luSVGOc6f?dataSource=Int

Hier funktioniert es gar nicht:
https://int.visualize.admin.ch/v/1U5a017M__2x?dataSource=Int

**To Reproduce**

1. Go to 'Säulendiagramm -> Farbe -> Parameter -> Gestapelt ->Interaktiv'
2. Click on 'Nach Veröffentlichung -> Einen Filter abwählen danach wieder anwählen

**Expected behavior**
Filter erscheinen und verschwinden nach auswählen.

**Screenshots or video**
filter error

- Visualize environment and version: [v3.19.18]
- Browser and version [Firefox]

bprusinowski:
Fixes #1058.

This PR fixes two problems with interactive color legend issues.

1. Stacked bar chart was using all segments defined in the data to construct series (coordinates to draw the chart) – when one segment was de-selected using interactive color legend, it would introduce NaNs and thus break the chart. Now we use plottable segments to create series.
2. Interactive color legend not affecting the chart – as we store abbreviations / labels inside interactive filters state, we need to also get segment abbreviation / label when applying interactive filters to the data (previously we would try to access segment value).

### DIFF
--- a/app/charts/column/columns-stacked-state.tsx
+++ b/app/charts/column/columns-stacked-state.tsx
@@ -32,9 +32,9 @@ import {
 } from "@/charts/column/constants";
 import {
   getLabelWithUnit,
+  getMaybeTemporalDimensionValues,
   getWideData,
   useDataAfterInteractiveFilters,
-  getMaybeTemporalDimensionValues,
   useOptionalNumericVariable,
   usePlottableData,
   useTemporalVariable,
@@ -244,9 +244,12 @@ const useColumnsStackedState = (
   const segmentFilter = segmentDimension?.iri
     ? chartConfig.filters[segmentDimension?.iri]
     : undefined;
-  const segments = useMemo(() => {
-    const uniqueSegments = Array.from(
-      new Set(plottableSortedData.map((d) => getSegment(d)))
+  const { segments, plottableSegments } = useMemo(() => {
+    const allUniqueSegments = Array.from(
+      new Set(plottableSortedData.map(getSegment))
+    );
+    const allPlottableSegments = Array.from(
+      new Set(preparedData.map(getSegment))
     );
 
     const sorting = fields?.segment?.sorting;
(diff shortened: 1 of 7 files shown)


app/charts/column/columns-stacked-state.tsx
```diff
@@ -257,9 +260,21 @@ const useColumnsStackedState = (
...
+      sorters,
+      getSortingOrders(sorters, sorting)
+    );
+
+    return {
+      segments: allSegments,
+      plottableSegments: allSegments.filter((d) =>
+        allPlottableSegments.includes(d)
+      ),
+    };
```
ptbrowne, line 274: I am under the impression that the `segments` returned here is only used for the order of the tooltip values. If so, I think the order could only rely on plottableSegments, and this memo could only return plottableSegments.

bprusinowski: I also thought that this is the case, but we also use them in color domain, to display the legend – if we just return `plottableSegments`, we would remove a color legend item permanently once de-selected (and completely remove the chart is we de-select them all 😅)

bprusinowski:
Thanks for reviewing @ptbrowne!

app/charts/column/columns-stacked-state.tsx
ptbrowne: Ahah, understood 👍

PR #1070, opened 2023-06-19
perf: Only send GQL extensions to the client when needed

bprusinowski:
When investigating the performance of SPARQL queries, I noticed that the payload sent to the client in the debug mode (query, time of execution) is sent always, even when debug mode is off.

As this additional data can account for significant amount of additional payload size that's not always needed, I restricted this behavior to only occur when the debug mode is explicitly turned on.

Change in payload size using default chart when editing the dataset (with server-side cache disabled):
- Bathing water quality: 507kB vs 273kB
- Forest fire danger: 378kB vs 252kB

### DIFF
--- a/app/graphql/client.tsx
+++ b/app/graphql/client.tsx
@@ -9,10 +9,16 @@ export const client = createClient({
   url: GRAPHQL_ENDPOINT,
   exchanges: [...devtoolsExchanges, ...defaultExchanges],
   fetchOptions: {
-    headers: {
-      "x-visualize-cache-control": flag("server-side-cache.disable")
-        ? "no-cache"
-        : "",
-    },
+    headers: getHeaders(),
   },
 });
+
+function getHeaders() {
+  const debug = flag("debug");
+  const disableCache = flag("server-side-cache.disable");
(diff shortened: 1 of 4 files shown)


bprusinowski:
cc @adintegra, not sure if this should be merged before the release on Wednesday, but might be worth to, as the debug information is also attached on PROD by default 👀

app/graphql/client.tsx
```diff
@@ -9,10 +9,34 @@ export const client = createClient({
...
+  if (debugMode) {
+    return {
+      "x-visualize-debug": "true",
+    };
+  } else {
+    return {
+      "x-visualize-debug": "",
+    };
+  }
+}
```
ptbrowne, line 42: Nit: Would find it simpler to have only one function getHeaders doing both things.

ptbrowne (approved):
LGTM, thanks Bartosz, why didn't I do that :/

app/graphql/client.tsx
bprusinowski: Thanks, done!

PR #1071, opened 2023-06-19
perf: Improve code splitting

bprusinowski:
- Do not import `ContentMDXProvider` in `_app.tsx`, to prevent adding ~2.8MB in payload size to every page, but rather only use it where it's necessary

## Further ideas
Right now, if we want to render a simple column chart, we bundle all column chart related logic, including the grouped chart state, stacked chart state, code for rendering grouped & stacked charts, etc.

I believe that this could be avoided if we split the code in a manner that we treat these are different chart types and not include them in one function (ChartColumnsVisualization) that conditionally renders a given chart subtype. I think this would be a bigger change, as we would need to treat these chart subtypes as chart types, e.g. "column-simple", "column-grouped", "column-stacked". With this we wouldn't bundle things that are not necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ You can also check the [release page](https://github.com/visualize-admin/visuali
 
 - Performance
   - The debug data is now only sent along with the payload when the debug mode is enabled
+  - ContentMDXProvider is now only imported within the pages that actually need it instead of being included in the root of the app
 
 # [3.20.1] - 2023-06-19
 
--- a/app/components/content-mdx-provider.tsx
+++ b/app/components/content-mdx-provider.tsx
@@ -3,7 +3,7 @@ import { Box } from "@mui/material";
 import { ReactNode } from "react";
 
 import { ContentLayout, StaticContentLayout } from "@/components/layout";
-import { Intro, Tutorial, Examples, Contribute } from "@/homepage";
+import { Contribute, Examples, Intro, Tutorial } from "@/homepage";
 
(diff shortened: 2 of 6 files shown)


bprusinowski:
@ptbrowne I am going to merge, let me know if you have any thoughts on this :)

ptbrowne:
Thanks Bartosz, LGTM.

On the further notes, I am wondering why every chart is bundled when the `dynamic` tag is used (I see that in `app/components/common-chart.tsx`) ? It would have seemed to me that the components wrapped in dynamic would only bring their code when the component was rendered ?

bprusinowski:
Yes, that's the case – but the column chart that we import is basically this, https://github.com/visualize-admin/visualization-tool/blob/91b2d00e38067535f2b9b15bb6a645621a6a54dc/app/charts/column/chart-column.tsx#L95-L217 which means that we have three charts inside one function.

bprusinowski:
I think the problem is that we have one function that renders different chart subtype based on some conditions – if it was split inside three functions, that would prevent bundling everything. I tried to split that, but still there was a root check on chart type (`column`) + additional checks below it on which subtype to render, and that didn't work, that's why I think these would need to be distinguishable at the same level as other dynamic imports (chart type: area, line, ...) 🤔  I hope it makes sense 😅

bprusinowski:
Yup, I used `@next/bundle-analyzer` and found the issue with MDX provider 😅 But it's true that icons are the next big thing to look at 👍

PR #1075, opened 2023-06-20
fix: Hot reload of browse pages

bprusinowski:
This PR fixes refreshing the `/browse` page while using `yarn dev`.

Not sure about the root cause of the issue, but recently the app started to throw an error when refreshing the `/browse` page while developing the app.

It seems that it's somehow related to bundling the app.

Screenshot 2023-06-20 at 12 54 35


ptbrowne (commented):
I am not sure I understand the solution 😅 How did you debug this ?

bprusinowski:
By investigating the chunks, in this case the one throwing the error (`en/browse`). As in the screenshot, it seemed that the `DatasetBrowser` is imported too soon into the page? (second line in the file). I changed this to define it inside the page, and not import from the components folder and it fixed the issue. Do you have any idea why this happened?

ptbrowne:
I guess it could be a sign of circular dependency error between two modules. The fact that you moved one component to the other goes into this direction. During a circular dependency, one of the module is still undefined. I guess importing more carefully (for example I see that EditorConfiguratorStateProvider in dataset-browser is imported from @/src and @src/ imports DatasetBrowser so it seems there is something on going there.

A imports B
B imports A (A is undefined at this points since it has not exported yet)

bprusinowski:
I see, thanks for investigating! Is then the proposed solution fine, or should I try to fix it using more specific imports?

app/src/index.ts
```diff
@@ -8,11 +8,11 @@ export {
...
-export { LocaleProvider, useLocale } from "../locales/use-locale";
 export {
-  i18n,
   defaultLocale,
+  i18n,
   locales,
   parseLocaleString,
 } from "../locales/locales";
+export { LocaleProvider, useLocale } from "../locales/use-locale";
+export { DatasetBrowser } from "../pages/browse";
```
ptbrowne, line 18: Is this necessary ? I think we should try to limit the number of things we export from a big index module otherwise we increase the risk of importing src and having a circular dependency.

ptbrowne:
I am fine with the current solution as it works but I think we should find something to limit this kind of error. 

1. Madge can detect circular dependencies but trying it quickly  on the repo gives no error so I think it does not handle '@' paths well.
2. I wonder if we could have eslint import rules for this kind of thing like "component files should not import from pages files"

ptbrowne (approved):

PR #1077, opened 2023-06-21
fix: Argos tests

bprusinowski:
Update GitHub e2e action workflow to always run screenshot upload step.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -59,14 +59,6 @@ jobs:
             "state": "failure",
             "description": "E2E tests failed"
           }'
-      - name: Set argos upload branch on master branch
-        run: echo "ARGOS_BRANCH=master" >> $GITHUB_ENV
-        if: github.ref == 'refs/heads/main'
-      - name: Set argos upload branch on PR branches
-        run: echo "ARGOS_BRANCH=$(git name-rev --name-only $GITHUB_SHA | sed 's-remotes/origin/--')" >> $GITHUB_ENV
-        if: github.ref != 'refs/heads/main'
-      - name: Upload screenshots to Argos
-        env:
-          ARGOS_COMMIT: ${{ github.sha }}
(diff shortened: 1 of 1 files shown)


bprusinowski:
@ptbrowne it seems that I can't test this action until it's on `main` branch... I think there was a problem with the screenshot upload step, it didn't got to that point if a failure in e2e tests appeared before it. I've added an `if: always()` condition and changed the action to use the one suggested in the docs (https://argos-ci.com/docs/github-actions#complete-example)

.github/workflows/e2e.yml
```diff
@@ -59,14 +59,6 @@ jobs:
...
-        run: echo "ARGOS_BRANCH=$(git name-rev --name-only $GITHUB_SHA | sed 's-remotes/origin/--')" >> $GITHUB_ENV
-        if: github.ref != 'refs/heads/main'
-      - name: Upload screenshots to Argos
-        env:
-          ARGOS_COMMIT: ${{ github.sha }}
-          ARGOS_TOKEN: ${{ secrets.ARGOS_TOKEN }}
-        run: yarn run argos upload e2e-screenshots
+      - name: Upload screenshots to argos-ci.com
+        if: always()
+        run: npx @argos-ci/cli upload e2e-screenshots
```
ptbrowne, line 64: So now the argos CLI detects correctly the branch even in the case of PR branches ?

ptbrowne: I guess ? https://github.com/argos-ci/argos-javascript/blob/bad0f5b5bf1dc1672b05cf08a4fec5d011ac4b0a/packages/core/src/ci-environment/git.ts#LL19C31-L19C62

ptbrowne (approved):

.github/workflows/e2e.yml
bprusinowski: Yes, looks like!

Issue #1060, opened 2023-06-06
Make it more apparent that area chart requires missing data imputation


bprusinowski:
It should be more visible (e.g. via an alert box above the chart) that an area charts needs missing data imputation. This only applies to the editor mode.

For context, see #1056.

cc @AFoletti

PR #1089, opened 2023-06-29
fix: Stacked column chart with missing values

### LINKED ISSUE #1088: Erratic chart behaviour
**Describe the bug**
A cube **without any empty values** in the measure dimension behaves a little erratically when switching from one plotted key dimension to another.
See the following video, and pay attention at how the "_Endverbrauch - Total_" plot changes from time to time.

https://github.com/visualize-admin/visualization-tool/assets/691243/2e1dc469-1c34-4d11-af81-4880bd7d7a1a

**To Reproduce**

1. Go to visualize INT
2. Search for the "(Long) Schweizerische Gesamtenergiestatistik" cube (from SFOE) and try to create a visualization for it
3. Pick the column graph, and activate the coloring by "Energy carrier"
4. Change multiple times the displayed value for the "Rubrik" key dimension. Observe how the resulting plot is not constant

**Expected behavior**
Same data, same plot, same configuration == same results, every time!

**Screenshots or video**
See above

- Visualize environment and version: (https://github.com/visualize-admin/visualization-tool/commit/91b2d00e38067535f2b9b15bb6a645621a6a54dc)

**Additional context**
A couple of probably important points:

1. V13 of this cube (the last version where the issue described here is present) does not have any missing values for the measure dimension. V12 of the cube on the other hands has missing values (just so you know you have this version at your disposal for tests should you need it). I would have suspected another "missing imputation" issue, but since I see the problem with "always full" data it should be something else...
2. The cube uses **a hierarchy** on the Energy Carrier key dimension used for colorizing. Maybe has an influence?
3. This is the very same **high prio cube** as earlier today. I would appreciate a quick signal from your side @bprusinowski 

Tagging @Rdataflow and @aresssera FYI

bprusinowski:
Fixes #1088.

As we use d3's `stacked` method in stacked bar chart, there can be situations similar to area chart, where we need imputation when some data point does not have some value defined in previous / following data point. We also need to impute such values, to make sure there are no NaNs returned when computing the coordinates for the chart.

I think it doesn't make sense to provide imputation options to users, as we can safely treat non-existing, imputed values as 0, as they do not have influence of how the chart is drawn, as in area chart (lines connecting the data points), and we do not show such values in the tooltip anyway.

### DIFF
--- a/app/charts/column/columns-stacked-state.tsx
+++ b/app/charts/column/columns-stacked-state.tsx
@@ -212,25 +212,6 @@ const useColumnsStackedState = (
     [preparedData, getX]
   );
 
-  const chartWideData = getWideData({
-    dataGroupedByX: preparedDataGroupedByX,
-    xKey,
-    getY,
-    getSegment,
-  });
-
-  const scalesDataGroupedByX = useMemo(
-    () => group(scalesData, getX),
-    [scalesData, getX]
-  );
-
-  const scalesWideData = getWideData({
-    dataGroupedByX: scalesDataGroupedByX,
-    xKey,
-    getY,
-    getSegment,
-  });
-
   const sumsBySegment = useMemo(
     () =>
       Object.fromEntries([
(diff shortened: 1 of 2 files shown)


app/charts/column/columns-stacked-state.tsx
```diff
@@ -448,6 +450,8 @@ const useColumnsStackedState = (
     return series;
   }, [chartWideData, fields.segment?.sorting, plottableSegments]);
 
+  console.log(series, plottableSegments, chartWideData);
```
ptbrowne, line 453: 👁️ ?

ptbrowne (approved):

AFoletti:
I did some quick visual tests. There are in my opinion still issues somewhere.
image

This is however the area chart, while the original ticket was just about the barchart. The barchart looks actually very good to me

AFoletti:
> @AFoletti I think we once talked about making the imputation call to action more prominent (#1060), maybe we could think about implementing that in the near future.
> 
> cc @sosiology @adintegra

Yes... I may be stoooopid, but I fall everytime in that trap. I think the call to action must be visible above the chart, without the need to actively open the category configuration
image

PR #1097, opened 2023-07-04
refactor: Consolidate chart logic

bprusinowski:
This PR consolidates common chart-state-computing logic by extracting some chart logic outside of individual chart states:
- computation of data needed by charts (`chartData`, `scalesData`, `segmentData`, `allData`). I introduced this distinction to make all charts "ready for interactive filters" (e.g. maps do not have them yet) and to have one place to "fix" things, instead of needing to go to every chart,
- sorting the data. Right now every chart type defines its own sorting / missing values filtering logic and this information is used when computing the data upfront, before it's passed to the chart.

It also fixes some minor things:
- column exit animations,
- inner band scale when toggling interactive legend filters,
- tests with dependencies that import `rdf-js` modules (_SyntaxError: Cannot use import statement outside a module_),
- scales now adapt to color legend filter when using time slider,
- it's now not possible to remove all color legend filters to prevent showing no chart.

### Reference: Interactive filters
- Time range
- Time slider
- Color legend

### New chart data structure
- `chartData`: data with all interactive filters applied, used to draw the shapes,
- `scalesData`: data with color legend and time range filters applied, used to compute the scales. As time slider filter is not applied to it, it holds the information about all potential scale values and makes it possible to fix the scales when animating the charts,
- `segmentData`: data with time range filter applied. Needed to hold the information about all color values that can be used in a chart, to save unchecked color legend items from being removed,
- `allData`: full dataset, needed to show the timeline when using time slider. We can't use `scalesData` here, due to the fact the the `useChartData` hook gets re-rendered and prevents the timeline from working it such case.

### Next steps
- Consolidate segments computation logic.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,28 +1,42 @@
 {
   "extends": "next",
-  "plugins": ["unused-imports", "import", "visualize-admin"],
+  "plugins": ["import", "visualize-admin"],
   "rules": {
-    "no-restricted-imports": ["error", {
-      "name": "lodash",
-      "message": "Please use direct imports instead, ex: lodash/mapValues, lodash/groupBy."
-    }],
-    "react/display-name": "off", 
+    "no-restricted-imports": [
+      "error",
+      {
+        "name": "lodash",
+        "message": "Please use direct imports instead, ex: lodash/mapValues, lodash/groupBy."
+      }
+    ],
+    "react/display-name": "off",
     "visualize-admin/no-large-sx": "error",
     "visualize-admin/make-styles": "error",
-    "unused-imports/no-unused-imports-ts": "error",
     "@next/next/no-html-link-for-pages": ["error", "app/pages/"],
-    "import/order": [2, {
-      "alphabetize": { "order": "asc" },
-      "newlines-between": "always",
-      "groups": ["builtin", "external", "internal", "parent", "sibling", "index", "object", "type"],
-      "pathGroups": [
-        {
-          "pattern": "@/**",
-          "group": "internal"
-        }
-      ],
-      "pathGroupsExcludedImportTypes": ["builtin"]
-    }]
+    "import/order": [
+      2,
+      {
+        "alphabetize": { "order": "asc" },
+        "newlines-between": "always",
+        "groups": [
+          "builtin",
+          "external",
+          "internal",
+          "parent",
+          "sibling",
+          "index",
+          "object",
+          "type"
+        ],
+        "pathGroups": [
+          {
+            "pattern": "@/**",
+            "group": "internal"
+          }
+        ],
+        "pathGroupsExcludedImportTypes": ["builtin"]
+      }
+    ]
   },
   "overrides": [
     {
--- a/app/charts/chart-loading-wrapper.tsx
+++ b/app/charts/chart-loading-wrapper.tsx
@@ -4,7 +4,7 @@ import { UseQueryResponse } from "urql";
 
 import { ChartProps } from "@/charts/shared/ChartProps";
 import { A11yTable } from "@/charts/shared/a11y-table";
-import { usePlottableData } from "@/charts/shared/chart-helpers";
+import { useChartData, usePlottableData } from "@/charts/shared/chart-helpers";
 import { getChartStateMetadata } from "@/charts/shared/chart-state";
 import Flex from "@/components/flex";
 import {
@@ -13,7 +13,7 @@ import {
   LoadingOverlay,
   NoDataHint,
 } from "@/components/hint";
-import { ChartConfig, isAnimationInConfig } from "@/configurator";
+import { ChartConfig } from "@/configurator";
 import {
   ComponentsQuery,
   DataCubeMetadataQuery,
@@ -94,6 +94,12 @@ export const ChartLoadingWrapper = <
     getY: chartStateMetadata?.assureDefined.getY,
   });
 
+  const { chartData, scalesData } = useChartData(plottableData, {
+    chartConfig,
+    getXDate: chartStateMetadata?.getXDate,
+    getSegment: chartStateMetadata?.getSegment,
+  });
+
   if (metadata && dimensions && measures && data) {
     const { title } = metadata;
 
@@ -111,7 +117,8 @@ export const ChartLoadingWrapper = <
           measures={measures}
         />
         {React.createElement(Component, {
-          data: plottableData,
+          chartData,
+          scalesData,
           dimensions,
           measures,
           chartConfig,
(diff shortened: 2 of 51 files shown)


.eslintrc.json
```diff
@@ -1,28 +1,42 @@
 {
   "extends": "next",
-  "plugins": ["unused-imports", "import", "visualize-admin"],
+  "plugins": ["import", "visualize-admin"],
```
ptbrowne, line 3: Just for my understanding: the auto fix still removes unused imports ?

app/domain/data.ts
```diff
@@ -228,31 +228,31 @@ export const isGeoDimension = (
...
 export const isOrdinalMeasure = (
-  dimension?: DimensionMetadataFragment
+  dimension?: { __typename?: string } | null
 ): dimension is OrdinalMeasure => {
   return dimension?.__typename === "OrdinalMeasure";
 };
 
 export const isTemporalDimension = (
-  dimension?: DimensionMetadataFragment | null
+  dimension?: { __typename?: string } | null
```
ptbrowne, line 255: Why this change ? Isn't it a bit less semantic ?

.eslintrc.json
bprusinowski: yes :)

app/charts/shared/chart-state.ts
```diff
@@ -1,8 +1,23 @@
...
+export type ChartStateMetadata = {
+  sortData: (data: Observation[]) => Observation[];
+};
+
+export const getChartStateMetadata = (
+  chartConfig: ChartConfig
+): ChartStateMetadata | undefined => {
+  switch (chartConfig.chartType) {
+    case "area":
+      return getAreasStateMetadata(chartConfig);
```
ptbrowne: By having a switch here, we might need to load all the getMetadata functions for all chart types (when we only need the getMetadata for 1 chart type, especially when viewing). I wonder if it would make sense to have a way so that the logic can be imported async.

app/charts/chart-loading-wrapper.tsx
```diff
@@ -86,12 +100,12 @@ export const ChartLoadingWrapper = <
...
         <A11yTable
           title={title}
-          observations={observations.data}
+          observations={observations}
           dimensions={dimensions}
           measures={measures}
         />
         {React.createElement(Component, {
-          data: observations.data,
+          data: observations,
```
ptbrowne: If I am not mistaken, the Component here is loaded async; we could "attach" the getMetadata function a static method to it, so that it's loaded async.

app/charts/map/map-state.tsx
```diff
@@ -578,6 +581,14 @@ const useMapState = (
   };
 };
 
+export const getMapStateMetadata = (): ChartStateMetadata => {
```
ptbrowne, line 593: I think we could have the same method name (getMapStateMetadata -> getMetadata) across all the chart files so that we can load the module async and just call getMetadata on it. This  way, it is even clearer that a chart file needs to implement the getMetadata method.

app/charts/chart-loading-wrapper.tsx
```diff
@@ -87,8 +87,17 @@ export const ChartLoadingWrapper = <
...
-  if (metadataData?.dataCubeByIri && dimensions && measures && data) {
-    const { title } = metadataData.dataCubeByIri;
+  // const { chartData, scalesData } = useChartData(data ?? [], {
+  //   interactiveFiltersConfig: chartConfig.interactiveFiltersConfig,
+  //   animationField: isAnimationInConfig(chartConfig)
+  //     ? chartConfig.fields.animation
+  //     : undefined,
+  //   getXDate: chartStateMetadata?.getXDate,
+  //   getSegment: chartStateMetadata?.getSegment,
+  // });
```
ptbrowne: Is this to remove ?

app/charts/shared/chart-state.ts
```diff
@@ -53,13 +55,13 @@ export const getChartStateMetadata = (
...
-      return getLinesStateMetadata(chartConfig);
+      return getLinesStateMetadata(chartConfig, observations, dimensions);
     case "map":
       return getMapStateMetadata();
     case "pie":
-      return getPieStateMetadata();
+      return getPieStateMetadata(chartConfig, observations, dimensions);
     case "scatterplot":
-      return getScatterplotStateMetadata();
+      return getScatterplotStateMetadata(chartConfig, observations, dimensions);
```
ptbrowne, line 69: So I guess that if we have the same getMetadata method everywhere, we can remove the switch.

app/charts/shared/ChartProps.tsx
```diff
@@ -3,7 +3,8 @@ import { Observation } from "@/domain/data";
 import { DimensionMetadataFragment } from "@/graphql/query-hooks";
 
 export type BaseChartProps = {
-  data: Observation[];
+  chartData: Observation[];
+  scalesData: Observation[];
   dimensions: DimensionMetadataFragment[];
   measures: DimensionMetadataFragment[];
 };
```
ptbrowne, line 16: Here we could have a little `/** comment */` with what you've written in PR description, so that when we are not sure what is scalesData, we can hover over it and see it :)

app/charts/column/columns-stacked.tsx
```diff
@@ -18,40 +13,32 @@ export const ColumnsStacked = () => {
...
+          width: bandwidth,
+          height: yScale(segment[0]) - yScale(segment[1]),
+          color,
+        };
+      });
     });
   }, [bandwidth, colors, getX, series, xScale, yScale]);
 
   React.useEffect(() => {
+    console.log("rendering columns", renderData);
```
ptbrowne, line 37: 👁️

app/charts/column/columns-stacked.tsx
```diff
@@ -18,40 +13,32 @@ export const ColumnsStacked = () => {
...
+
+        return {
+          key: `${key}-${x}`,
+          x: xScale(x) as number,
+          y: yScale(segment[1]),
+          width: bandwidth,
+          height: yScale(segment[0]) - yScale(segment[1]),
+          color,
+        };
+      });
```
ptbrowne, line 32: 👍

app/jest.config.js
```diff
@@ -12,6 +12,10 @@ const config = {
     "^@/(.*)$": "<rootDir>/$1",
     "\\.(css)$": "<rootDir>/test/style-mock.js",
   },
+  transform: {
+    "node_modules/@rdf*": "ts-jest",
+  },
+  transformIgnorePatterns: ["node_modules/!@rdf*"],
```
ptbrowne, line 18: 👏

ptbrowne (approved):
👏  🚀 

So nice to have found all this different "data" :) I have some comments here and there and I think the ability to import async is important, but overall very very nice 💯

PR #1103, opened 2023-07-11
feat: Add a way to normalize Area and Bar charts + animations continuation

bprusinowski:
- Shared chart variables getters
- Added logic to normalize the Area and Stacked Bar charts (not yet  connected to the UI option)
- Changes the aspect ratios of Pie and Scatter charts to fit on the screen (only in Edit mode)
- Interactive time range filter is now constrained by filters set in the X field
- Interactive time range filter now keeps component iri, which previously was always empty
- Time slider values are now shown in filters list at the top of the charts
- Tooltip message is now displayed when user wants to de-select last color legend item
- Added TemporalOrdinalDimension (and a way to use it with TimeSlider)
- Whiskers are now animated


Rdataflow:
@kronmar stay tuned on _ordinal time_ and see https://visualization-tool-purx4zcwb-ixt1.vercel.app/create/new?cube=https://environment.ld.admin.ch/foen/nfi/nfi_ordinal_time/cube/2023-1&flag__timeslider=true (currently WIP, publish is broken)

app/charts/column/columns-stacked-state.tsx
```diff
@@ -108,22 +108,15 @@ const useColumnsStackedState = (
...
-      ]),
-    [getSegment, getY, scalesData]
-  );
+  const sumsBySegment = useMemo(() => {
+    return Object.fromEntries([
+      ...rollup(
+        scalesData,
+        (v) => sum(v, (x) => getY(x)),
+        (x) => getSegment(x)
+      ),
```
ptbrowne, line 126: Here we need to spread because fromEntries does not take an iterator ?

app/charts/shared/chart-data-filters.tsx
```diff
@@ -334,17 +336,16 @@ const DataFilterTemporalDimension = ({
...
-  );
+  const timeIntervalWithProps = React.useMemo(() => {
+    return getTimeIntervalWithProps(
+      options[0].value as string,
+      options[options.length - 1].value as string,
+      timeUnit,
+      timeFormat,
+      formatLocale
+    );
+  }, [options, timeUnit, timeFormat, formatLocale]);
```
ptbrowne, line 345: It seems you would like the functionality "Add {} to arrow function" from the "JS Refactoring" extension 😁 

image

image

app/locales/fr/messages.po
```diff
@@ -333,6 +333,10 @@ msgstr "Tout sélectionner"
 msgid "controls.filter.select.none"
 msgstr "Tout déselectionner"
 
+#: app/charts/shared/legend-color.tsx
+msgid "controls.filters.interactive.color.min-1-filter"
+msgstr "Au moins un filtre doit être sélectionné."
```
ptbrowne, line 338: 💯

app/charts/chart-config-ui-options.ts
```diff
@@ -6,7 +6,7 @@ import {
   SortingOrder,
   SortingType,
 } from "@/config-types";
-import { isTemporalDimension } from "@/domain/data";
+import { isTemporalDimension, isTemporalOrdinalDimension } from "@/domain/data";
```
ptbrowne, line 9: Is it to support the inventory time periods from WSL ?

app/configurator/interactive-filters/time-slider.tsx
```diff
@@ -169,14 +191,49 @@ const Slider = () => {
...
+        },
       });
     }
-  }, [timeline.value, IFState.timeSlider.value, dispatch]);
+  }, [
+    timeline.value,
+    IFState.timeSlider.type,
+    IFState.timeSlider.value,
+    dispatch,
+  ]);
```
ptbrowne, line 258: Why do we need an useEffet here ? Couldn't the dispatch be done in an event handler ? Or directly in the reducer ?

app/configurator/interactive-filters/time-slider.tsx
```diff
@@ -75,17 +75,19 @@ export const TimeSlider = (props: TimeSliderProps) => {
...
       msDuration: animationDuration * 1000,
     };
-    const values = chartState.allData
-      .map((d) => d[component.iri])
-      .filter(Boolean) as string[];
+    const uniqueValues = Array.from(
+      new Set(
+        chartState.allData
+          .map((d) => d[component.iri])
+          .filter(Boolean) as string[]
```
ptbrowne, line 86: I think there is a isTruthy or a truthy helper that can replace the Boolean here, and would help not having to do the `as string[]`.

ptbrowne (approved):
Thanks Bartosz. The code looks LGTM and I did not see any blocker.

---

General observations

- chart-state
    * Some logic concerning how dimensions are found could be memo-ed ?
    * Maybe we could have an index iri -> dimension/measure to not have to find()
    * Wonder if we could not share code for the dimension hooks useNumerical/useTemporal etc...

    ```
    const useNumericalXVariables = makeVariables({
        dimFilter: isNumericalMeasure,
        field: 'x',
        isOptional: true
        label: getLabelWithUnit
    })
    ```

- I'd like to be able to exercise the changes functionally. It's a bit hard right now since I need to know which charts to exercise, with the right options etc... I think it'd be good to invest some time in developing stories. I think we have the start of this via the mock chart config. Maybe having storybook could help here ?

app/charts/shared/chart-data-filters.tsx
bprusinowski: Ooo, definitely! 👀

app/charts/chart-config-ui-options.ts
bprusinowski: Exactly :)

app/charts/column/columns-stacked-state.tsx
bprusinowski: In fact we don't need to spread, good catch!

app/configurator/interactive-filters/time-slider.tsx
bprusinowski: Hmmm, I think it can't be done in event handler, as this `useEffect` is also called when timeline is playing (and the `onChange` event only happens when user interacts with the timeline) :(

Or did you mean something else with the reducer? I think we wouldn't have access to timeline in the InteractiveFilters reducer 🤔

bprusinowski:
Thanks for reviewing @ptbrowne! I addressed the comments in #1105 and created a new issue to add Storybook (#1109).

For the shared logic to create variable getters, I am not sure it would be super easy, as for some variables we use abbreviations, and for other (measures) we don't (also we need custom names for variables, like getX, getY, etc, so I think we would need to unpack, rename and then pass them further)? Let me know if I missed something in this approach!

PR #1105, opened 2023-07-14
feat: Animation pt 3 & normalised charts

bprusinowski:


app/charts/shared/brush/index.tsx
```diff
@@ -321,6 +321,7 @@ export const BrushTime = () => {
     const g = select(ref.current);
 
     const coord = [brushWidthScale(closestFrom), brushWidthScale(closestTo)];
+    console.log(coord);
```
ptbrowne: 👁️

app/charts/chart-loading-wrapper.tsx
```diff
@@ -74,6 +75,8 @@ export const ChartLoadingWrapper = <
 
   if (metadata && dimensions && measures && observations) {
     const { title } = metadata;
+    const dimensionsByIri = keyBy(dimensions, (d) => d.iri);
+    const measuresByIri = keyBy(measures, (d) => d.iri);
```
ptbrowne, line 79: Nice 💯 

Nit: Maybe you should memo the two variables so that props of the created chart below are stable. That would enable the use of React.memo on the charts.

app/charts/shared/axis-width-band.tsx
```diff
@@ -77,7 +95,37 @@ export const AxisWidthBandDomain = () => {
...
+              .duration(TRANSITION_DURATION)
+              .attr(
+                "transform",
+                `translate(${margins.left}, ${chartHeight + margins.top})`
+              )
+              .call(axisBottom(xScale).tickSizeOuter(0))
+          ),
+        (exit) => exit.transition().duration(TRANSITION_DURATION).remove()
+      );
+
```
ptbrowne, line 128: Nit: Seems like we are repeating ourselves here for each axis, could we refactor ?

app/charts/line/lines-state.tsx
```diff
@@ -204,15 +205,9 @@ const useLinesState = (
...
-  const chartHeight = chartWidth * aspectRatio;
-  const bounds = {
-    width,
-    height: chartHeight + margins.top + margins.bottom,
-    margins,
-    chartWidth,
-    chartHeight,
-  };
+  const bounds = getChartBounds(width, margins, aspectRatio);
+  const { chartWidth, chartHeight } = bounds;
```
ptbrowne, line 209: 💯

app/charts/column/columns-grouped.tsx
```diff
@@ -57,16 +58,43 @@ export const ErrorWhiskers = () => {
...
+          .data(renderData, (d) => d.key)
+          .call(renderWhisker);
+      };
+
       select(ref.current)
-        .selectAll<SVGGElement, RenderWhiskerDatum>("g")
-        .data(renderData, (d) => d.key)
-        .call(renderWhisker);
+        .selectAll<SVGGElement, null>(".content")
+        .data([null])
```
ptbrowne, line 71: For my information, why do we put a null here ?

app/charts/scatterplot/scatterplot.tsx
```diff
@@ -50,11 +51,33 @@ export const Scatterplot = () => {
...
+                  .duration(TRANSITION_DURATION)
+                  .attr(
+                    "transform",
+                    `translate(${margins.left} ${margins.top})`
+                  )
+              )
+              .call(renderCircles, renderData),
+          (exit) => exit.remove()
+        )
+        .call(renderCircles, renderData);
```
ptbrowne, line 78: Nit: Seems like this logic is repeated across many files

Do you think it would be useful to have `renderAnimated(ref.current, renderCircles, renderData)` ?

app/charts/area/areas-state.tsx
```diff
@@ -184,7 +184,7 @@ const useAreasState = (
     );
   }, [getXAsString, getY, scalesData]);
 
-  const normalize = fields.segment?.calculation === "percent";
+  const normalize = interactiveFiltersConfig?.calculation.type === "percent";
```
ptbrowne: Why is it in the interactiveFilters config ? It seems we could have a stacked percent bar chart without anything interactive 🤔  ?

app/charts/table/table-state-props.ts
```diff
@@ -1,9 +1,28 @@
...
+export const useTableStateVariables = (
+  props: ChartProps<TableConfig>
+): TableStateVariables => {
+  const { chartConfig } = props;
+
+  const baseVariables = useBaseVariables(chartConfig);
+
+  return {
+    ...baseVariables,
+  };
```
ptbrowne, line 23: Nit: No need to spread here ?

app/locales/en/messages.po
```diff
@@ -203,6 +203,22 @@ msgstr "Horizontal Axis"
...
+#: app/configurator/components/field-i18n.ts
+msgid "controls.calculation.identity"
+msgstr "Absolute figures"
+
+#: app/configurator/components/field-i18n.ts
+msgid "controls.calculation.percent"
+msgstr "Relative figures"
+
+#: app/charts/shared/interactive-filter-calculation-toggle.tsx
+msgid "controls.caluculation.show-in-percentage"
```
ptbrowne, line 219: caluculation ? 👁️

app/locales/fr/messages.po
```diff
@@ -203,6 +203,22 @@ msgstr "Axe horizontal"
...
+msgid "controls.calculation.allow-normalization"
+msgstr "Permettre la normalisation des données"
+
+#: app/configurator/components/field-i18n.ts
+msgid "controls.calculation.identity"
+msgstr "Chiffres absolus"
+
+#: app/configurator/components/field-i18n.ts
+msgid "controls.calculation.percent"
+msgstr "Chiffres relatifs"
```
ptbrowne, line 216: Maybe "Pourcentages" here.

app/locales/fr/messages.po
```diff
@@ -203,6 +203,22 @@ msgstr "Axe horizontal"
 msgid "controls.axis.vertical"
 msgstr "Axe vertical"
 
+#: app/configurator/components/chart-options-selector.tsx
+msgid "controls.calculation.allow-normalization"
+msgstr "Permettre la normalisation des données"
+
+#: app/configurator/components/field-i18n.ts
+msgid "controls.calculation.identity"
+msgstr "Chiffres absolus"
```
ptbrowne, line 212: Maybe "Valeurs absolues" here ?

ptbrowne (approved):
You're on 🔥 I did not see any blockers. Maybe some time I see duplicated logic (mostly concerning how elements are animated), I wonder if this could be abstracted for easier reuse.

LGTM 👏

ptbrowne:
I noticed while testing that the switch "Show in %" is disabled, is this wanted ? It is a bit disconcerting ?

app/charts/area/areas-state.tsx
bprusinowski: Honestly I wasn't 100% sure where exactly to put this (that's why initially it was inside the segment field). I thought it would be simpler to just have it inside the `interactiveFiltersConfig`, because there we do not need to do the sorts of checks "if segment exists" and then "if calculation property exists in segment" (and now we still can publish a 100%-based chart without actually enabling interactive filter [when `active` is false]).

I think that overall the refactor to keep it again in the segment wouldn't change anything in the functionality, but maybe it would be more "correct" 👍 So it would look like:
- For Stacked Column Chart and Stacked Area Chart we have a `calculation` property inside `segment`
- For every chart we have `calculation` property in `interactiveFiltersConfig` with `active`: boolean and `type`: "identity" | "percent"

In fact when I'm looking at the above proposal, it seems that we would duplicate the information inside `segment` and `interactiveFiltersConfig`. Maybe we can sync about this, let me know what do you think 👀

bprusinowski: The current approach for reference:
- we keep the information about calculation in `interactiveFiltersConfig` (static config that is only changeable while editing the chart). The calculation type is "identity" for every chart, except for stacked charts that could use "percent",
- the dynamic part is done via `InteractiveFiltersState`, which overrides the properties from `interactiveFiltersConfig` for each end user.

bprusinowski:
Thanks for reviewing @ptbrowne 🙇 

You mean that the `Allow normalization of the data` checkbox is not disabled? I just noticed this and will fix soon 👍 

Screenshot 2023-08-04 at 13 59 31

app/charts/column/columns-grouped.tsx
bprusinowski: I used the `join` pattern here to have access to every lifecycle event of the element (and not to animate the `transform` property initially) – because we are using `selectAll`, and the `[null]` is an 1-element array that makes the logic render one element. It could be anything, but as the data in this context doesn't matter, it's a convention (I think?) to use `null` :)

app/charts/area/areas-state.tsx
bprusinowski: Also, I just remembered, when the `calculation` was an existing property only for stacked mode in column chart and was removed for grouped chart, it made the radio buttons not checked at all in the screenshot sent here: https://github.com/visualize-admin/visualization-tool/pull/1105#issuecomment-1665496338

ptbrowne:
I meant the switch button on top right of the "preview" chart of the edit mode

app/charts/area/areas-state.tsx
bprusinowski: Decision: keep the calculation type in `segment` (or think again why it should be in `Y`), remove the calculation type from `interactiveFiltersConfig`. Also, duplicate the section to select the type in the `Y` field.

PR #1112, opened 2023-08-07
feat: Double-value tooltips (percentage + absolute)

### LINKED ISSUE #1093: Tooltips: Display absolute values + percentages
As a user viewing a percentage chart based on absolute values, e.g. pie chart or bar chart (issue [#1092](https://github.com/bafu/umweltdatenkiosk-planning/-/issues/509)), I would like the tooltip to show me not only the absolute values but also the % values.

Image

### LINKED ISSUE #1111: Tooltip units for figures with relative measurements
**Describe the bug**
If a barplot is normalized, the measurement's tooltip inherit the absolute value's unit. 

**To Reproduce**
See for example [here](https://test.visualize.admin.ch/en/v/esoDl7HPVFwK). Hovering over displays the bars shows the unit in cubic meters (and with arguably way too many digits)

**Expected behavior**
Units should either change to %, or the original values should be displayed. Maybe even both?

**Screenshots or video**
image

- Visualize environment and version: v3.20.3 ([99cea84](https://github.com/visualize-admin/visualization-tool/commit/99cea84a75bd78e59eea0124d134d21fe37d7c7b))
- Browser and version: Edge Version 115.0.1901.188 (Official build) (64-bit), but the same behavior is seen in Chrome Version 115.0.5790.171 (Offizieller Build) (64-Bit)

bprusinowski:
Closes #1093.
Closes #1111.

This PR:
1. adds support for **double-value tooltips** for charts that display percentages (pie, stacked area, stacked column),

**How to test**
- Look for the `Einmalvergütung für Photovoltaikanlagen` dataset.
- Create a stacked column chart.
- Change its mode to `relative figures`.
- Hover over bars to see that the tooltip displays both percentages and original values.
- Change chart type to `area` or `pie` – see that the tooltip still works in this new mode.

2. **improves tooltip positioning** for selected chart types (area, column, line) and fixes the bug with too many event dispatches in the brush component. This was made possible by refactoring the interactive filters state to use Zustand's store instead of React's context to be able to access current state outside of the component context,

**How to test**
- Create an area, column or line chart.
- See that the tooltip is centered when hovering over data points.
- Make the chart stacked (area or line).
- See that the tooltip is attached horizontally to the middle of shape's height and flips the position around 25% width of the chart.

3. **makes the animations _disabled_** during time range filtering (by using another Zustand store for storing the transition duration),

**How to test**
- Look for the `Einmalvergütung für Photovoltaikanlagen` dataset.
- Create a stacked area chart.
- Normalise the chart and see that it's animating.
- Enable interactive time filter from the X field.
- Interact with the interactive filter and see that there is no animation.

4. disables the possibility to enable animations when sorting by total size and vice versa.

**How to test**
- Go to the Bathing water quality dataset.
- Use Parameter as X field.
- Use Bathing site as color (stacked column chart).
- Change the sorting to total size.
- See that the animation is disabled.
- Remove total size sorting.
- Enable animation.
- Go to sorting and see that total size is disabled.

### Notes
See #1113.

_I am going to consolidate the d3 rendering (`join` pattern) and properly disable transitions along the way. Right now it works, but the transition with 0ms duration is still being fired, and a minor lag is noticeable when brushing the area chart on the home page. It should not be a transition with 0ms duration, but rather not be a transition at all._


app/charts/column/columns-stacked-state.tsx
```diff
@@ -52,12 +52,12 @@ import {
 } from "@/charts/shared/stacked-helpers";
 import useChartFormatters from "@/charts/shared/use-chart-formatters";
 import { InteractionProvider } from "@/charts/shared/use-interaction";
-import { useInteractiveFilters } from "@/charts/shared/use-interactive-filters";
 import { Observer, useWidth } from "@/charts/shared/use-width";
 import { ColumnConfig } from "@/configurator";
 import { Observation } from "@/domain/data";
 import { useFormatNumber } from "@/formatters";
 import { getPalette } from "@/palettes";
+import { useInteractiveFiltersStore } from "@/stores/interactive-filters";
```
ptbrowne, line 60: My main concern concerning using a zustand store is that it is global, meaning that if I am not mistaken, it will be problematic to display 2 charts on the same page.

bprusinowski: Yes, but we'll be able to have something like `createInteractiveFiltersStore` function, and that store could be then created within each chart (and there might be a case, where they should be global, as will be the case for dashboards) 👀

app/charts/shared/brush/index.tsx
```diff
@@ -106,6 +106,11 @@ export const BrushTime = () => {
         xEnd,
       ]);
 
+      // Need to use current state as the function is not updated during brushing
+      // and the local state accessed here is not up to date. This leads to
+      // making a dispatch on each brush move, which makes the animations laggy
+      // and generally shouldn't happen.
+      const { from, to } = useInteractiveFiltersStore.getState().timeRange;
```
ptbrowne, line 120: Reading the comment on the commit, I am wondering why are we using an old brush component ? Why is there several versions of the brush ?

Here if I am correct, the brush is "disconnected" (in the sense that if something external to the brush changes from/to, the brush will not be re-rendered) from the interactive filters store, because this is the one controlling it ?

app/charts/column/columns-stacked-state.tsx
ptbrowne: Yes 👍  Then you will need a provider around the chart to provide the created store right ? In the end it will not be that different than with the provider right ? Do you see an advantage ? Not saying that we should change it back, jsut wanting to know that there is advantage I am not seeing.

app/charts/shared/brush/index.tsx
bprusinowski: I wasn't 100% sure about the cause, but the problem was that we accessed the old interactive filters state inside the `brushed` function. I guess this function is "frozen" once user starts interacting with the brush (that's what I meant with the old brush 😅), and when the interactive filters state updates, this is not detected, because we are using the old callback. By changing the logic to the line you commented, we are able to always access up-to-date state.

app/charts/column/columns-stacked-state.tsx
bprusinowski: Yes, we would need a provider then (or change the logic to pass the object down to the components). I used Zustand to be able to use `getState()` method which works outside of the React component context (and allowed to use up-to-date state in the brush callback). Other way of solving it would be to create a "global" variable getter, but I wanted to avoid that 👀

app/stores/interactive-filters.ts
```diff
@@ -58,9 +58,8 @@ export const useInteractiveFiltersStore = create<
     },
     removeCategory: (category: string) => {
       return set((state) => {
-        const { categories } = state;
-        delete categories[category];
-        return { ...state, categories };
+        delete state.categories[category];
+        return { ...state, categories: { ...state.categories } };
```
ptbrowne, line 66: `return {...state, categories: omit(state.categories, category) )` could be a little clearer.

app/charts/column/columns-stacked-state.tsx
bprusinowski: Also, with atomic state picks we can only re-render the components when a given value changes (and I believe in the old way, we were re-rendering a component when any value changed inside the reducer) 🤔

app/components/metadata-panel.tsx
```diff
@@ -304,9 +308,15 @@ export const MetadataPanel = ({
...
+        disableEnforceFocus
         ModalProps={{ container }}
         PaperProps={{ style: { position: "absolute" } }}
-        disableEnforceFocus
+        SlideProps={{
+          onEnter: setInstantDuration,
+          onEntered: setDefaultDuration,
+          onExit: setInstantDuration,
+          onExited: setDefaultDuration,
+        }}
```
ptbrowne, line 319: Shouldn't we "unsetDefaultDuration" at some point ?

ptbrowne: Ah excuse me, I read wrong and did not see setDefaultDuration.

ptbrowne (approved):
LGTM 👏 

Nit: Concerning the useTransitionStore, I think we should be careful about adding too much one-off stores to be used by the charts, it can be become too much in the future. Maybe have this particular store renamed into useChartOptionStore ? This store will also need to be created dynamically and passed through a provider in case of multiple charts, we should consider if necessary to merge it with useInteractiveFilterStore (and in this case rename).

app/stores/interactive-filters.ts
bprusinowski: I wrote this down and will tackle this in next PR 👌

PR #1113, opened 2023-08-10
refactor: Consolidate d3 rendering

bprusinowski:
This PR aims to:
- **consolidate D3 rendering** across charts and axes by adding `renderContainer` and `maybeTransition` methods,
- **properly disable transitions** with the help of `maybeTransition` function. This function makes it possible to specify the rendering logic when transitions are enabled (`t`) and disabled (`s`). I also shared the logic to base this decision on `transitionStore`,
- **fix brush disabling the transitions** when metadata panel is being toggled,
- improve performance of using Zustand stores (atomic state picks),
- **fix jumping charts** during interactions with interactive filters,
- introduce minor style adjustments.

### DIFF
--- a/app/charts/shared/brush/index.tsx
+++ b/app/charts/shared/brush/index.tsx
@@ -25,6 +25,7 @@ import { estimateTextWidth } from "@/utils/estimate-text-width";
 // Brush constants
 export const HANDLE_HEIGHT = 14;
 export const BRUSH_HEIGHT = 3;
+export const HEIGHT = HANDLE_HEIGHT + BRUSH_HEIGHT;
 
 export const BrushTime = () => {
   const ref = useRef<SVGGElement>(null);
@@ -332,13 +333,13 @@ export const BrushTime = () => {
   }, [brushWidth]);
 
   return (
-    <>
+    <g
+      transform={`translate(0, ${
+        chartHeight + margins.top + margins.bottom - HEIGHT * 1.5
+      })`}
+    >
       {/* Selected Dates */}
-      <g
-        transform={`translate(0, ${
-          chartHeight + margins.top + margins.bottom / 2
-        })`}
-      >
+      <g>
         {closestFrom && closestTo && (
           <text
             fontSize={labelFontSize}
@@ -353,11 +354,7 @@ export const BrushTime = () => {
       </g>
 
       {/* Brush */}
-      <g
-        transform={`translate(${brushLabelsWidth}, ${
-          chartHeight + margins.top + margins.bottom / 2
-        })`}
-      >
+      <g transform={`translate(${brushLabelsWidth}, 0)`}>
         {/* Visual overlay (functional overlay is managed by d3) */}
         <rect
           x={0}
@@ -368,13 +365,8 @@ export const BrushTime = () => {
         />
       </g>
       {/* actual Brush */}
-      <g
-        ref={ref}
-        transform={`translate(${brushLabelsWidth}, ${
-          chartHeight + margins.top + margins.bottom / 2
-        })`}
-      />
-    </>
+      <g ref={ref} transform={`translate(${brushLabelsWidth}, 0)`} />
+    </g>
   );
 };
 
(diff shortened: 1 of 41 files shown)


app/charts/shared/brush/index.tsx
```diff
@@ -28,16 +28,10 @@ export const BRUSH_HEIGHT = 3;
...
-  const { setDefaultDuration, setInstantDuration } = useTransitionStore(
-    (d) => ({
-      setDefaultDuration: d.setDefaultDuration,
-      setInstantDuration: d.setInstantDuration,
-    })
-  );
+  const timeRange = useInteractiveFiltersStore((d) => d.timeRange);
+  const setTimeRange = useInteractiveFiltersStore((d) => d.setTimeRange);
+  const setDefaultDuration = useTransitionStore((d) => d.setDefaultDuration);
+  const setInstantDuration = useTransitionStore((d) => d.setInstantDuration);
```
ptbrowne: Nit: 
const { timeRange, setTimeRange } =  useInteractiveFiltersStore((d) => { timeRange: d.timeRange, setTimeRange: d.setTimeRange });

app/charts/shared/chart-data-filters.tsx
```diff
@@ -49,6 +49,7 @@ export const ChartDataFilters = ({
 }) => {
   const [filtersVisible, setFiltersVisible] = React.useState(false);
   const { componentIris } = dataFiltersConfig;
+  console.log("ChartDataFilters");
```
ptbrowne: 👁️

app/charts/shared/use-sync-interactive-filters.tsx
```diff
@@ -18,15 +18,17 @@ import { useInteractiveFiltersStore } from "@/stores/interactive-filters";
...
+  const resetCategories = useInteractiveFiltersStore((d) => d.resetCategories);
+  const dataFilters = useInteractiveFiltersStore((d) => d.dataFilters);
+  const setDataFilters = useInteractiveFiltersStore((d) => d.setDataFilters);
+  const updateDataFilter = useInteractiveFiltersStore(
+    (d) => d.updateDataFilter
+  );
+  const setTimeRange = useInteractiveFiltersStore((d) => d.setTimeRange);
+  const setCalculationType = useInteractiveFiltersStore(
+    (d) => d.setCalculationType
+  );
```
ptbrowne, line 31: Nit: Wondering, maybe it'd be cool to have next to useInteractiveFIltersStore definition more specialized hooks

const useTimeRange = () => useInteractiveFiltersStore(x => ({ timeRange: x.timeRange, setTimeRange: x.setTimeRange ]) 

etc...

app/stores/interactive-filters.ts
```diff
@@ -53,17 +53,14 @@ export const useInteractiveFiltersStore = create<
...
+      set((state) => ({
+        categories: { ...state.categories, [category]: true },
+      }));
     },
     removeCategory: (category: string) => {
-      return set((state) => {
+      set((state) => {
         delete state.categories[category];
-        return { ...state, categories: { ...state.categories } };
+        return { categories: { ...state.categories } };
```
ptbrowne, line 63: omit ?

ptbrowne (approved):
Cool that the chart rendering logic is consolidated !! 👏

app/charts/shared/brush/index.tsx
bprusinowski: Yes, we could do it like this, but then we need the `shallow` option passed to the `useInteractiveFiltersStore` (see shallow here: https://docs.pmnd.rs/zustand/recipes/recipes), otherwise a new object is created at each render at everything re-renders 😨 With atomic picks, it does the `===` comparison by default.

I felt into this trap and spent some time finding out what's happening in other component 😅 

I can change this to destructuring with `shallow` 👍

app/charts/shared/use-sync-interactive-filters.tsx
bprusinowski: Yup, and then we could have the shallow options built-in into this hook, this would be cleaner 👍

bprusinowski: Will change that in next PR :)

bprusinowski:
Hi @adintegra, thanks for reviewing! I also noticed and fixed this in #1115 👌

PR #1115, opened 2023-08-11
feat: UI improvements

bprusinowski:
This PR:
- makes the field editor drawer options collapsible,
- consolidates border colors across the app,
- fixes the initial scroll bug in the editor drawer,
- makes the field title more prominent,
- fixes computation of normalised charts' left margins,
- introduces StandardErrorDimension, to be able to filter them out in non-hacky way,
- fixes the "Show standard error" option not being shown (can be tested with `NFI: Topics by stage of stand development` dataset and a column chart),
- adds standard errors to grouped column charts' tooltips,
- fixes the filters badge count (now we only count single filters, the ones that the user selects in this place),
- aligns the formatting of multi- and single-value tooltips,
- adds a new `Layout options` section when editing the segment field.


app/configurator/components/ui-helpers.ts
```diff
@@ -145,7 +145,7 @@ export const getIconName = (name: string): IconName => {
     case "y":
       return "yAxis";
     case "segment":
```
ptbrowne, line 147: I am curious, why this change ? It seems like Color is more user friendly.

app/themes/federal.tsx
```diff
@@ -711,7 +711,7 @@ theme.components = {
         fontWeight: "bold",
         lineHeight: 1.5,
         borderBottom: "1px solid",
-        borderBottomColor: theme.palette.grey[500],
+        borderBottomColor: theme.palette.divider,
```
ptbrowne, line 714: 💯

app/configurator/components/ui-helpers.ts
bprusinowski: We decided to change it, as this section grew significantly over the time - now we also have the `Chart mode` here, as color technically is a sub-option of segmentation (color palette). Not sure if this explanation makes sense, but segmentation and color and not on the same level conceptually, at least we came to such conclusion 😅

app/domain/data.ts
```diff
@@ -274,7 +274,7 @@ export const isStandardErrorResolvedDimension = (dim: ResolvedDimension) => {
 };
 
 export const isStandardErrorDimension = (dim: DimensionMetadataFragment) => {
-  return dim?.related?.some((r) => r.type === "StandardError");
+  return dim.__typename === "StandardErrorDimension";
```
ptbrowne, line 277: Looks cleaner 😁

app/components/metadata-panel.tsx
```diff
@@ -629,6 +629,7 @@ const DimensionValues = ({ dim }: { dim: DimensionMetadataFragment }) => {
       return <DimensionValuesNominal values={sortedValues} />;
     case "NumericalMeasure":
     case "TemporalDimension":
+    case "StandardErrorDimension":
```
ptbrowne, line 632: Gotta love the exhaustive check 😁

app/configurator/components/chart-controls/section.tsx
```diff
@@ -9,83 +9,86 @@ import {
...
+  (theme) => ({
+    root: {
+      display: "flex",
+      alignItems: "center",
+      width: "100%",
+      padding: theme.spacing(4),
+      paddingBottom: theme.spacing(4),
+      marginBottom: ({ gutterBottom }: SectionTitleStylesProps) =>
+        gutterBottom ? 0 : -theme.spacing(2),
+      transition: "padding-bottom 300ms ease",
```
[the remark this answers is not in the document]
bprusinowski: It seems that we not use it anyways (the animation), as there is a static spacing(4) as padding bottom. I will remove it 👍

app/themes/federal.tsx
```diff
@@ -669,7 +669,7 @@ theme.components = {
         "& .MuiSwitch-track": {
           borderRadius: 16 / 2,
           opacity: 1,
-          backgroundColor: theme.palette.background.paper,
+          backgroundColor: "#4F4F4F",
```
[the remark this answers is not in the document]
bprusinowski: Yes, you're right, in fact it's Secondary / Active in Figma so I guess I'll need to update the secondary colors then 👍

ptbrowne (approved):

PR #1121, opened 2023-08-17
chore: Correctly set Sentry environments

### LINKED ISSUE #1118: Align Sentry environments with Visualize deployment environments
Currently, the Visualize TEST, INT and PROD environments aren't fully reflected in Sentry. While "Production" appears to be correctly configured, there is no differentiation between TEST (i.e. the development environment) and INT (integration/test environment).

image

While not critical, it could be useful to have the three environments (along with the Vercel environments) show up correctly in Sentry so that the differentiation of metrics in TEST and INT is clear.

bprusinowski:
Closes #1118.

This PR aligns the Visualize deployment environments with Sentry environments. It means that we will have four environments:
- `vercel` (events from Vercel deployments),
- `test` (events from https://test.visualize.admin.ch/),
- `int` (events from https://int.visualize.admin.ch/),
- `prod` (events from https://visualize.admin.ch/).

@adintegra let me know if this makes sense, or if you'd like to keep the current separation when it comes to Vercel (`vercel-development`,` vercel-production`). It seems that `vercel-production` is listening for events from the deployments of the `main` branch (https://visualization-tool-alpha.vercel.app/), and we use a `test` deployment to test these changes anyway.

If we stick to the new, consolidated `vercel` environment, it would make sense to hide the depreciated` -production` and `-development` environments from Sentry.io.


app/sentry-utils.ts
adintegra, line 1: Do we need a default return value for getSentryEnv() in case `IS_VERCEL` isn't defined?

adintegra (approved):
Thanks for the changes, @bprusinowski 🚀 LGTM

app/sentry-utils.ts
bprusinowski: In case the `IS_VERCEL` is undefined, we'll default to the default env logic (https://github.com/getsentry/sentry-javascript/blob/ff268873130e80049e887ac3dfce48d35e548c08/packages/nextjs/src/client/index.ts#L50), which will then result in `NODE_ENV`.

We could add something like "not set", but I think this should really happen, as these are our only deployment environments (Vercel, TEST, INT, PROD), and events from localhost are not sent anyway. What do you think?

adintegra: Thanks for that clarification @bprusinowski. Had not taken into account the code you pointed out. Perfect!

Issue #1106, opened 2023-07-14
Visualize cannot display big shared dimension in a map


CDiGallo:
Edit:
In the meantime I totally destroyed the cube in this post. The problem stays the same but please use the link from the first comment below.

**Describe the bug**
Shared dimension municipalities cannot get loaded as a map on visualize. Even on 1 Municapilites i waited longer nthan 270 min.
The other charts, such as bar chart work in <1 second.

**To Reproduce**

1. for cantons it works: https://int.[visualize.admin.ch](https://int.visualize.admin.ch/v/52lYVlmgczBm?dataSource=Int)/v/52lYVlmgczBm?dataSource=Int
2. for municpalities not: https://int.visualize.admin.ch/en/create/2BeD8NYUJLMG?dataSource=Int
3.  Click on bar chart -> works
4. click on maps -> (probably) infinite loading

**Expected behavior**
see the canton visualization. I want to have the municipalities in the color based on a measure-dimension.

**Screenshots or video**
If applicable, add screenshots or a short video to help put your problem into a context.

**Environment (please complete the following information):**

- Visualize environment and version: INT
- Browser and version EDGE

**Additional context**

CDiGallo:
I made some further investigations.
I tested in an extra [cube](https://test.cube-creator.lindas.admin.ch/app/cube-projects/cube-project!!municpaltiestest4entries-l0yd0gwej07/publication) how visualize behaves with less municipalities (so confirming, that it is actually a perfomance issue).

with less than a hundred cases, it is very fast.
With 1.5k municipalties it has around 5 mins or a bit less, but it still works. -> https://test.visualize.admin.ch/de/v/yof4TUv0bmCU

Just for added information

Rdataflow:
just for the records:
fetching the geometries in SPARQL is fast - the bottleneck might be somewhere else...
https://s.zazuko.com/2iRjLUE

PR #1129, opened 2023-08-18
fix: Infinite geographical shapes loading

bprusinowski:
When testing #1106, I noticed that some cubes still load the data for quite long time (e.g. [this Elcom cube](https://test.visualize.admin.ch/en/browse?previous=%7B%22includeDrafts%22%3Atrue%2C%22order%22%3A%22SCORE%22%2C%22search%22%3A%22elcom%22%7D&dataset=https%3A%2F%2Fenergy.ld.admin.ch%2Felcom%2Felectricityprice&dataSource=Prod)).

After investigation, it turned out that we wrongly handle potential undefined geometries when fetching the shapes. This PR fixes that (and also removes a case of infinite loading spinner when there is an error during geo shapes fetching – now an empty map will be shown).

cc @sosiology

### DIFF
--- a/app/charts/map/chart-map.tsx
+++ b/app/charts/map/chart-map.tsx
@@ -112,7 +112,6 @@ export const ChartMapVisualization = ({
 
   const ready =
     (areaDimensionIri !== "" &&
-      // Check if original, unfiltered number of shapes is bigger than 0.
       (shapes?.topology?.objects?.shapes as any)?.geometries) ||
     (symbolDimensionIri !== "" && coordinates) ||
     (shapes?.topology?.objects?.shapes as any)?.geometries ||
--- a/app/charts/map/chart-map.tsx
+++ b/app/charts/map/chart-map.tsx
@@ -113,9 +113,9 @@ export const ChartMapVisualization = ({
   const ready =
     (areaDimensionIri !== "" &&
       // Check if original, unfiltered number of shapes is bigger than 0.
-      (shapes?.topology?.objects?.shapes as any)?.geometries?.length) ||
-    (symbolDimensionIri !== "" && coordinates?.length) ||
-    (shapes?.topology?.objects?.shapes as any)?.geometries?.length ||
+      (shapes?.topology?.objects?.shapes as any)?.geometries) ||
+    (symbolDimensionIri !== "" && coordinates) ||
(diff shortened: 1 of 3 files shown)


ortnever:
Hi @bprusinowski , Is it already fixed on int? Because I have the same problem with : 
https://int.visualize.admin.ch/fr/v/gZCYVF8rylJF
I haven't been able to display the graph since this week, although it was working perfectly before my holiday.

bprusinowski:
Hi @ortnever, it is not yet on INT, it will soon be deployed to TEST, but I think the problem is different with the Heavy Metal Soil Contamination cube.

We are not detecting that the `Station` dimension is a geographical dimension. I checked the cube in Cube Creator and it is marked as `NominalDimension` without a data kind of `Geographic shape`.

I assume something changed when the cube was last updated, because it seems that at some point it was marked correctly (as it was possible to create a map chart visible in the link you shared).

I would advise to update the data kind of the Station dimension, and then it should work correctly (now it's not even possible to select a map chart type when creating a new chart with this dataset).

Screenshot 2023-08-18 at 11 38 15


Let me know if this fixes the problem 🙇

ortnever:
The URL I shared above concerns an other cube. And this cube has geographical Dimension (Municipalities).
image
image

Or did I understand something wrong?

ortnever:
image

bprusinowski:
@ortnever it looks like the municipalities from this cube do not implement `http://www.opengis.net/ont/geosparql#hasGeometry` property.

I ran the query against int.lindas.ch (https://s.zazuko.com/3hwRpTt) and it returns no results, but when ran against test.lindas.ch (https://s.zazuko.com/hEmYZC) it correctly returns values (and as the cube is published on INT, we query the INT LINDAS, which returns no results).

What's strange here is that the second query returns different iris than the one passed to the VALUES `<https://test.lindas.admin.ch/municipality/6808>` instead of `<https://ld.admin.ch/municipality/6808>`.

I am not sure what causes this error, but I guess it's related to some linking of values on LINDAS?

Rdataflow:
missing geometry is related to LINDAS.
@bprusinowski can you show a error message in case geometries are missing? current behaviour is to show the map w/o error is misleading for the user... one might think toggling some more buttons might help - which isn't the case.

PR #1130, opened 2023-08-18
feat: Cascading interactive filters

### LINKED ISSUE #566: Cascading filters in publishing mode
This is probably an improvement and not a bug.
Currently cascading filters only work in edit mode.  Cascading filters should also work in publish mode if possible. To be discussed depending on the investment needed.

see Gitlab https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/511

### LINKED ISSUE #1094: Improvement to Interactive Filters in Published Mode
In Edit mode, selecting a value in a data filter restricts the choice of values in the other filters in order to avoid selecting combinations for which no data exists ("Cascading Filter" behaviour).

As a user of a chart with several interactive filters, I would like the content of the other filters to adapt when I have selected a value in one of the filters.

Exqample : https://test.visualize.admin.ch/fr/v/xAVhqoA3PtsH

See Issue 566 in Github : Cascading filters in publishing mode https://github.com/visualize-admin/visualization-tool/issues/566

### LINKED ISSUE #1149: Display of absolute values in % in a bar chart - Small error in the display of the vertical axis
When using the "Display absolute values in % in a bar chart" function, 100% is not displayed correctly on the y axis (00% instead of 100%).
See the chart :  https://int.visualize.admin.ch/fr/v/JEOzXRikOWZy
image

bprusinowski:
Closes #1094.
Closes #566.
Closes #1149.

This PR:
- adds cascading mode to interactive filters,
- makes the date picker usable as interactive filter, instead a select or text input element (when possible),
- adds `ChartLoadingState` which consolidates the loading logic per chart. This fixes a problem of `<NoDataHint>` being displayed when interactive filters are reloading,
- keeps the order of filters by storing them inside a database (so what's displayed in editor mode matches what's published).

Chart loading state could be further improved in some way, as e.g. it doesn't apply to regular filters, because they exist in `ChartConfigurator`, which is way above the `Chart` component. I didn't want to elevate this logic up, as when dashboards come, we shouldn't block every chart when filters are being loaded / re-loaded. We would need to think about a solution once we start tackling the dashboards :)


app/charts/chart-loading-wrapper.tsx
```diff
@@ -83,19 +92,20 @@ export const ChartLoadingWrapper = <
   if (metadata && dimensions && measures && observations) {
     const { title } = metadata;
 
-    return observations.length > 0 ? (
+    return (
       <Box
-        data-chart-loaded={
-          !(fetchingMetadata && fetchingComponents && fetchingObservations)
-        }
+        data-chart-loaded={!chartLoadingState.loading}
```
ptbrowne, line 97: I wonder if we could not simplify the chartLoadingState logic by just having a context provider at the ChartLoadingWrapper level. This way we could remove the chart state observable ?

```jsx
<ChartLoadingProvider value={{ fetching }}>
      <Box
        data-chart-loaded={!chartLoadingState.loading}
		...
	...
</ChartLoadingProvider>
```

bprusinowski: I also thought about this, but unfortunately we need to access the loading state "above" the chart, in interactive filters :( That's why I've moved it to the nearest closest parent of both. Tell me if I miss something 😅

PR #1147, opened 2023-08-25
feat: Disable stacking of non-ratio measures (Area & Column)

### LINKED ISSUE #1132: Interval scales and normalisation
See [here](https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/529). In short, as stacking of interval scales doesn't make sense, so wouldn't a normalisation of such charts make sense. Take again the example of temperature in degree celsius. What would 100% mean?

A normalisation should only work for ratio scales.

**Usecases and impact**
One could prevent the generation of non-sense charts. I'm really hard-pressed to come up with a proper interval scale, where it makes sense to stack/and or normalise the measurements. And if visualize would distinguish between ratio and interval scale, one could prevent such cases.

### LINKED ISSUE #1140: Segmentation: distinction between `qudt:IntervalScale` and `qudt:RatioScale`

bprusinowski:
Closes #1140.
Closes #1132.

This PR:
- disables stacking of non-ratio measures (for area and column charts),
- fixes missing importation warning message.

### How to test
123

### DIFF
--- a/app/charts/chart-config-ui-options.ts
+++ b/app/charts/chart-config-ui-options.ts
@@ -262,6 +262,10 @@ const isMissingDataPresent = (chartConfig: AreaConfig, data: Observation[]) => {
   return checkForMissingValuesInSegments(grouped, segments);
 };
 
+export const disableStacked = (d?: DimensionMetadataFragment): boolean => {
+  return d?.scaleType !== "Ratio";
+};
+
 export const chartConfigOptionsUISpec: ChartSpecs = {
   area: {
     chartType: "area",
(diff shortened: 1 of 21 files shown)


app/locales/fr/messages.po
```diff
@@ -623,6 +620,10 @@ msgstr "Ajoutez un titre et une description"
 msgid "controls.segment"
 msgstr "Segmentation"
 
+#: app/charts/chart-config-ui-options.ts
+msgid "controls.segment.stacked.disabled-by-scale-type"
+msgstr "La mise en page empilée ne peut être activée que si la dimension mappée sur l'axe vertical a une échelle de rapport."
```
ptbrowne, line 625: msgstr "Le graphique en barres empilées ne peut être activé que si la dimension de l'axe vertical a une échelle de type ratio."

app/charts/chart-config-ui-options.ts
```diff
@@ -284,8 +288,23 @@ export const chartConfigOptionsUISpec: ChartSpecs = {
...
+          const yIri = chartConfig.fields.y.componentIri;
+          const yDimension = components.find((d) => d.iri === yIri);
+          const disabledStacked = disableStacked(yDimension);
+
+          if (disabledStacked) {
+            return {
+              disabled: true,
+              warnMessage: t({
+                id: "controls.segment.stacked.disabled-by-scale-type",
+                message:
```
ptbrowne, line 302: Stacked layout can only be enabled if the vertical axis dimension has a ratio scale.

PR #1155, opened 2023-08-29
refactor: Configurator state side effects

bprusinowski:
This PR extracts the side effect of changing a field or one of its options to the UI encodings. I believe this should make it much more clear what happens when a given option is changed.

The reason for this change was that currently, the logic to create chart option selectors, and what happens when they change, is hugely separated, which made it difficult to handle changes to the app in the past, as unexpected things would happen if you forgot to check / add a new side effect in the configurator state – now it's all consolidated in UI encodings.

So, instead of having:
- UI encodings defined in `/charts/config-ui-options`,
- rendering of UI encodings in `/configurator/components/chart-options-selector`,
- side effects logic defined in `/configurator/configurator-state`,

we have:
- UI encodings and side effects defined in  `/charts/config-ui-options`,
- rendering of UI encodings in `/configurator/components/chart-options-selector`.

Configurator state reducer calls `getChartFieldChangeSideEffect` or `getChartFieldOptionChangeSideEffect` respectively when field or option changes.


app/charts/chart-config-ui-options.ts
```diff
@@ -91,6 +86,46 @@ export type EncodingOption =
...
     };
 
+const makeOnColorComponentChange = (type: "areaLayer" | "symbolLayer") => {
+  return (
+    draft: ConfiguratorStateConfiguringChart,
+    action: {
+      type: "scaleType";
+      value: "continuous" | "discrete";
+    }
+  ) => {
```
ptbrowne: Since the sideEffect is receiving the field, couldn't we just get the field type at runtime ?

```
const colorComponentChange = (draft, action, field) => {
  const type = field.type
....
}
```

ptbrowne:
👍 Really nice change. It should reduce the number of touched files when introducing a new field / feature, that's really nice 💯

bprusinowski:
Thanks for reviewing @ptbrowne! Yes, that should be done like you suggested, I'll make the change 👍

PR #1159, opened 2023-08-31
feat: Dashboards I

bprusinowski:
This PR contains the initial work to prepare for having multiple charts at once.

See [internal documentation](https://www.notion.so/interactivethings/Dashboards-15503c1027634755b43d28c60624b031) for more context.


app/configurator/configurator-state.tsx
```diff
@@ -573,29 +594,27 @@ const transitionStepNext = (
...
 
         return {
+          version: CONFIGURATOR_STATE_VERSION,
           state: "CONFIGURING_CHART",
           dataSet: draft.dataSet,
           dataSource: draft.dataSource,
           meta: draft.meta,
-          activeField: undefined,
-          chartConfig,
+          chartConfigs: [chartConfig],
```
ptbrowne, line 610: Can you try without replacing the array here ? 

chartConfigs.splice(0, chartConfig)

chartConfigs: chartConfigs

ptbrowne: Maybe

```
chartConfigs: chartConfigs.map(config => {
  if(config.key === chartConfig.key) { return chartConfig }
  return config
})
```

bprusinowski: Here the problem was that getter and setter were lost quite quickly, I think because of some spread pattern down the line, so this approach didn't work very well :( I think it's maybe even because of Immer, when it creates a copy of the object, we lose getters and setters

ptbrowne: Ah I see, and did you try without the getter / setter, but still with immer ?

app/configurator/configurator-state.tsx
```diff
@@ -888,58 +913,54 @@ const reducer: Reducer<ConfiguratorState, ConfiguratorStateAction> = (
...
             dimensions,
             measures,
           });
-          draft.activeField = undefined;
-          draft.chartConfig = deriveFiltersFromFields(
-            draft.chartConfig,
-            metadata.dimensions
-          );
+
+          for (const k in previousConfig) {
```
bprusinowski: @ptbrowne it kinda worked with this, when I used the` draft.chartsConfigs[index] = newChartConfig`, there were some Proxy issues. I think it should be the same as when doing splice 🤔

app/configurator/configurator-state.tsx
bprusinowski: It turned out to work ok with replacing the config at given index. I guess the problem I had previously was related to something else :D

bprusinowski:
Hey @ptbrowne 👋 I'll clean up some things a bit in the coming days, but in case you have some time, it would be great to have a review anyways, as the fundamental things shouldn't change now 🙏

app/utils/chart-config/api.ts
```diff
@@ -11,6 +11,7 @@ export const createConfig = async (state: ConfiguratorStatePublishing) => {
     method: "POST",
     data: {
       data: {
+        version: state.version,
         dataSet: state.dataSet,
         dataSource: state.dataSource,
         meta: state.meta,
```
ptbrowne, line 17: How come the version was not saved ? How could migration work then ?

app/configurator/configurator-state.tsx
```diff
@@ -1455,6 +1455,7 @@ const ConfiguratorStateProviderInternal = ({
                     };
                   }),
                 ],
+                activeChartKey: state.chartConfigs[0].key,
```
ptbrowne, line 1494: It seemed to me that the activeChartKey was only used for edition mode, why do we need to save it in the database ?

app/utils/chart-config/api.ts
bprusinowski: Previously we only had versions of chart config, and not configurator state – now we also keep versions of it, so we have two migrations (config state + chart configs).

app/configurator/configurator-state.tsx
bprusinowski: This was mostly done to keep it simpler to maintain the configurator state, as otherwise I think I needed to extract the active key when retrieving the config from database (and we have a clash between "state saved to db" type and "state before saving to db" types). But it's true that it's not necessary to store it in db, I'll remove it 👍

ptbrowne: It was just a question, no need to remove it if it makes things a little bit simpler, maybe a comment is sufficient ;)

bprusinowski: Right – I'll add a comment then 👌

ptbrowne (approved):
LGTM 👏 👏

PR #1204, opened 2023-10-06
chore: Add time measuring to SPARQLClientStream and CONSTRUCT queries

bprusinowski:


Rdataflow:
@bprusinowski hooray :tada: this already makes `cube-hierarchy-query` queries and geometries fully transparent :smile: now only rdf-cube-view-query remains to takle ... 

cc @adintegra

bprusinowski:
Hi @Rdataflow, yes, it looks that now we only miss `cube-view-query` timings ⏲️

PR #1206, opened 2023-10-06
docs: Describe chart preview via API + expose DebugPanel via flag

### LINKED ISSUE #1200: Update Documentation: Parameteres to call up chart (API)
Document as described in the comment here: https://github.com/visualize-admin/visualization-tool/pull/1152#issuecomment-1733172848

bprusinowski:
Closes #1200.

As the types can differ a lot between different chart types, and there are some hidden dependencies to some properties (e.g. multi filter connected to temporal X field, some sorting options only available for column charts, segmentation not available for maps, etc, etc), I decided to point to the file with types in the docs and **expose the configurator & interactive filters debug panel via flag**.

By doing this, editors will be able to edit a chart, tailor it to their needs, and dump the state to the console to copy and modify it slightly to fit their needs. With this we don't need to prepare a complete list of all combination possibilities between the types, which would be really difficult, but rather encourage to pre-create a chart and edit it afterwards, making sure that all necessary side effects have been applied.

### DIFF
--- a/app/docs/chart-preview-via-api.mdx
+++ b/app/docs/chart-preview-via-api.mdx
@@ -1,0 +1,160 @@
+<style>{`
+  table:not([class]) {
+     margin-top: 1rem;
+     font-size: 0.875rem;
+     cell-spacing: none;
+     border-spacing: 0;
+     border-collapse: collapse;
+  }
+
+  table:not([class]) tr:nth-child(2n) {
+    background: #eee;
+  }
+
+  table:not([class]) td, table:not([class]) th {
+    border-bottom: #ccc 1px solid;
+    margin-top: 0;
+    padding: 0.25rem 0.5rem;
+  }
+
+  table:not([class]) tr {
+    margin-bottom: 0;
+  }
+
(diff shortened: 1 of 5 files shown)


app/docs/chart-preview-via-api.mdx
```diff
@@ -0,0 +1,160 @@
...
+          "https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/AnzahlAnlagen",
+      },
+    },
+  },
+}`}
+/>
+
+Note that it's encouraged to visit the Visualize application with &flag_debug==true added to the URL to enable
+the debug mode, which will show the chart config directly below the chart that is being edited. It also enables
+the "Dump to console" button, which will log the chart config to the browser console, for easier re-use.
```
Rdataflow, line 160: @bprusinowski typo? - I guess `&flag__debug=true`

bprusinowski: Yes, thanks for catching 🙇

PR #1208, opened 2023-10-10
feat: Allow edition of published charts

bprusinowski:
This PR allows the edition of published charts.

- ~Configurator state key is now stored inside the config~
- During publishing, we check if user is logged in and if a chart of given id exists in the database; if so, we update the chart config, assuring that they can only update the charts they created (client side and API side guards)
- New `?edit=configuratorStateKey` initializes chart state with `configuratorStateKey ` = `edit mode`
- Old `?from=configuratorStateKey` initializes chart state with a new chart id = `new chart mode`

### DIFF
--- a/app/components/chart-selection-tabs.tsx
+++ b/app/components/chart-selection-tabs.tsx
@@ -29,6 +29,7 @@ import { fetchChartConfig } from "@/utils/chart-config/api";
 import { createChartId } from "@/utils/create-chart-id";
 import { getRouterChartId } from "@/utils/router/helpers";
 import useEvent from "@/utils/use-event";
+import { useFetchData } from "@/utils/use-fetch-data";
 
 type TabsState = {
   popoverOpen: boolean;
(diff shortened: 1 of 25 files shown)


app/components/chart-selection-tabs.tsx
```diff
@@ -277,16 +281,58 @@ const PublishChartButton = () => {
...
+        }
+      }
+
+      if (session.status !== "loading") {
+        setLoaded(true);
+      }
+    };
+
+    run();
+  }, [chartId, session]);
```
ptbrowne, line 307: * It seems like the editingPublishedChart could be computed.
* I think it would be a bit more readable if we could get rid of the effect and use something similar to useQuery from react-query

```
// Not tested (asked chatgpt but it looked good)
const useFetchData = ({ queryFn, enabled = true } = {}) => {
  const [data, setData] = useState(null);
  const [error, setError] = useState(null);
  const [fetchStatus, setFetchStatus] = useState('idle'); // Possible values: 'idle', 'fetching', 'success', 'error'

  useEffect(() => {
    if (!enabled) {
      return;
    }

    const fetchData = async () => {
      setFetchStatus('fetching');
      try {
        const result = await queryFn();
        setData(result);
        setFetchStatus('success');
      } catch (err) {
        setError(err);
        setFetchStatus('error');
      }
    };

    fetchData();
  }, [enabled, queryFn]);

  return { data, error, fetchStatus };
};


// Then in your code
  const { asPath } = useRouter();
  const session = useSession();
  const { loaded, data } = useFetchData({ queryFn: fetchChartData, enabled: session.data.user && chartId })
  
    const editingPublishedChart = session.data.user.id && config?.user_id === session.data.user.id
```

I feel like using the same pattern as react-query makes it closer to declarative code, and easier to read, what do you think ?

ptbrowne: It would prevent also infinite loading when the fetch fails ;)

bprusinowski: Yup, good point! Thanks for taking a look and making a proposal, I'll adjust the code 👌

ptbrowne (approved):

Issue #1190, opened 2023-09-22
Login Feature


sosiology:
- [ ] Align with Patrick @bprusinowski 
- [x] Define Scope for this Sprint @bprusinowski @adintegra @sosiology 
- [x ] Refine Estimate @sosiology 
- [ ] UI Implementation @bprusinowski 

**DoD**
User can login with test credentials and create and edit charts with a stable URL

**Design**
https://projects.invisionapp.com/share/PD12Y05GFZ6V#/screens/470445250

sosiology:
Blocked as we wait for feedback from VSHN


------

Michel Rios has just commented SBAR-750 / Create account on Keycloak for visualize admins | Michel Rios has just commented SBAR-750 / Create account on Keycloak for visualize admins
-- | --
Michel Rios has just commented SBAR-750 / Create account on Keycloak for visualize admins

Hi Mark

Thank you for your message.

We'll discuss the topic at latest in our planning meeting and propose you slots for the meeting afterwards.

I wish you a nice rest of the week.

Kind regards
Michel Rios

PR #1210, opened 2023-10-11
feat: Login improvements

bprusinowski:
Contributes towards #1190.

This PR **adds**:
- Published date to the table
- Actions tooltip
- Copy chart option
- Delete chart option (with a modal to confirm the action)
- Link to dataset (tied to its title)
- Sorting of user configs by date
- Preview mode to table (homepage, 3 results)

It also **fixes** a problem of not extracting components iris for combo charts and not carrying over the locale when clicking `Browse all datasets` button.

### DIFF
--- a/app/login/components/profile-content-tabs.tsx
+++ b/app/login/components/profile-content-tabs.tsx
@@ -1,3 +1,4 @@
+import { t } from "@lingui/macro";
 import { TabContext, TabList, TabPanel } from "@mui/lab";
 import { Box, Tab, Theme } from "@mui/material";
 import { makeStyles } from "@mui/styles";
@@ -55,7 +56,13 @@ export const ProfileContentTabs = (props: ProfileContentTabsProps) => {
         <Box className={rootClasses.sectionContent}>
           <Box sx={{ borderBottom: 1, borderColor: "divider" }}>
             <TabList className={classes.tabList} onChange={handleChange}>
-              {["Home", "My visualizations"].map((d) => (
+              {[
+                "Home",
+                t({
+                  id: "login.profile.my-visualizations",
+                  message: "My visualizations",
+                }),
+              ].map((d) => (
                 <Tab key={d} className={classes.tab} label={d} value={d} />
               ))}
             </TabList>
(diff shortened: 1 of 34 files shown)


app/login/components/profile-tables.tsx
```diff
@@ -35,10 +35,22 @@ export const ProfileVisualizationsTable = (
 
   return (
     <Box className={rootClasses.sectionContent}>
-      <Typography variant="h2">My visualizations</Typography>
+      <Typography variant="h2" sx={{ mb: 4 }}>
+        My visualizations
+      </Typography>
+      <Typography variant="h3" sx={{ mb: 2 }}>
+        My published visualizations
+      </Typography>
```
ptbrowne: You might need translations here ?

app/login/components/profile-content-tabs.tsx
```diff
@@ -35,11 +35,13 @@ const useStyles = makeStyles<Theme>((theme) => ({
...
 
 type ProfileContentTabsProps = {
+  userId: number;
   userConfigs: ParsedConfig[];
 };
 
 export const ProfileContentTabs = (props: ProfileContentTabsProps) => {
-  const { userConfigs } = props;
+  const { userId, userConfigs: _userConfigs } = props;
+  const [userConfigs, setUserConfigs] = React.useState(_userConfigs);
```
ptbrowne, line 45: Not sure I understand the change 🤔 ?

ptbrowne (approved):
Looks great !

app/login/components/profile-tables.tsx
bprusinowski: Yup, added them in following commit 🇩🇪 🇫🇷 🇮🇹

app/login/components/profile-content-tabs.tsx
bprusinowski: Ah, this commit was a preparation for a next one, probably they should be one – as we fetch the user configs on the server side, there was a need to `useState` so that when user removes a chart, we can simply handle the change on the frontend, without re-querying configs (+ to make the table re-render once it's removed).

PR #1218, opened 2023-10-13
performance: Use CONSTRUCT query for dimension values metadata

bprusinowski:
This PR aims at improving the performance of dimension values fetching by replacing several SELECT queries with one CONSTRUCT query.

### Observations
I compared the performance of the old and new way of fetching and included the results below. I based these measurements on four datasets:
- Einmalvergütung für Photovoltaikanlagen,
- Gebäudeprogramm - CO2-Wirkungen je Massnahmenbereich,
- State accounts - Function,
- NFI: Topics by stage of stand development,

opening each one in the editing mode and triggering fetching of dimension metadata values the same number of times for each dataset.

---

| Key   | Value |
|---------------------|-----|
| Number of queries   | 510 |
| Avg. time SELECT    | 242ms |
| Avg. time CONSTRUCT | 249ms (+3%) |

I've also removed the conditions to limit inclusion of some properties based on scale type, as I am not 100% sure if they are needed anymore (e.g. if dimension values description can really only be present when scale type is either ordinal or nominal). Also, when they were added to the data, I think it makes sense to include them even when scale type is mapped wrongly. When they were removed, the performance of CONSTRUCT queries based on 406 new queries was better than SELECT queries by 0.04%.

cc @adintegra @RDataFlow

### DIFF
--- a/app/next.config.js
+++ b/app/next.config.js
@@ -99,6 +99,11 @@ module.exports = withPreconstruct(
           new IgnorePlugin({ resourceRegExp: /^(pg-native|vue)$/ })
         );
 
+        config.resolve.fallback = {
+          ...config.resolve.fallback,
+          fs: false,
+        };
+
         return config;
       },
     })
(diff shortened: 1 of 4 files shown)


bprusinowski:
Closed in favor of #1378.

PR #1228, opened 2023-10-19
perf: Retrieve (min|max)Inclusive values stored behind sh:or

### LINKED ISSUE #1219: perf: also consider dimensions with `{min,max}Inclusive` according to new solution provided by zazuko
*Userstory*
 - As a data provider I'd like visualize to reuse conforming `{min,max}Inclusive` values from the shape (instead of firing another slow query per dimension to lindas)

*DoD*
- [ ] visualize also considers `{min,max}Inclusive` values also according to the new solution (within `rdf:List` behind `sh:or`)

see solution described in https://gitlab.ldbar.ch/bafu/visualize/-/issues/537#note_13077

thus there exist two official places where sh:{min,max}Inclusive is to expect:
1. directly on the dimension (implemented in visualize)
    ```
    <dim> sh:minInclusive ?min .
    ```
2. behind sh:or packed in a rdf:List (to be implemented yet in visualize)
    ```
    <dim> sh:or ( [..., sh:minInclusive ?min, ...] )
    ```

the second case can thus be queried traversing the properties:
```
<dim> sh:or/rdf:rest*/rdf:first/sh:minInclusive ?min .
```

see also code at https://github.com/visualize-admin/visualization-tool/blob/598b756ef7a9365f6d231b332aef2a8a93738f7b/app/rdf/queries.ts#L227-L228

see data samples at https://s.zazuko.com/phdBiS

cc  @kronmar

bprusinowski:
Closes #1219.

This PR includes retrieval or min and max values stored behind `sh:or`.

### DIFF
--- a/app/rdf/queries.ts
+++ b/app/rdf/queries.ts
@@ -232,16 +232,18 @@ export const getCubeDimensionValues = async ({
       ];
     }
 
-    const firstPointer = dimension.out(ns.sh.or).out(ns.rdf.first);
-    const firstMin = firstPointer.out(ns.sh.minInclusive);
-    const firstMax = firstPointer.out(ns.sh.maxInclusive);
+    const listPointer = dimension
+      .out(ns.sh.or)
+      .out([ns.rdf.first, ns.rdf.rest]);
+    const listMin = listPointer.out(ns.sh.minInclusive);
+    const listMax = listPointer.out(ns.sh.maxInclusive);
 
     if (
-      typeof firstMin.value !== "undefined" &&
-      typeof firstMax.value !== "undefined"
+      typeof listMin.value !== "undefined" &&
+      typeof listMax.value !== "undefined"
     ) {
-      const min = +firstMin.value;
-      const max = +firstMax.value;
+      const min = +listMin.value;
+      const max = +listMax.value;
 
       return [
         { value: min, label: `${min}` },
--- a/app/rdf/queries.ts
+++ b/app/rdf/queries.ts
@@ -218,19 +218,36 @@ export const getCubeDimensionValues = async ({
 }): Promise<DimensionValue[]> => {
   const { dimension, cube, locale, data } = rdimension;
 
-  if (
-    typeof dimension.minInclusive !== "undefined" &&
-    typeof dimension.maxInclusive !== "undefined" &&
-    data.dataKind !== "Time" &&
-    data.scaleType !== "Ordinal"
-  ) {
-    const min = parseObservationValue({ value: dimension.minInclusive }) ?? 0;
-    const max = parseObservationValue({ value: dimension.maxInclusive }) ?? 0;
+  if (data.dataKind !== "Time" && data.scaleType !== "Ordinal") {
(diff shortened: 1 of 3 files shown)


Rdataflow:
@kronmar can you try this and give your feedback? THX

app/rdf/queries.ts
```diff
@@ -218,19 +218,36 @@ export const getCubeDimensionValues = async ({
...
+      const min = parseObservationValue({ value: dimension.minInclusive }) ?? 0;
+      const max = parseObservationValue({ value: dimension.maxInclusive }) ?? 0;
+
+      return [
+        { value: min, label: `${min}` },
+        { value: max, label: `${max}` },
+      ];
+    }
+
+    const firstPointer = dimension.out(ns.sh.or).out(ns.rdf.first);
```
Rdataflow, line 235: @bprusinowski just a first quick question:
would this also retrieve the second or third element in a list behind sh:or?

asking because I didn't see the part
 `rdf:rest*/rdf:first`

bprusinowski: @Rdataflow thanks for taking a look at this. I am not sure how to properly test this, but you're right that right now we only check for the first element in the list. I think I fixed the issue now by also considering `ns.rdf.rest` predicate (I am not sure if this will catch remaining elements, but the results for [this NFI dataset](https://test.visualize.admin.ch/en/create/new?cube=https://environment.ld.admin.ch/foen/nfi/nfi_C-2171/cube/2023-6) are the same in both approaches).

Rdataflow: @bprusinowski FYI rdf:List behaves like this image

thus the goal it to retrieve
- `sh:or`
   - then step over **0, 1 or even many** `rdf:rest`\*
     - to get to the `rdf:first`
       - finally containing the relevant sh:minInclusive / sh:maxInclusive

I'm not sure about the effect of the array in this piece of code though / tom+ might know best
```
      .out(ns.sh.or)
      .out([ns.rdf.first, ns.rdf.rest]);
```

bprusinowski: I see, and is it guaranteed that there is only one min/max value across the whole list (so we can stop traversal as soon as there is a value found)? I am wondering why would there be several entries in this list, and assuming there are e.g. 10 entries, if 3rd has these values, what's in the remaining 7? I think there should only be one min/max values, attached directly either to dimension, or `sh:first`?

Rdataflow: @bprusinowski yes, there must be only 1 entry for minInclusive / maxInclusive property within the same List. go with this assumption. 👍 
- this needs to be guaranteed by the cube validator... see https;://cube.link/#validate-the-cube (still WIP as of now)

fyi: there might be other entries on datatype cube:Undefined and what more is or will be...

Rdataflow: LGTM 👍 
@kronmar can you give it a try?

bprusinowski:
@Rdataflow @kronmar I'll merge to TEST, it could be tested there afterwards 👍

PR #1237, opened 2023-10-27
fix: Dates with timezones are now correctly parsed

### LINKED ISSUE #1233: An xsd:dateTime dimension displays "-" when values have a timezone as "+01:00"
**Describe the bug**
We have two similar cubes for the "Waldbrand" dataset.
Those cubes have a schema:validFrom dimension with a xsd:dateTime.
That dimension metadata has been defined with the following values, including sh:datatype and meta:dataKind:
```
        [ schema:name "Aktiv seit"@de, "Active since"@en, "Actif depuis"@fr, "Attivo dal"@it ;
            qudt:scaleType qudt:IntervalScale ;
            sh:nodeKind sh:Literal ;
            sh:datatype xsd:dateTime ;
            meta:dataKind [a time:GeneralDateTimeDescription ;
                time:unitType time:unitSecond] ;
            sh:path schema:validFrom ],
```

The RDF Cube Schema gives an example here: https://zazuko.github.io/cube-link/#example-10

The problem:
We are trying this out on test.visualize, the datat being on Lindas TEST.
Just do a search on "waldbrand" to find both cubes

Visualize displays the validFrom as an empty value:
image

On that screen, clicking on "Run SPARQL Query", will show the underlying values:
image

I suspect that the problem comes from the xsd:dateTime with a time zone specified as "+01:00" or "+02:00".
But this should be a valid xsd:dateTime IMHO.
In the pipeline, if I truncate the values without this timezone, that Visualize seems fine.

- Visualize environment and version: TEST v3.23.1
- Browser and version: Firefox 118.0.2

bprusinowski:
Fixes #1233.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ You can also check the [release page](https://github.com/visualize-admin/visuali
   - Search string is now correctly persisted in search box when refreshing the page (Browse page)
   - Fixed issue with two queries being sent when refreshing the page when search string was entered (Browse page)
   - Fixed issue with filtering / unfiltering subthemes that resulted in 404 error (Browse page)
+  - Dates with timezones are now correctly parsed
 - Performance
   - Improved performance of searching for and retrieving datasets (Browse page)
   - Improved the performance of data download
--- a/app/formatters.ts
+++ b/app/formatters.ts
@@ -100,14 +100,19 @@ export const dateFormatterFromDimension = (
         dim.timeUnit.toLowerCase() as keyof typeof localFormatters
       ];
     const parser = timeParse(dim.timeFormat);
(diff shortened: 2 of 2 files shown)


Rdataflow:
@bprusinowski well done
- [ ] one minor observation I'd like you to fix:
    - in edit mode table view clicking on the dimension `Active since` I get the metadata w/ timezone whereas in the table I get the timestamps in local time.
         ```
         Available values
         Min: 2022-08-18T14:08:43+02:00
         Max: 2023-09-25T11:19:08+02:00

bprusinowski:
@Rdataflow see https://github.com/visualize-admin/visualization-tool/pull/1242

PR #1241, opened 2023-11-01
style: Combo charts UI alignments

bprusinowski:
This PR updates the Combo charts-related styles along with several other styling adjustments.

- Removed text input from `ColorPicker`
- Removed axis orientation radio buttons from `ComboColumnLine` chart
- Added `OpenMetadataPanelWrapper` to legend items in `ComboLineSingle` chart
- Updated UI strings for left / right axes (`ComboColumnLine` chart)

### DIFF
--- a/app/charts/combo/chart-combo-line-single.tsx
+++ b/app/charts/combo/chart-combo-line-single.tsx
@@ -83,6 +83,13 @@ export const ChartComboLineSingle = React.memo(
     const { chartConfig, measures } = props;
     const { interactiveFiltersConfig } = chartConfig;
 
+    const getLegendItemDimension = React.useCallback(
+      (label) => {
+        return measures.find((measure) => measure.label === label);
+      },
+      [measures]
+    );
+
     return (
       <ComboLineSingleChart aspectRatio={0.4} {...props}>
         <ChartContainer>
@@ -100,9 +107,7 @@ export const ChartComboLineSingle = React.memo(
           <LegendColor
             chartConfig={chartConfig}
             symbol="line"
-            getLegendItemDimension={(label) => {
-              return measures.find((measure) => measure.label === label);
-            }}
+            getLegendItemDimension={getLegendItemDimension}
           />
         </ChartControlsContainer>
       </ComboLineSingleChart>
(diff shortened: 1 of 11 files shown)


app/charts/combo/chart-combo-line-single.tsx
```diff
@@ -97,7 +97,13 @@ export const ChartComboLineSingle = React.memo(
           <Tooltip type="multiple" />
         </ChartContainer>
         <ChartControlsContainer>
-          <LegendColor chartConfig={chartConfig} symbol="line" />
+          <LegendColor
+            chartConfig={chartConfig}
+            symbol="line"
+            getLegendItemDimension={(label) => {
+              return measures.find((measure) => measure.label === label);
+            }}
```
lloydrichards, line 105: Should this be memoized? Im assuming that the this will be re-rendered with every change to the chart or config?

app/configurator/components/chart-controls/color-picker.tsx
```diff
@@ -75,45 +67,22 @@ const useColorPickerStyles = makeStyles((theme: Theme) => ({
...
 }));
 
 export const ColorPicker = ({ selectedColor, colors, onChange }: Props) => {
-  const [inputColorValue, setInputColorValue] = useState(selectedColor);
-
+  const classes = useColorPickerStyles();
   const selectColor = useCallback(
     (_color) => {
-      setInputColorValue(_color);
       // Make sure onChange is only called with valid colors
```
lloydrichards, line 76: Is there no validation errors incase this fails?

lloydrichards (approved):
LGTM.  Just a nit and comment that shouldn't impact anything major 👍

app/configurator/components/chart-controls/color-picker.tsx
bprusinowski: Before this PR, we had a text input to specify our own color – if it was invalid, it simply didn't update the state. Here we removed the text input completely to only allow the color from the active color palette, selectable via color swatch buttons, so they will always be correct 👀

app/charts/combo/chart-combo-line-single.tsx
bprusinowski: You mean `getLegendItemDimension` (with `useCallback`)? I can add this 👍 The `LegendColor` function looks to be already memoized 👀

app/configurator/components/chart-controls/color-picker.tsx
lloydrichards: okay, so then by not just call the `d3Color()` directly instead of creating the callback?  doesnt seem like it does anything anymore 🤷‍♂️

bprusinowski: Good point!

PR #1223, opened 2023-10-17
chore: Update hierarchy query library

bprusinowski:
This PR updates the cube-hierarchy-query library to v2.2.0.

It also updates the way of how we fetch dimension values metadata for hierarchies by using new syntax to directly specify the properties we need to fetch along the hierarchy. However, this seem to make it significantly less performant that our current approach, which means we should probably only update library, but keep the old way of fetching things.

**New auto-generated query used to fetch hierarchies and metadata ([NFI dataset](https://test.visualize.admin.ch/browse?dataset=https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fnfi%2Fnfi_C-90%2Fcube%2F2023-2&dataSource=Prod)) – approx. 15s**
```
PREFIX schema: <http://schema.org/>
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>

CONSTRUCT { 
  ?resource1 schema:identifier ?resource2 .
  ?resource1 schema:name ?resource3 .
  ?resource1 schema:description ?resource5 .
  ?resource1 schema:position ?resource7 .
  ?resource1 schema:alternateName ?resource8 .
  ?resource1 rdf:type ?resource10 .
  ?resource11 schema:containedInPlace ?resource1 .
  ?resource11 schema:identifier ?resource12 .
  ?resource11 schema:name ?resource13 .
  ?resource11 schema:description ?resource15 .
  ?resource11 schema:position ?resource17 .
  ?resource11 schema:alternateName ?resource18 .
  ?resource11 rdf:type ?resource20 .
}
WHERE {
  VALUES ( ?resource1 ) {
    ( <https://ld.admin.ch/country/CHE> )
  }
  {   
    ?resource1 schema:identifier ?resource2 .
  } UNION {
  {
    VALUES ( ?resource1 ) {
     ( <https://ld.admin.ch/country/CHE> )
    }
    ?resource1 schema:name ?resource3 .
    FILTER (lang(?resource3) = "de" ) }
  } UNION {    
{ 
VALUES ( ?resource1 )
{
( <https://ld.admin.ch/country/CHE> )
}

      ?resource1 schema:description ?resource5 .
      FILTER (lang(?resource5) = "de" ) }

    } UNION {
      
?resource1 schema:position ?resource7 .

    } UNION {
      
{ 
VALUES ( ?resource1 )
{
( <https://ld.admin.ch/country/CHE> )
}

      ?resource1 schema:alternateName ?resource8 .
[...]

PR #1244, opened 2023-11-02
feat: Add a way to retrieve dimensions from multiple cubes

bprusinowski:
This PR adapts the application logic to support retrieving dimensions from multiple cubes at once.

### GQL / SPARQL
- Consolidated several GQL queries together into `DataCubesComponents` that supports retrieving the components from several cubes at once
- Removed `Components` query (replaced by `DataCubesComponents`)
- Removed `ComponentsWithHierarchies` and `DimensionHierarchy` queries (hierarchies are now fetched by default in `DataCubesComponents`). As we generally use them in many different places across the app, it's more developer-friendly to have one source of truth (and also, it might be even more performant to have one query, instead of several different ones fired from different parts of the app). We can also see how [the new version](https://github.com/visualize-admin/visualization-tool/pull/1223) of `query-cube-hierarchy` library improves performance, as then it might be even less worrying, or we can bring back the old way of separating hierarchies fetching to the specific places that use them
- Removed `DimensionValues` query (no need to have a separate query to fetch dimension values, as we can just specify one dimensionIri in  `DataCubesComponents` to get the same result)
- Removed dimensions & measures from `SearchCubes` query
- We will now get all hierarchy levels, not limited to 6 as before (by using scalar instead of nested retrieval)
- To further improve performance and limit resolver chains, I've extracted the dimension types from GQL to TypeScript. It also means that we now explicitly parse them inside the resolver
- Removed `DimensionMetadataFragment` type, which means we can now better type `Dimension` and `Measure`, instead of mostly using the type that contains both

### Misc
- Fixed default empty filter selection when initializing a segment and hierarchy is not present

---
## Context
This PR is a first part of preparations for merging the cubes. For the technical concept, see [this Notion document](https://www.notion.so/interactivethings/Combine-two-cubes-in-the-same-chart-d5711b414a7747e79edfc3cfe17d2c9b).

- [x] `DataCubesComponents`
- [ ] `DataCubesMetadata`
- [ ] `DataCubesObservations`

## How to test
1. Click around in the application and make sure nothing is broken 💥 
2. Open [GQL playground](https://visualization-tool-git-feat-several-cubes-in-the-sa-cef96f-ixt1.vercel.app/api/graphql) and send e.g. the following query to see that all dimensions from both cubes were fetched.

**Query**
```graphql
query DataCubesComponents(
  $sourceType: String!
  $sourceUrl: String!
  $locale: String!
  $filters: [DataCubeFilter!]!
) {
  dataCubesComponents(
    sourceType: $sourceType
    sourceUrl: $sourceUrl
    locale: $locale
    filters: $filters
  )
}
```

**Variables**
```js
{
  "sourceType": "sparql",
  "locale": "en",
  "filters": [{ "iri": "https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/9", "latest": true }, { "iri": "https://environment.ld.admin.ch/foen/ubd0104/6", "latest": true }]
}
```


Rdataflow:
@bprusinowski can you add an additional point for `how to test` CI screenshots 😄

bprusinowski:
Hi @Rdataflow, in fact we would need to switch back to Argos for screenshot testing, as we used the limit already on Percy (so not possible to check them on CI now) 😞

bprusinowski:
@ptbrowne it looks like the payload is too large (https://vercel.com/ixt/visualization-tool/3pbpsBYim6U2gRc2bE1eDnDovxMx/logs?page=2&timeline=past30Minutes&startDate=1700148221941&endDate=1700150021941&selectedLogId=1700149691375354119628444121&selectedLogTimestamp=1700149691375&forceEndDate=1700149937685), the underlying message says it's error 413 👀

app/graphql/resolvers/rdf.ts
```diff
@@ -157,7 +159,25 @@ export const dataCubesComponents: NonNullable<
 
   return (
     await getCubesDimensions(cubes, { locale, sparqlClient, filters, cache })
-  ).map((d) => ({ iri: d.data.iri }));
+  ).map(async (dimension) => {
+    const { cube, data } = dimension;
+
+    return {
+      __typename: resolveComponentType(dimension),
```
ptbrowne: Previously this was automatically done by the resolver, but now we have to put it manually right, because we tried to shorten the resolver chain, is this right ?

app/charts/shared/chart-state.ts
```diff
@@ -177,14 +176,14 @@ export type TemporalXVariables = {
...
-  { dimensionsByIri }: { dimensionsByIri: ComponentsByIri }
+  { dimensionsByIri }: { dimensionsByIri: DimensionsByIri }
 ): TemporalXVariables => {
   const xDimension = dimensionsByIri[x.componentIri];
   if (!xDimension) {
     throw Error(`No dimension <${x.componentIri}> in cube!`);
   }
 
-  if (!isTemporalDimension(xDimension)) {
+  if (!isDataCubeTemporalDimension(xDimension)) {
```
ptbrowne: Why did we have to put "DataCube" everywhere in those methods ?

ptbrowne:
I will retake a look tomorrow.

### What I like

- Simplifying resolver chain
- Less different type of queries
- Better typing client side

### What I have questions on

- Why we put everything into 1 query client side ? Seems like it would not be as cache friendly & we would have unwanted invalidations when adding a new cube to compare.
- Seems like the client hooks could hide that they are doing multiple graphql requests & merge everything client side. We use this kind of technique on UIS.

-> useDataCubesComponentsQuery could be a custom hook relying on
    multiple useDataCubesComponentQuery (without "s") calls

app/graphql/resolvers/rdf.ts
```diff
@@ -136,6 +145,150 @@ export const possibleFilters: NonNullable<QueryResolvers["possibleFilters"]> =
...
+      );
+      const values: DimensionValue[] = await dimensionValuesLoader.load(
+        component
+      );
+      values.sort((a, b) =>
+        ascending(
+          a.position ?? a.value ?? undefined,
+          b.position ?? b.value ?? undefined
+        )
+      );
```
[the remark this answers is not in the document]
bprusinowski: Yup, this could be moved there 👍

app/graphql/schema.graphql
```diff
@@ -354,9 +354,25 @@ enum SearchCubeResultOrder {
...
+}
+
 # The "Query" type is special: it lists all of the available queries that
 # clients can execute, along with the return type for each.
 type Query {
+  dataCubesComponents(
+    sourceType: String!
+    sourceUrl: String!
+    locale: String!
+    filters: [DataCubeFilter!]!
```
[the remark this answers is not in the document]
bprusinowski: Yes, this name should be more clear :)

app/graphql/resolvers/rdf.ts
bprusinowski: Yes 👍

app/charts/shared/chart-state.ts
bprusinowski: That was a temporary solution to gradually replace "new" (DataCube) types with "old" types (without prefix). Probably wasn't super necessary, but I found it easier to navigate the changes then 👀

bprusinowski:
Thanks for reviewing @ptbrowne. I agree it would be better to separate the components queries per cube, so cube-specific queries are not re-fetched when adding another one 👍 I would be interested to learn how to do it, I guess we would need to use `client.query` directly instead of current `useXYZQuery` hook so we have several queries combined together in `Promise.all`?

PR #1245, opened 2023-11-02
fix: Dataset preview not showing

bprusinowski:
This PR fixes the Framer dataset preview animations not showing sometimes by adding keys where needed.


lloydrichards (approved):
LGTM 👍

PR #1248, opened 2023-11-03
fix: Flickering username

bprusinowski:
This PR removes providers presence checking each time we want to get current user, as there doesn't seem a reason to do so at this point, and this causes the username to switch back between currently logged in username and "Sign in" string when navigating through the pages.

### DIFF
--- a/app/login/utils.ts
+++ b/app/login/utils.ts
@@ -1,7 +1,6 @@
 import { Theme } from "@mui/material";
 import { makeStyles } from "@mui/styles";
-import { getProviders, useSession } from "next-auth/react";
-import React from "react";
+import { useSession } from "next-auth/react";
 
 import { HEADER_HEIGHT } from "@/components/header";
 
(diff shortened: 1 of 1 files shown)


lloydrichards (approved):
Makes sense.   Looks like each time `useUser()` is called, the `useProviders()` is initialised with the default "loading" state causing it to return null. Maybe its better to keep the `useProviders()`  and just remove its nesting from useUser incase you need it later within a component 🤷‍♂️

bprusinowski:
It might be worth keeping, but I don't really see the point. It looks like it returns a list of [providers configured for login](https://next-auth.js.org/getting-started/client#getproviders), but in our case it'll always be just one (Keycloak). Also, it might be easier to just check for the presence of the [env variables used to create the provider](https://github.com/visualize-admin/visualization-tool/blob/e65ac3c1cf9ca8c19df6cfbb59b59376a7e5e753/app/pages/api/auth/%5B...nextauth%5D.ts#L10-L18), instead of calling an async hook 👀

PR #1249, opened 2023-11-03
chore: Remove deprecated validThrough cube filters

### LINKED ISSUE #1195: phase out deprecated `schema:validThrough` FILTER
**phase out deprecated `schema:validThrough` FILTER**
A clear and concise description of what the problem with the current state is. Ex. I'm always frustrated when [...]

drop `schema:validThrough` FILTER

keep deprecated code in repo 🙃 

**Usecases and impact**
tidy repo

**Additional context**
on all 3 endpoints there exist 3 (three) cubes featuring `schema:validThrough` https://s.zazuko.com/yHE7Eq
all of them also feature successor `schema:expires` so it's safe to drop deprecated `schema:validThrough` all over the places 👍

bprusinowski:
Closes #1195.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ You can also check the [release page](https://github.com/visualize-admin/visuali
 - Maintenance
   - Added retrieval of dimension units via `qudt:hasUnit` (but kept `qudt:unit` for backward compatibility)
   - Improved GQL debug panel (added resolver variables and rectangles to visually indicate resolving times)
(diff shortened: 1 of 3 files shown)


lloydrichards (approved):
LGTM 👍 I'm guessing there is some tests in E2E that are still testing this filter though?

bprusinowski:
I think not, at least we don't explicitly have `validThrough` anywhere in the app now, except in RDF typings. As this filter was essentially the same as `noExpires`, it shouldn't affect the tests 👍

Issue #589, opened 2022-07-07
Standard Error visualization for missing errors


kronmar:
Currently in the bar plots, if the standard error for some observation/measurement is given as as "null" (i.e. no standard error is given), the error bar is displayed as a thick line width height zero.

This should not happen as it implies that an error was provided (hovering over the observation clears this by providing the real error, meaning "x [unit] +/- null [unit]"). Instead, no visualization of this error should be plotted.

![image](https://user-images.githubusercontent.com/95312586/177784930-8b5c61cc-16a6-41ef-a2b1-ced6f2d424b5.png)

lloydrichards:
An example of a dataset would be [NFI: Topics by forest type and stage of development](https://opendata.swiss/en/perma/nfi_C-501%40bundesamt-fur-umwelt-bafu)

kronmar:
Oh this is a bit of an older issue, I no longer had on my radar... 

But maybe to help with it, here is a concrete case where such a thing would happen: [https://visualize.admin.ch/en/v/lKB-Vc3eRFwg](https://visualize.admin.ch/en/v/lKB-Vc3eRFwg)

PR #1252, opened 2023-11-06
fix(column): :recycle: remove ErrorWhisker and Tooltip when error is null

lloydrichards:
It was requested to remove the Standard Errors from the column visualization (#589) so that the ErrorWhiskers component does not show for these columns.

While I was able to filter the chartDatum for any values where the getYError was null, I was also able to remove the getError from the getAnnotationinfo when null
### Before
Screenshot 2023-11-06 at 14 36 00
### After
Screenshot 2023-11-06 at 14 36 10

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ You can also check the [release page](https://github.com/visualize-admin/visuali
 
 ## Unreleased
 
-Nothing yet.
+- Fixes
+  - remove ErrorWhisker and Tooltip when error is null in Column charts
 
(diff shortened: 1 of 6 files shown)


bprusinowski:
Nice! There is only one caveat: we can also have standard error whiskers for grouped column charts, I think the logic needs to be applied there too 😁

Nit: you can also add a relevant bullet point to the CHANGELOG, so at the end we have a nice list of all the changes made to the app ✨

bprusinowski (approved):
LGTM 💯

Issue #437, opened 2022-03-01
Display a warning / error when there are negative values in stacked area chart


bprusinowski:
A stacked area chart isn't able to properly display negative values; that's why if it encounters some, results may look quite weird and confusing. We should display some kind of a warning / error that at least would let user know that it could be better to switch to a stacked bar chart or present some call to action to disable affected series from a chart.

bprusinowski:
cc @sosiology

lloydrichards:
Example dataset: [Greenhouse gas emissions by sector and by gas](http://localhost:3000/en/browse?previous=%7B%7D&dataset=https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fubd000502%2F4&dataSource=Prod)

PR #1251, opened 2023-11-06
chore(DX): :technologist: add debug launcher and fix extension recommendations

lloydrichards:
Some minot tweaks to the DX by:

1. Replacing the extension recommendations with a non-depreciated MDX extension
1. Added launch.json with setup for using vscode debugger
1. Added documentation in README.md

PR #1253, opened 2023-11-06
refactor(components): :recycle: created a Banner component for displaying warnings or errors

lloydrichards:
> A stacked area chart isn't able to properly display negative values; that's why if it encounters some, results may look quite weird and confusing. We should display some kind of a warning / error that at least would let user know that it could be better to switch to a stacked bar chart or present some call to action to disable affected series from a chart. (#437)

To keep things generic and possibly reusable in the future, I've created a Banner component which will use the useChartState() to conditionally render a warning message at the top of the chart.

Screenshot 2023-11-06 at 15 54 18

I'm open to suggestions on how to improve this and whether there might be a better place to throw or create this warning handeling.

Additionally, I needed to update the translations with a message for this warning so if there is a better way of doing this, le tme know 🙏


app/charts/area/chart-area.tsx
```diff
@@ -78,26 +79,31 @@ export const ChartAreas = memo((props: ChartProps<AreaConfig>) => {
...
+          <ChartControlsContainer>
+            <LegendColor
+              chartConfig={chartConfig}
+              symbol="square"
+              interactive={interactiveFiltersConfig?.legend.active}
+            />
+          </ChartControlsContainer>
+        )}
+      </AreaChart>
+    </>
```
ptbrowne, line 107: Is the wrapping  in Fragment wanted ?

app/components/banner.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { Box } from "@mui/material";
+
+import { HintYellow } from "@/components/hint";
+
+import { useChartState } from "../charts/shared/chart-state";
+
+/**
+ * Banner to display errors or warnings for a chart. Each chart type can have specific warnings or errors depending on the data.
+ *
+ * @returns React.ReactNode | null
```
ptbrowne, line 11: No need for the returns, it is inferred by typescript ?

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
...
+    "configurations": [
+        {
+            "type": "chrome",
+            "request": "launch",
+            "name": "Launch Chrome against localhost",
+            "url": "http://localhost:3000",
+            "webRoot": "${workspaceFolder}/app"
+        }
+    ]
+}
```
ptbrowne, line 15: Not sure this is intended to have in this PR. I think those kind of files should remain personal no ? Everyone has their favorite browser etc..

ptbrowne (approved):
LGTM, but I think a good secondary step should be to extract the business logic of extracting warnings + severity out of the chart staet, from the Banner itself. This way we could more easily test Banner against variants.

Something like this:

Here I find that from the parent, it is difficult to infer that the Banner goes over all the data to find warnings.

```jsx
const WarningNegativeValues = () => <>
   <Trans id="dataset.error.negative-values">
      Careful, this dataset contains negative values and might not
      display correctly with this chart type.
      <br />
      <strong>Try using another chart type.</strong>
    </Trans>
</>

const getWarningFromChartState = chartState => {
	return hasNegativeValues(chartState) && area ? {
       severity : 'warning',
       Message: <WarningNegativeValues /> } : null 
}

const warning = useMemo(() => {
  return getWarningFromChartState(chartState)
})

return  <>
  { warning ? <Banner  warning={warning} /> : null }
  <Area />
   <Etc />
</>
```

I find it easier to see that Banner is responsible for warnings. What do you think ?

lloydrichards:
> What do you think ?

This was how I wanted to build it initially, but looking at the [chart-area.tsx](https://github.com/visualize-admin/visualization-tool/blob/main/app/charts/area/chart-area.tsx) it seems like all the chart data is drilled down using providers, so wanted to keep the same pattern going.  if the same data can be extracted from props, and the logic kept at the highest level, then maybe a hook that returns some kind of generic Failure object which can be passed to the Banner as props 🤔

.vscode/launch.json
lloydrichards: This is kind of a ghost from the PR #1251 where I'm adding this anyways.  Could remove it i guess, from this PR just to keep it tidy and leave the discussion on whether its important to include it in the other one 🤷‍♂️

app/charts/area/chart-area.tsx
lloydrichards: ah, no no this was from when i was trying to place the Banner somewhere that felt good

app/components/banner.tsx
```diff
@@ -0,0 +1,37 @@
...
+ */
+export const Banner = () => {
+  const state = useChartState();
+
+  switch (state.chartType) {
+    case "area":
+      const someValueIsNegative = state.allData.some(
+        (d) => +(state.getY(d) || 0) < 0
+      );
```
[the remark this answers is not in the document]
lloydrichards: wasn't able to memoize the value, but abstracted the warning into its own useMemo conditional on the `state.chartType` which should save some performance.

ptbrowne (approved):

app/components/banner.tsx
```diff
@@ -1,37 +1,40 @@
...
+          (d) => +(state.getY(d) || 0) < 0
         );
-    default:
-      return null;
-  }
+        if (someValueIsNegative) return <NegativeValueWarning />;
+      default:
+        return null;
+    }
+  }, [state.chartType]); // eslint-disable-line react-hooks/exhaustive-deps
```
ptbrowne, line 26: Wondering why the eslint warning was removed ? I think it's fine to also pass state.allData and state.getY. You do not know whether getY would change the number of negative values for example.

ptbrowne:
Yep ok, understood the need for coherence across charts, thanks for the explanation 👍  I have just seen the disabled eslint warning, and I think it could be a pitfall though, in general I think we should reserve disabling the exhaustive hook warning only for very specific cases, and I am not sure it is that useful here.

app/components/banner.tsx
lloydrichards: im just not sure what happens with the `useChartState()` that might cause it to be rebuilt.  I figure the only time we want the banner to rerender is when its first loaded when the chart type changes, but maybe I can try with exhaustive-deps

ptbrowne:
(Just to be clear, I had put LGTM, so feel free to merge anytime)

lloydrichards:
> (Just to be clear, I had put LGTM, so feel free to merge anytime)

Thanks! I want to just wait til Bartosz is back and can have a look since it also relates to error/failure handling and whether this will be displayed on the published charts too 🙏

app/components/banner.tsx
bprusinowski: I think here we should use `state.chartData`, as the error should dynamically adjust to e.g. interactive filters (to not show it when some negative values are present in the whole dataset, but not in the current filter selection).

app/components/banner.tsx
```diff
@@ -0,0 +1,37 @@
...
+      );
+      if (someValueIsNegative)
+        return (
+          <Box sx={{ mb: 4 }}>
+            <HintYellow iconName="datasetError" iconSize={64}>
+              <Trans id="dataset.error.negative-values">
+                Careful, this dataset contains negative values and might not
+                display correctly with this chart type.
+                <br />
+                <strong>Try using another chart type.</strong>
```
bprusinowski: I wonder here if the error shouldn't be a bit more end-user facing, like

`Careful, this chart might not display correctly due to negative values.`

I am not sure if we should have "Try using another chart type" displayed in published chart, what do you think?

app/components/banner.tsx
bprusinowski: If we switch to `state.chartData`, we also need to add it here.

bprusinowski (commented):
Cool 💯 

There are some things mentioned in the comments, mostly about using `chartData` instead of `allData`. This could result in banner disappearing and reappearing when using interactive filters, but it would reflect the reality, and only be displayed when negative values are actually used in chart 👀 

Also, one nit comment: we recently started to consolidate the chart-related logic [here](https://github.com/visualize-admin/visualization-tool/blob/97a70de080ebe42c4c1ff22509dc0c64ab4cc19f/app/charts/chart-config-ui-options.ts#L526), so you can have most of the logic in one place instead of spread out across components.

While currently the `chartConfigOptionsUISpec` is mostly about editor UI elements, I think it would make sense to extend the `ChartSpec` type with e.g. `getBannerMessage`, and move the negative warning logic there, so in the end it's not stored in `Banner`, but rather accessed through `chartSpec?.getBannerMessage`. What do you think?

bprusinowski:
@lloydrichards for the chart data vs all data see https://github.com/visualize-admin/visualization-tool/blob/97a70de080ebe42c4c1ff22509dc0c64ab4cc19f/app/charts/shared/chart-state.ts#L477-L490

app/components/banner.tsx
adintegra: I agree with this. Also, not sure it makes sense in the published context as the user would not have control to change anything. My preference would probably not to have this messaging in the published chart. What do you think?

bprusinowski: I think it would make sense if it's not displayed in published chart – we have a persistent banner already if the data is imputed in area chart, but here as we are not altering the data artificially, it should be fine to not have it displayed 👍

lloydrichards: I would push back slightly on not showing the banner in published views.  Having negative and positive numbers result in the area being distorted and visualization delivering misleading results.  Public users might misinterpret the results without every knowing that the information is misleading.  This might not be the case for all banners, but for the case of negative numbers this could be exploited.

lloydrichards:
> While currently the `chartConfigOptionsUISpec` is mostly about editor UI elements, I think it would make sense to extend the `ChartSpec` type with e.g. `getBannerMessage`, and move the negative warning logic there, so in the end it's not stored in `Banner`, but rather accessed through `chartSpec?.getBannerMessage`. What do you think?

I like the idea of keeping all the logic in one place, but will first need to figure out if UISpec is specific to the public chart or the configurator view.  If the Banner is to share responsibility between these views then having this outside of the chartConfigOptionsUISpec makes more sense, unless there is precedence to conditionally render UI elements already?

lloydrichards:
> Maybe we could have a new property there to store banner-related logic, or even restructure a bit to have a clear distinction, e.g.

This is the same `ChartSpec` that is stored as the `ChartConfig` in the `ConfiguratorConfig`?  I personally wouldnt want to misconstrue configuration and error handling, especially if this logic is being build over time.  I think a more functional approach of accepting the config and returning a message depending on the view would make more sense 🤷‍♂️ like:

```ts
type Failure = ChartError | ConfigWarning | VisualizationWarning | etc

export const useConfiguratorFailureMessages = (config: ChartSpec) => Failure[]

export const useChartFailureMessages = (config: ChartSpec) => Failure[]
```

bprusinowski:
I think a difference in the end would be that we have another function to access it instead of passing it as part of `ChartSpec` – as long as they're close together that's fine too 👌

The problem at one point was that we had such a "separation of concerns" approach to having different chunks of logic throughout the app, and it became extremely difficult to foresee all the consequences of changing something in a particular place in the app (e.g. a side effect of changing a particular component for a particular chart type); after the expansion of the high-level "chart spec" it became easier (you just look at one object and see the consolidated view of what will happen if there are missing values or a given option changes), that's why I suggested to follow this approach :)

PR #1260, opened 2023-11-09
fix: Duplicated hierarchies

### LINKED ISSUE #1257: embedded charts broken
**Describe the bug**
embedded charts are broken
https://visualize.admin.ch/de/v/IYdcAI0tH0Io?dataSource=Prod
and show duplicate legend and duplicate hierarchy in filters 👀 

**To Reproduce**

1. Go to https://visualize.admin.ch/de/v/IYdcAI0tH0Io?dataSource=Prod
2. see duplicate legend
3. open filter
4. see broken hierarchy (introduced intermediate level & duplicate elements)

**Expected behavior**
chart as displayed here https://visualize.admin.ch/de/v/yHbeA1GiyqXB?dataSource=Prod
- single legend
- no intermediate level in hierarchy
  - no duplicates in hierarchy

**Screenshots or video**
- duplicate legend
   image
- duplicate hierarchy
   image

- Visualize environment and version: PROD v3.22.9 ([a09cdc3](https://github.com/visualize-admin/visualization-tool/commit/a09cdc30abf0744481670f6549f5766610073b1c))

@bprusinowski a rapid fix would be helpful

### LINKED ISSUE #1263: Do not fetch shapes when getting raw cube
Currently, we initialize the cube along with its shape, even before we know if a newer cube is available (see https://github.com/visualize-admin/visualization-tool/blob/6bfaf3d2cdfe42ea2c287c482bde59b3c4efd1be/app/graphql/context.tsx#L27-L29).

We should only fetch the shape when we are sure that a given cube is the "final" one, to avoid sending unnecessary DESCRIBE query to LINDAS.

bprusinowski:
Fixes #1257.
Closes #1263.

This PR:
- fixes hierarchies querying by only fetching the hierarchies based on cube's shape (and retrieving correct hierarchy  names),
- removes fetching of legacy hierarchies (`hasHierarchy`), as there is only [one such hierarchy on LINDAS](https://s.zazuko.com/2PTZ2X2),
- improves data fetching performance by not fetching shape for raw cubes (as a newer one might be needed anyway).

## Context

It looks like there's too many hierarchies fetched in the example chart from #1257 (https://visualize.admin.ch/de/v/IYdcAI0tH0Io?dataSource=Prod).

There are two issues visible in this chart:
- (apparent) duplicated hierarchies,
- hierarchy that (most probably) shouldn't be fetched at all.

https://github.com/visualize-admin/visualization-tool/assets/52032047/cc8b26fc-194d-4fb3-9007-9d277f7ce6c1

Screenshot 2023-11-09 at 22 36 08

### (Apparent) duplicated hierarchies
First issue occurs due to retrieving name from `hierarchyPointer.out(ns.cubeMeta.nextInHierarchy)` instead of `hierarchyPointer` directly. @ptbrowne do you remember the context here (it looks like this has been done to [avoid duplication](https://github.com/visualize-admin/visualization-tool/commit/cf8e06c00bcfc2c19dab141f281834d9abb16dc0))? After looking at [docs](https://cube.link/#hierarchies) and investigating the data on LINDAS, I think we should base directly on `hierarchyPointer` – if we do it, the name of hierarchies are unique (`Schweiz - Produktionsregion - Wirtschaftsregion` and `Schweiz - Produktionsregion`).

### Hierarchy that (most probably) shouldn't be fetched at all
Second one is more weird – generally, when we fetch a cube, two queries are fired, [one for cube and the other one for cube shape](https://github.com/zazuko/rdf-cube-view-query/blob/eb7d4b1c05c717d0a9a5be2547d4e691bc0e1d3e/lib/Cube.js#L67-L70). It looks like this behavior results in more hierarchies that we're actually interested in, as (I think, also based on [kronmar's query](https://s.zazuko.com/3aAR4HP)), we should only fetch hierarchies that are specified in the cube's shape.

I didn't investigate enough to say for sure why these additional hierarchies are being fetched along with `cube`, but if we just fetch `cube shape`, it retrieves correct hierarchies.

~~It's a bit unfortunate, as basically we need to fetch shape twice, once initially, and afterwards again when fetching hierarchies, as it looks we can't distinguish if a given triple came from `cube` or `shape`. Some optimisation might be possible, but I think it would need to be somehow connected with changes to `cube-view-query` library, to be able to tell if a given triple came from one or the other query.~~

I introduced an `ExtendedCube` class to allow us to store shape triples inside `shapePtr`, thus removing a need of fetching the shape twice. It might be useful to introduce such improvement directly to `cube-view-query` library, or to better understand why there is hierarchy mismatch when fetching cube and shape.

cc @Rdataflow


app/rdf/query-hierarchies.ts
```diff
@@ -127,25 +133,25 @@ export const queryHierarchy = async (
...
-      nodes: (await getHierarchy(h).execute(sparqlClientStream, rdf)) || [],
-      hierarchyName: getName(h.out(ns.cubeMeta.nextInHierarchy), locale),
-    }))
+  const hierarchies = await Promise.all(
+    hierarchyPointers.map(async (pointer) => {
+      return {
+        // @ts-ignore
+        nodes: await getHierarchy(pointer).execute(sparqlClientStream, rdf),
+        hierarchyName: getName(pointer, locale),
+      };
```
lloydrichards, line 142: Why the @ts-ignore ? Also i think you can save the `return` and just keep it inline

lloydrichards (approved):
Can't see anything out of the ordinary, but also don't have much experience with the cube queries.

app/rdf/query-hierarchies.ts
```diff
@@ -73,24 +73,39 @@ const toTree = (
...
+  const hierarchyCube = new Cube({
+    source: cube.source,
+    term: rdf.namedNode(cubeIri),
+  });
+  // Need to set the shape query manually, because we didn't initialize the cube,
+  // thus can't access its observationConstraint.
+  hierarchyCube.shapeQuery = () => {
+    return `DESCRIBE <${cube.out(ns.cube.observationConstraint)}>`;
+  };
+  await hierarchyCube.fetchShape();
```
Rdataflow, line 100: two questions:
- does this trigger yet another query?
  - i.e. might this have an impact on performance?
  - can separate queries be avoided? i.e. by filtering cubeIRI to only the single relevant cube?
  - nb: is this query traced in debug mode? (once the let let's you reuse your client)

app/rdf/query-hierarchies.ts
bprusinowski: Here the types are quite messy, see e.g. the image below – I tried to fix them, but failed 😅 But it should still work though, as we had this comment previously 👍 

Screenshot 2023-11-10 at 11 55 24

app/rdf/query-hierarchies.ts
bprusinowski: I refactored the PR a bit and modified the `Cube` class from cube-view-query library, so we can differentiate if given triple comes from describing the `cube` or its `shape`. Now we won't fire any additional query :)

Rdataflow: with your refactored changes there is still the shape for the expired cube being fetched somehow... 
```sparql
DESCRIBE <https://environment.ld.admin.ch/foen/nfi/nfi_C-90/cube/2023-2/shape/>
```
do you init that cube somewhere before updating to non-expired?

bprusinowski: I was testing with [this chart](https://visualize.admin.ch/de/v/IYdcAI0tH0Io?dataSource=Prod), which is based on v1 of the cube, while the newest one is v3 – and with the change we fetch the correct shape.

Screenshot 2023-11-10 at 13 37 53

But it's true that we might fetch some data related to cube that won't be used (as a newer cube is available) – we need it to determine if we need to fetch newer cube, so I'm not sure we can avoid this. We might not use full `DESCRIBE` query, but rather smaller one to just fetch things we need, but that would be a performance improvement not related to hierarchy bug 👍 

Could you share more information on how you encountered below?

```
DESCRIBE <https://environment.ld.admin.ch/foen/nfi/nfi_C-90/cube/2023-2/shape/>
```

Rdataflow: @bprusinowski you can get such information as well using
- mitmproxy
- introduce full traceability (once the lib is ready)
sure let's tackle this in the perf context. :+1: 
NB: did you already start with the perf improvement for preview? https://gitlab.ldbar.ch/bafu/visualize/-/issues/547

bprusinowski: Ah, I missed the point, I thought that we fetch wrong shape when querying dimensions, and it was about generally not fetching outdated shapes 🤦‍♂️ 

I tackled this here: https://github.com/visualize-admin/visualization-tool/pull/1260/commits/d08bd37e770b62f5c3e7e8bcf5a04926f23884fc

bprusinowski: I've seen that Tomasz updated the packages, but didn't yet get around to implementing them 😄

Rdataflow:
@bprusinowski LGTM :+1:

app/rdf/query-hierarchies.ts
Rdataflow: @bprusinowski as a follow up to this PR it would be helpful to 
1. introduce tracability based on newly released version
2. refactor preview completely based on proposed optimization from GL issue (1 single CONSTRUCT query) and fine tune that query :smile:

app/graphql/resolvers/rdf.ts
```diff
@@ -73,21 +75,27 @@ export const searchCubes: NonNullable<QueryResolvers["searchCubes"]> = async (
...
+    return parseCube({ cube, locale: locale ?? defaultLocale });
   };
 
 export const possibleFilters: NonNullable<QueryResolvers["possibleFilters"]> =
   async (_, { iri, filters }, { setup }, info) => {
     const { sparqlClient, loaders, cache } = await setup(info);
-    const cube = await loaders.cube.load(iri);
+    const rawCube = await loaders.cube.load(iri);
+    // Currently we always default to the latest cube.
+    const cube = await getLatestCube(rawCube);
```
ptbrowne, line 97: For the splitted issue, getLatestCube could be put on ExtendedCube.

app/rdf/extended-cube.ts
```diff
@@ -0,0 +1,29 @@
+import clownface, { AnyPointer } from "clownface";
+import { Cube, CubeOptions } from "rdf-cube-view-query";
+import rdf from "rdf-ext";
+import DatasetExt from "rdf-ext/lib/Dataset";
+
+export class ExtendedCube extends Cube {
```
ptbrowne, line 6: Cna you add a docstring describing the general goal ? So that when we see a method that should be there, we can be sure that this class is the right one ?

app/rdf/query-hierarchies.ts
```diff
@@ -73,33 +73,29 @@ const toTree = (
...
-    .out(ns.cubeMeta.hasHierarchy)
+    .has(ns.cubeMeta.inHierarchy)
+    .out(ns.cubeMeta.inHierarchy)
     .toArray();
-  if (legacyHierarchies) {
-    return legacyHierarchies;
-  }
-  return [];
+
+  return uniqBy(hierarchies, (hierarchy) => getName(hierarchy, locale));
```
ptbrowne, line 98: For the splitted issue : getDimensionHierarchies could be on ExtendedCube. This way we will hide the complexity of the shapePtr there.

ptbrowne (approved):
👍 LGTM, I think it's good to have a splitted issue with the future of ExtendedCube. Thanks !

PR #1264, opened 2023-11-10
chore: Make all queries traceable (DebugPanel)

bprusinowski:
This PR updates `cube-view-query` library to allow to pass SPARQL client directly into `Source`, instead of being re-created each time.

This allows us to keep track of queries made through this library inside the `DebugPanel`, which means we will now track every query :)

### DIFF
--- a/app/charts/index.spec.ts
+++ b/app/charts/index.spec.ts
@@ -1,6 +1,7 @@
 import { ColumnConfig, TableFields } from "@/configurator";
 import { ComponentsQuery } from "@/graphql/query-hooks";
 import { DataCubeMetadata } from "@/graphql/types";
+import { RDFCubeViewQueryMock } from "@/test/cube-view-query-mock";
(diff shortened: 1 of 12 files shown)


app/charts/index.spec.ts
```diff
@@ -11,6 +12,8 @@ import {
   getPossibleChartTypes,
 } from "./index";
 
+RDFCubeViewQueryMock;
```
bprusinowski, line 15: Is there a nicer way to doing this?

Rdataflow:
@bprusinowski congrats, this looks very promising :smile:

app/charts/index.spec.ts
ptbrowne: Maybe instead ignore the eslint rule that remove it automatically in case of no use, at least the ignore would be at the place of the import and we would directly understand it.

Issue #1266, opened 2023-11-13
Mouse Over Box is cutt off


ortnever:
The Mouse Over Box should appear on the right Side so as not to be cut off. The Problem appears only by embedding in AEM or iFrame. The URL to share is ok : https://visualize.admin.ch/en/v/uCDwLYAzMW5c

![image](https://github.com/visualize-admin/visualization-tool/assets/79834090/eac9b813-8036-4adb-a190-aaf68fc6bb16)

1. Go to 'https://visualize.admin.ch/en/v/uCDwLYAzMW5c"
2. Click on 'Embed Code"
3. Copy URL "Embed Code for AEM "External Application" - Minimal View : https://visualize.admin.ch/api/embed-aem-ext/en/F3IdN_gPmLuP?embedOptions=%7B%22showDatasetTitle%22%3Afalse%2C%22showDownload%22%3Afalse%2C%22showLandingPage%22%3Afalse%2C%22showSource%22%3Atrue%2C%22showMetadata%22%3Afalse%2C%22showSparqlQuery%22%3Afalse%2C%22showDatePublished%22%3Afalse%2C%22showTableSwitch%22%3Afalse%7D
4. Insert in AEM
5. See error

**Expected behavior**
We should always see the whole Box by Mouse oVER.

**Screenshots or video**
If applicable, add screenshots or a short video to help put your problem into a context.

**Environment (please complete the following information):**

- Visualize environment and version: prod
- Browser and version: eDGE

**Additional context**

PR #1268, opened 2023-11-14
fix(interaction): :bug: create conditional Xplacement based on container width

lloydrichards:
>The Mouse Over Box should appear on the right Side so as not to be cut off. The Problem appears only by embedding in AEM or iFrame. The URL to share is ok : https://visualize.admin.ch/en/v/uCDwLYAzMW5c
>
>image
>
>#1266

## Goals/Scope
After investigating it seems the bug is cased on small containers where the tooltip size is larger then the allotted space.  Since larger screens still work, I looked to create a solution that could conditionally place the tooltip based on the charts size rather then measuring the window.

## Description
Created a fold function that accepts the `chartWidth` and returns a function that can be used to define the behaviour of the xPlacement depending on that size.  Here we can define the `xs` (smallest) and `md` (@breakpoint 600) to define the mobile-first behaviours.  

## Comments
I'm not sure how to test the embedded links without deploying this to INT, any suggestions are welcome in the comments 👇

### DIFF
--- a/app/charts/shared/interaction/tooltip-box.tsx
+++ b/app/charts/shared/interaction/tooltip-box.tsx
@@ -69,6 +69,30 @@ const usePosition = () => {
   return [box, handleRef] as const;
 };
 
+/**
+ * Responsible for folding the container size into a single value based on breakpoints.
+ * After accepting the container width, it accepts functions each breakpoint to return the generic value.
+ */
+const foldContainerSize =
+  <T,>(containerWidth: number) =>
+  ({
+    xs,
+    md,
+    lg,
+  }: {
+    xs: (width: number) => T;
+    md?: (width: number) => T;
+    lg?: (width: number) => T;
+  }) => {
+    if (lg && containerWidth >= 900) {
+      return lg(containerWidth);
+    }
+    if (md && containerWidth >= 600) {
+      return md(containerWidth);
+    }
+    return xs(containerWidth);
+  };
+
 export const getCenteredTooltipPlacement = (props: {
   chartWidth: number;
   xAnchor: number;
@@ -82,7 +106,10 @@ export const getCenteredTooltipPlacement = (props: {
         y: "top",
       }
     : {
-        x: xAnchor < chartWidth * 0.25 ? "right" : "left",
+        x: foldContainerSize<Xplacement>(chartWidth)({
+          xs: (w) => (xAnchor < w * 0.5 ? "right" : "left"),
+          md: (w) => (xAnchor < w * 0.25 ? "right" : "left"),
+        }),
         y: "middle",
       };
 };


app/charts/shared/interaction/tooltip-box.tsx
```diff
@@ -69,6 +69,30 @@ const usePosition = () => {
   return [box, handleRef] as const;
 };
 
+/**
+ * Responsible for folding the container size into a single value based on breakpoints.
+ * After accepting the container width, it accepts functions each breakpoint to return the generic value.
+ */
+const foldContainerSize =
+  <T,>(containerWidth: number) =>
```
bprusinowski, line 77: Super-nit: I think we can replace `<T,>` with` <T>`? Is there a reason for this or is it a typo? 😅

bprusinowski (approved):
LGTM!

I think there's no easy way to preview embedded charts from deployment previews, but – there are two small-ish charts on the homepage that seem to work correctly (50% of chart width as a breakpoint) 🎉

app/charts/shared/interaction/tooltip-box.tsx
```diff
@@ -82,7 +106,10 @@ export const getCenteredTooltipPlacement = (props: {
         y: "top",
       }
     : {
-        x: xAnchor < chartWidth * 0.25 ? "right" : "left",
+        x: foldContainerSize<Xplacement>(chartWidth)({
+          xs: (w) => (xAnchor < w * 0.5 ? "right" : "left"),
+          md: (w) => (xAnchor < w * 0.25 ? "right" : "left"),
+        }),
```
ptbrowne, line 112: I feel like the logic is a bit brittle, we are never sure that this will work correctly right & we must rely on the > 0.5 & 0.25 heuristics. For me it would be better the Tooltip from MUI here.

* It is already included in the bundle (see datatable file)
* It will correctly check (through Popper) the boundary of the container and position the Tooltip accordingly
* It would allow us to remove a lof the code here.

lloydrichards: my resistance to replacing the Tooltip with the MUI version is that the two APIs work very differently with the MUI wrapping the target, and the current one taking the position from the  state of `useInteraction()`.  I think this is mostly in order to be able to follow the cursor in some instances, and in others to lock to something like a bar or point in a chart.

Seeing as the Tooltip is used in all 13 chart types with multiple variants and conditional logic, I think the effort to replace all the behaviours with the MUI version will take considerable effort (especially from someone new to the project)

ptbrowne: Fair point, you're right 👍  

In the future, I would guess but not sure that we could make either Popper or Tooltip work directly maybe without wrapping.

Nit: For what it's worth, I had trouble understanding the fold code at first glance. I find that we could simplify the logic a bit.

```
const isSmall = containerWidth < 600;
const leftThreshold = (isSmall ? 0.5 : 0.25) * containerWidth
const placement = xAnchor < leftThreshold ? 'right' : 'left'
```

lloydrichards: i agree that when its a binary option like at the moment with small and larger screens then the fold seems excessive, but I want to make sure the solution is robust, so in the future (or if there needs to be edge cases) then its possible to add more options without the need for a switch:
```ts
x: foldContainerSize<Xplacement>(chartWidth)({
          xs: (w) => (xAnchor < w * 0.5 ? "right" : "left"), // places tooltip on opposite side of screen
          md: (w) => (xAnchor < w * 0.25 ? "right" : "left"),  // on medium size charts, places based on 1/4 ratio
          lg: (w) => "right", // on large size charts, places always on right
        }),
```

ptbrowne: OK, fair and my point is a nit. I think that shaping the code so that breakpoints (especially more than 2) are used in the future should not be a goal, as usually I find breakpoints make things more brittle than sturdy.

PR #1272, opened 2023-11-16
feat: Add a way to retrieve observations from multiple cubes

bprusinowski:
This PR adapts the application logic to support retrieving observations from multiple cubes at once (merging them along the way).

### GQL / SPARQL
- Removed `DataCubeObservations` query (replaced by `DataCubesObservations`)
- Observations queried from multiple cubes are now merged on the server-side by joinBy dimensions (assumed to be temporal dimensions of the same unit in the first iteration). This would need to be guaranteed by the front-end
- Removed unused variables (`rawObservations`, `sparql`) from observations query
- Introduced separate filter types per query (components, metadata, observations)
- Optimized hierarchy query to not re-fetch dimension values in case no filters were applied

---
## Context
This PR is a third part of preparations for merging the cubes. For the technical concept, see [this Notion document](https://www.notion.so/interactivethings/Combine-two-cubes-in-the-same-chart-d5711b414a7747e79edfc3cfe17d2c9b).

- [x] `DataCubesComponents`
- [x] `DataCubesMetadata`
- [x] `DataCubesObservations`

## How to test
1. Click around in the application and make sure nothing is broken 💥 
2. Open [GQL playground](https://visualization-tool-git-feat-multiple-cubes-observations-ixt1.vercel.app/api/graphql) and send e.g. the following query to see that observations from both cubes was fetched (and merged when encountering the same value from joinBy dimensions).

**Query**
```graphql
query DataCubesObservations(
  $sourceType: String!
  $sourceUrl: String!
  $locale: String!
  $filters: [DataCubeObservationFilter!]!
) {
  dataCubesObservations(
    sourceType: $sourceType
    sourceUrl: $sourceUrl
    locale: $locale
    filters: $filters
  )
}
```

**Variables**
```js
{
  "sourceType": "sparql",
  "locale": "en",
  "filters": [
    {
      "iri": "https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/9",
      "latest": true,
      "filters": {
        "https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/Kanton": {
          "type": "single",
          "value": "https://ld.admin.ch/canton/1"
        }
      },
      "joinBy":"https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/Jahr"
    }
,
    {
      "iri": "https://energy.ld.admin.ch/sfoe/bfe_ogd18_gebaeudeprogramm_co2wirkung/4",
      "filters": {
        "https://energy.ld.admin.ch/sfoe/bfe_ogd18_gebaeudeprogramm_co2wirkung/region": {
          "type": "single",
          "value": "https://ld.admin.ch/country/CHE"
        },
        "https://energy.ld.admin.ch/sfoe/bfe_ogd18_gebaeudeprogramm_co2wirkung/massnahmenbereich": {
          "type": "single",
          "value": "https://energy.ld.admin.ch/sfoe/bfe_ogd18_gebaeudeprogramm_co2wirkung/ogd18_catalog/Haustechnik"
        }
      },
      "joinBy":"https://energy.ld.admin.ch/sfoe/bfe_ogd18_gebaeudeprogramm_co2wirkung/Jahr"
    } 
  ]
}
```

### DIFF
--- a/app/graphql/resolvers/rdf.ts
+++ b/app/graphql/resolvers/rdf.ts
@@ -16,6 +16,7 @@ import { truthy } from "@/domain/types";
 import { Loaders } from "@/graphql/context";
 import {
   DataCubeComponentFilter,
+  DataCubeObservationFilter,
   DataCubeResolvers,
   DimensionResolvers,
   QueryResolvers,
@@ -328,6 +329,8 @@ export const dataCubesObservations: NonNullable<
   }
 
   const { loaders, sparqlClient, cache } = await setup(info);
+  // If the cube was updated, we need to also update the filter with the correct iri.
+  const filtersWithCorrectIri: DataCubeObservationFilter[] = [];
 
   const data: Observation[] = [];
   const sparqlEditorUrls: DataCubesObservations["sparqlEditorUrls"] = [];
@@ -344,6 +347,11 @@ export const dataCubesObservations: NonNullable<
       const cube = latest ? await getLatestCube(rawCube) : rawCube;
       // TODO: optimize to avoid fetching the shape at all
       await cube.fetchShape();
+      filtersWithCorrectIri.push({
+        ...filter,
+        iri: cube.term?.value!,
+      });
+
       const { query, observations } = await getCubeObservations({
         cube,
         locale,
--- a/app/graphql/resolvers/rdf.ts
+++ b/app/graphql/resolvers/rdf.ts
@@ -253,7 +253,10 @@ export const dataCubesComponents: NonNullable<
               locale,
               sparqlClient,
               sparqlClientStream,
-              cache
+              cache,
+              // Only pass values if there are no filters, as we need to fetch
+              // the full, not filtered hierarchy.
+              cubeFilters?.filters ? undefined : values
             )
           : null;
         const baseDimension = {
--- a/app/graphql/resolvers/rdf.ts
+++ b/app/graphql/resolvers/rdf.ts
@@ -323,6 +323,10 @@ export const dataCubesMetadata: NonNullable<
 export const dataCubesObservations: NonNullable<
   QueryResolvers["dataCubesObservations"]
 > = async (_, { locale, filters }, { setup }, info) => {
+  if (filters.length > 1 && filters.some((f) => f.joinBy === undefined)) {
+    throw new Error("Can't query multiple cubes observations without joinBy!");
+  }
+
   const { loaders, sparqlClient, cache } = await setup(info);
 
   const data: Observation[] = [];
--- a/app/graphql/resolvers/rdf.ts
+++ b/app/graphql/resolvers/rdf.ts
@@ -5,9 +5,11 @@ import { LRUCache } from "typescript-lru-cache";
 
 import { Filters } from "@/configurator";
 import {
+  DataCubesObservations,
   Dimension,
   DimensionValue,
   Measure,
+  Observation,
   TemporalDimension,
 } from "@/domain/data";
 import { truthy } from "@/domain/types";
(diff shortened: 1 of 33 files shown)


app/graphql/resolvers/rdf.ts
```diff
@@ -314,6 +322,107 @@ export const dataCubesMetadata: NonNullable<
...
+  if (filters.length > 1 && filters.some((f) => f.joinBy === undefined)) {
+    throw new Error("Can't query multiple cubes observations without joinBy!");
+  }
+
+  const { loaders, sparqlClient, cache } = await setup(info);
+  // If the cube was updated, we need to also update the filter with the correct iri.
+  const filtersWithCorrectIri: DataCubeObservationFilter[] = [];
+
+  const dataByCubeIri: Record<string, Observation[]> = {};
+  const sparqlEditorUrls: DataCubesObservations["sparqlEditorUrls"] = [];
```
lloydrichards, line 337: Not a fan of instantiating empty arrays/props and relying on loops and callbacks in order to fill out the values.  I think with `Promise.all()` there is a risk of partial failure or skipped steps that might make this harder to test.  I think rather having side-effects within the promise all its easier to understand if you return values and rely on functors like map, reduce, filter etc.

app/graphql/resolvers/rdf.ts
```diff
@@ -314,6 +322,107 @@ export const dataCubesMetadata: NonNullable<
   );
 };
 
+export const dataCubesObservations: NonNullable<
+  QueryResolvers["dataCubesObservations"]
+> = async (_, { locale, filters }, { setup }, info) => {
```
lloydrichards, line 327: It seems like there are quite a few places where this function can throw an error.  I'm not sure what the suspense boundary looks like above this, but maybe these errors could be caught within this function and returned a little more gracefully?  Unless graphQL does a good job at handeling all these thrown errors?

lloydrichards (commented):
I can't speak to the effect of implementation since I'm not experienced enough with how the cubes are fetched, but nothing jumps out from the changes to `useDataCubesObservationsQuery()` within the chart components.

Only place where i see concern is the style in which the `dataCubesObservations()` is structures as it relies heavily on loops and iteration which can be harder to unit test and expand upon in the future as they rely on side-effect.  Have added some comments, but these are more architectural concerns then suggestions.

app/graphql/resolvers/rdf.ts
bprusinowski: When an error is thrown here, like the `"Can't query multiple cubes observations without joinBy!"`, we would see it in an error banner in the place of the chart with this exact message. I feel like handling this _correctly_ would be a part of bigger change as discussed on Wednesday; I think I'd prefer to first have a clear picture in mind how to handle such errors in general (in this case, server-side errors) if that makes sense 😄

app/graphql/resolvers/rdf.ts
bprusinowski: Yup, good point! It would work in the case of the below `Promise.all`, however I think it would be trickier to change the approach when getting `mergedObservations`, as there we need to access the "past" value in current iteration to overwrite it "outside" of the loop we're in. I'll refactor the `Promise.all` and see if there are other approaches to deal with the second case 👍

bprusinowski:
Thanks for reviewing @lloydrichards! I am going to merge this PR introduce the changes in a new one (together with suggestions from @ptbrowne on improvements to the caching).

PR #1273, opened 2023-11-17
refactor: Multi-cube ChartConfig

### LINKED ISSUE #1267: perf: regression for GQL `possibleFilters`
**Performance Regression**
yesterdays merges to main triggered a performance serious regression for GQL `possibleFilters`
image

**To Reproduce**

1. measure timing
    ```
    curl https://test.visualize.admin.ch/api/graphql -k --resolve test.visualize.admin.ch:443:193.246.94.41 -X POST -H 'content-type: application/json' -H 'x-visualize-cache-control: no-cache' -H 'Pragma: no-cache' -H 'Cache-Control: no-cache' --data-raw '{"query":"query PossibleFilters($iri: String!, $sourceType: String!, $sourceUrl: String!, $filters: Filters!) {\n  possibleFilters(\n    iri: $iri\n    sourceType: $sourceType\n    sourceUrl: $sourceUrl\n    filters: $filters\n  ) {\n    iri\n    type\n    value\n    __typename\n  }\n}\n","operationName":"PossibleFilters","variables":{"iri":"https://environment.ld.admin.ch/foen/nfi/nfi_C-90/cube/2023-3","sourceType":"sparql","sourceUrl":"https://lindas.admin.ch/query","filters":{"https://environment.ld.admin.ch/foen/nfi/unitOfReference":{"type":"single","value":"https://ld.admin.ch/canton/23"},"https://environment.ld.admin.ch/foen/nfi/classificationUnit":{"type":"single","value":"https://environment.ld.admin.ch/foen/nfi/ClassificationUnit/Total"},"https://environment.ld.admin.ch/foen/nfi/inventory":{"type":"single","value":"https://environment.ld.admin.ch/foen/nfi/Inventory/150"},"https://environment.ld.admin.ch/foen/nfi/unitOfEvaluation":{"type":"single","value":"https://environment.ld.admin.ch/foen/nfi/UnitOfEvaluation/2382"},"https://environment.ld.admin.ch/foen/nfi/evaluationType":{"type":"single","value":"https://environment.ld.admin.ch/foen/nfi/EvaluationType/1"}},"filterKey":"https://environment.ld.admin.ch/foen/nfi/unitOfReference, https://environment.ld.admin.ch/foen/nfi/classificationUnit, https://environment.ld.admin.ch/foen/nfi/inventory, https://environment.ld.admin.ch/foen/nfi/unitOfEvaluation, https://environment.ld.admin.ch/foen/nfi/evaluationType"}}' -s -o /dev/null -w '%{time_total}'
    ```
2. observe increase (+1 s) compared to reference performance (~ 850 ms)

**Expected behavior**
same or better performance

- Visualize environment and version: TEST

@bprusinowski can you check the reason for this regression and provide a solution ?

cc @ortnever @zellersabine @claudiodigallo @adintegra @sosiology

### LINKED ISSUE #1274: `possibleFilters` broken
**Describe the bug**
can no longer add a filter

**To Reproduce**

1. Go to https://test.visualize.admin.ch/create/new?cube=https://environment.ld.admin.ch/foen/gefahren-waldbrand-warnung/1
2. Click on 'add filter' - canton
4. See no filter appeares

**Expected behavior**
a list of cantons appearing

- Visualize environment and version:
  - TEST v3.24.1 ([d72b50b](https://github.com/visualize-admin/visualization-tool/commit/d72b50b0981e8011eb84f76c7cf068950e051eb3))
  - INT v3.24.1 ([e60bae8](https://github.com/visualize-admin/visualization-tool/commit/e60bae877eeafea028ab2c08218c73e54ed39bcf))

**Additional context**
possibly related #1267

bprusinowski:
Fixes #1267.
Fixes #1274.

This PR:
- changes the `ChartConfig` type to be based on multiple cubes instead of one, and introduces the first iteration of the cube merging feature,
- fixes the `PossibleFilters` query by bringing back the old way of adding LIMIT to the observations query, as it turned out that the new preview mode from the `cube-view-query` library doesn't work with filters,
- improves caching of fetching multiple cubes by merging cube-scoped queries on the client side (`DataCubesMetadata`, `DataCubesComponents`, `DataCubesObservations`),
- refactors merging of cubes algorithm to use reduce instead of pushing to external array.

### How to test
1. Go to [this link](https://visualization-tool-git-refactor-multi-cube-chart-config-ixt1.vercel.app/en/create/new?cube=https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/9&dataSource=Prod).
2. Open web inspector and find the [LocalStorage](https://developer.chrome.com/docs/devtools/storage/localstorage/#:~:text=Open%20DevTools%20on%20the%20website,the%20table%2C%20select%20a%20pair.).
3. Add [a new entry](https://developer.chrome.com/docs/devtools/storage/localstorage/#create) to the LocalStorage with the following information.

**Key**
```js
vizualize-configurator-state:IW-jMUdmG_u5
```

**Value**
```js
{"version":"3.0.1","state":"CONFIGURING_CHART","dataSource":{"type":"sparql","url":"https://lindas.admin.ch/query"},"meta":{"title":{"de":"","fr":"","it":"","en":""},"description":{"de":"","fr":"","it":"","en":""}},"chartConfigs":[{"key":"O7oRFeSgxhpT","version":"3.0.0","meta":{"title":{"en":"","de":"","fr":"","it":""},"description":{"en":"","de":"","fr":"","it":""}},"cubes":[{"iri":"https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/9","filters":{"https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/Kanton":{"type":"single","value":"https://ld.admin.ch/canton/1"}},"joinBy":"https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/Jahr"},{"iri":"https://energy.ld.admin.ch/sfoe/bfe_ogd18_gebaeudeprogramm_co2wirkung/4","filters":{"https://energy.ld.admin.ch/sfoe/bfe_ogd18_gebaeudeprogramm_co2wirkung/region":{"type":"single","value":"https://ld.admin.ch/country/CHE"},"https://energy.ld.admin.ch/sfoe/bfe_ogd18_gebaeudeprogramm_co2wirkung/massnahmenbereich":{"type":"single","value":"https://energy.ld.admin.ch/sfoe/bfe_ogd18_gebaeudeprogramm_co2wirkung/ogd18_catalog/Haustechnik"}},"joinBy":"https://energy.ld.admin.ch/sfoe/bfe_ogd18_gebaeudeprogramm_co2wirkung/Jahr"}],"chartType":"comboLineDual","interactiveFiltersConfig":{"legend":{"active":false,"componentIri":""},"timeRange":{"componentIri":"https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/Jahr","active":false,"presets":{"type":"range","from":"2014","to":"2022"}},"dataFilters":{"active":false,"componentIris":[]},"calculation":{"active":false,"type":"identity"}},"fields":{"x":{"componentIri":"https://energy.ld.admin.ch/sfoe/bfe_ogd18_gebaeudeprogramm_co2wirkung/Jahr"},"y":{"leftAxisComponentIri":"https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/InstallierteLeistungkW","rightAxisComponentIri":"http://schema.org/amount","palette":"category10","colorMapping":{"https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/InstallierteLeistungkW":"#1f77b4","http://schema.org/amount":"#ff7f0e"}}}}],"activeChartKey":"O7oRFeSgxhpT"}
```

4. Go to [this link](https://visualization-tool-git-refactor-multi-cube-chart-config-ixt1.vercel.app/en/create/IW-jMUdmG_u5?dataSource=Prod).
5. See that the chart shows two lines – each one coming from a different dataset.
6. _Note that if you e.g. change chart type or some options of the chart, it will get broken. There is still a need to do some work for the feature to be properly implemented, including UI elements and additional logic._

Screenshot 2023-11-21 at 16 42 49

---
## Context
This PR introduces the first iteration of the cube merging feature. For the technical concept, see [this Notion document](https://www.notion.so/interactivethings/Combine-two-cubes-in-the-same-chart-d5711b414a7747e79edfc3cfe17d2c9b).

### DIFF
--- a/app/graphql/hooks.ts
+++ b/app/graphql/hooks.ts
@@ -20,72 +20,87 @@ const useQueryKey = (options: object) => {
   }, [options]);
 };
 
-export const useDataCubesMetadataQuery = (options: {
+type DataCubesMetadataOptions = {
   variables: Omit<DataCubeMetadataQueryVariables, "cubeFilter"> & {
     cubeFilters: DataCubeMetadataFilter[];
   };
   pause?: boolean;
-}) => {
-  const { variables, pause } = options;
+};
+
+export const executeDataCubesMetadataQuery = async (
+  variables: DataCubesMetadataOptions["variables"],
+  /** Callback triggered when data fetching starts (cache miss). */
+  onFetching?: () => void
+) => {
   const { locale, sourceType, sourceUrl, cubeFilters } = variables;
+
+  const queries = await Promise.all(
+    cubeFilters.map((cubeFilter) => {
+      const cubeVariables = { locale, sourceType, sourceUrl, cubeFilter };
+      const cached = client.readQuery<
+        DataCubeMetadataQuery,
+        DataCubeMetadataQueryVariables
+      >(DataCubeMetadataDocument, cubeVariables);
+
+      if (cached) {
+        return Promise.resolve(cached);
+      }
+
+      onFetching?.();
+
+      return client
+        .query<DataCubeMetadataQuery, DataCubeMetadataQueryVariables>(
+          DataCubeMetadataDocument,
+          cubeVariables
+        )
+        .toPromise();
+    })
+  );
+  const error = queries.find((q) => q.error)?.error;
+  const fetching = !error && queries.some((q) => !q.data);
+
+  return {
+    data: fetching
+      ? undefined
+      : { dataCubesMetadata: queries.map((q) => q.data!.dataCubeMetadata) },
+    error,
+    fetching,
+  };
+};
+
+export const useDataCubesMetadataQuery = (
+  options: DataCubesMetadataOptions
+) => {
   const [result, setResult] = React.useState<{
     data?: { dataCubesMetadata: DataCubeMetadata[] };
     error?: Error;
     fetching: boolean;
-  }>({ fetching: !pause });
+  }>({ fetching: !options.pause });
 
   const queryKey = useQueryKey(options);
 
+  const executeQuery = React.useCallback(
+    async (options: DataCubesMetadataOptions) => {
+      const result = await executeDataCubesMetadataQuery(
+        options.variables,
+        () => setResult((prev) => ({ ...prev, fetching: true }))
+      );
+      setResult(result);
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [queryKey]
+  );
+
   React.useEffect(() => {
-    if (pause) {
+    if (options.pause) {
       return;
     }
 
-    const run = async () => {
-      const queries = await Promise.all(
-        cubeFilters.map((cubeFilter) => {
-          const cubeVariables = { locale, sourceType, sourceUrl, cubeFilter };
-          const cached = client.readQuery<
-            DataCubeMetadataQuery,
-            DataCubeMetadataQueryVariables
-          >(DataCubeMetadataDocument, cubeVariables);
-
-          if (cached) {
-            return Promise.resolve(cached);
-          }
-
-          if (!result.fetching) {
-            setResult({
-              ...result,
-              fetching: true,
-            });
-          }
-
-          return client
-            .query<DataCubeMetadataQuery, DataCubeMetadataQueryVariables>(
-              DataCubeMetadataDocument,
-              cubeVariables
-            )
-            .toPromise();
-        })
-      );
-      const error = queries.find((q) => q.error)?.error;
-      const fetching = !error && queries.some((q) => !q.data);
-
-      setResult({
-        data: fetching
-          ? undefined
-          : { dataCubesMetadata: queries.map((q) => q.data!.dataCubeMetadata) },
-        error,
-        fetching,
-      });
-    };
-
-    run();
+    executeQuery(options);
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [queryKey]);
 
-  return [result];
+  return [result, executeQuery] as const;
 };
 
 type DataCubesComponentsOptions = {
@@ -114,9 +129,7 @@ export const executeDataCubesComponentsQuery = async (
         return Promise.resolve(cached);
       }
 
-      if (onFetching) {
-        onFetching();
-      }
+      onFetching?.();
 
       return client
         .query<DataCubeComponentsQuery, DataCubeComponentsQueryVariables>(
--- a/app/graphql/hooks.ts
+++ b/app/graphql/hooks.ts
@@ -1,0 +1,73 @@
+import React from "react";
+
+import { DataCubeMetadata } from "@/domain/data";
+
+import { client } from "./client";
+import {
+  DataCubeMetadataDocument,
+  DataCubeMetadataFilter,
+  DataCubeMetadataQuery,
+  DataCubeMetadataQueryVariables,
+} from "./query-hooks";
+
+const useQueryKey = (options: object) => {
+  return React.useMemo(() => {
+    return JSON.stringify(options);
+  }, [options]);
+};
+
+export const useDataCubesMetadataQuery = (options: {
+  variables: Omit<DataCubeMetadataQueryVariables, "cubeFilter"> & {
+    cubeFilters: DataCubeMetadataFilter[];
+  };
+  pause?: boolean;
+}) => {
+  const { variables, pause } = options;
+  const { locale, sourceType, sourceUrl, cubeFilters } = variables;
+  const [result, setResult] = React.useState<{
+    data?: { dataCubesMetadata: DataCubeMetadata[] };
+    error?: Error;
+    fetching: boolean;
+  }>({ fetching: !pause });
+
+  const queryKey = useQueryKey(options);
+
+  React.useEffect(() => {
+    if (pause) {
+      return;
+    }
+
+    const run = async () => {
+      const queries = await Promise.all(
+        cubeFilters.map((cubeFilter) =>
+          client
+            .query<DataCubeMetadataQuery, DataCubeMetadataQueryVariables>(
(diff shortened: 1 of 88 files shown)


app/graphql/hooks.ts
```diff
@@ -39,19 +39,31 @@ export const useDataCubesMetadataQuery = (options: {
...
+          if (cached) {
+            return Promise.resolve(cached);
+          }
+
+          if (!result.fetching) {
+            setResult({
+              ...result,
+              fetching: true,
+            });
+          }
```
ptbrowne: I am wondering why we need a setResult inside each promise ? I think it would be only necessary to have one outside the map ?

Why do we need to read explicitly from the cache ? I thought that client.query would do that for us ?

ptbrowne: Ah is it so that we do not have a flash of "loading" when all the queries are already in the cache ? I guess the urql way here would be to pipe the client.query result and subscribe to some kind of "request" event from it, setting the fetching state there. I think this "request" event would not be fired when the query would come from the cache.

bprusinowski: It was done to properly set the `fetching` property – it we read all queries from cache, I think we shouldn't set fetching to `true`, to avoid flashes of loading indicators – and when at least one query is not read from cache, then we set the fetching directly from the promise, as otherwise we would have awaited them all and not set the `fetching = true` at all.

Does it make sense? 😅 Basically instead of always having `setFetching(true)` at the top, I wanted to only set this when we actually didn't hit the cache 👀

ptbrowne: That said, maybe the urql way is not very understandable either, I have trouble exactly making sense of it in their source. https://github.com/urql-graphql/urql/blob/main/packages/react-urql/src/hooks/useQuery.ts#L346

bprusinowski: And yes, it would also be read from cache when using `client.query`, but then it was easier to do `readQuery` so we know when to set the `fetching` prop

app/graphql/hooks.ts
```diff
@@ -186,3 +195,114 @@ export const useDataCubesComponentsQuery = makeUseQuery<
...
+  >((acc, query) => {
+    const cubeIri = query.operation.variables?.cubeFilter.iri;
+    const url = query.data?.dataCubeObservations?.sparqlEditorUrl;
+
+    if (cubeIri && url) {
+      acc.push({ cubeIri, url });
+    }
+
+    return acc;
+  }, []);
```
ptbrowne, line 294: Nit: I would do a filter & a map in those cases.

app/graphql/hooks.ts
```diff
@@ -186,3 +195,114 @@ export const useDataCubesComponentsQuery = makeUseQuery<
...
+            }
+
+            const existing = acc[key];
+            acc[key] = Object.assign(existing ?? {}, observation);
+          }
+
+          return acc;
+        }, {})
+      : // If we are fetching data from a single cube, we can just return the data
+        (queries[0].data?.dataCubeObservations?.data as Observation[]);
```
ptbrowne, line 282: after I see that we do Object.values on the observations so it seems like in the case of multiple cubes, observations is an index, while when it is 1 cube, it is only an array ?

app/graphql/hooks.ts
```diff
@@ -186,3 +195,114 @@ export const useDataCubesComponentsQuery = makeUseQuery<
...
+  return {
+    data: {
+      dataCubesObservations: {
+        data: Object.values(observations),
+        sparqlEditorUrls,
+      },
+    },
+    error,
+    fetching,
+  };
```
ptbrowne, line 305: I would extract the merging into another function that could be unit tested.

app/graphql/hooks.ts
```diff
@@ -172,20 +173,22 @@ export const executeDataCubesComponentsQuery = async (
...
+              (acc, query) => {
+                const { dimensions, measures } =
+                  query.data?.dataCubeComponents!;
+                acc.dimensions.push(...dimensions);
+                acc.measures.push(...measures);
+
+                return acc;
+              },
+              { dimensions: [], measures: [] }
+            ),
```
ptbrowne, line 190:
```
dataCubesComponents: {
  dimensions: queries.flatMap(q => q.data?.dataCubeComponents?.dimensions),
  measures: queries.flatMap(q => q.data?.dataCubeComponents?.measures),
}
```

app/graphql/hooks.ts
bprusinowski: Yes! In this case we can call Object.values on both types to get what we need (for array it will return the same array) 👍 But maybe it could be a bit more explicit :)

ptbrowne: Oh I did not know that Object.values was working for arrays. TIL.

Nonetheless, I think it is always good if the type of a variables does not change across code branches, maybe we could call Object.values a bit earlier so that we are sure that observations is always an array ?

ptbrowne (approved):
I have no blocker, but I think it would be good to add some unit tests, particularly around query merging.

LGTM 👍 

On a form level, I think reduce is a bit too present. I find sometimes that is more readable to use other forms (filter then map, or flatMap). This is not a blocker and can be a matter of taste. Iguess.

PR #1281, opened 2023-11-24
refactor: Common joinBy dimension

bprusinowski:


app/configurator/configurator-state.tsx
```diff
@@ -1197,28 +1196,38 @@ const reducer: Reducer<ConfiguratorState, ConfiguratorStateAction> = (
...
+        if (chartConfig.interactiveFiltersConfig) {
+          chartConfig.interactiveFiltersConfig.timeRange = {
+            componentIri: dimension.iri,
+            active: chartConfig.interactiveFiltersConfig.timeRange.active,
+            presets: {
+              type: "range",
+              from,
+              to,
+            },
+          };
```
ptbrowne, line 1230: Can we have a unit test here ?

app/charts/shared/chart-helpers.tsx
```diff
@@ -212,12 +248,16 @@ export const extractChartConfigComponentIris = (chartConfig: ChartConfig) => {
...
-      Boolean
+  return (
+    uniq(
+      [...fieldIris, ...additionalFieldIris, ...filterIris, ...IFIris].filter(
+        Boolean
+      )
     )
-    // Important so the order is consistent when querying.
-  ).sort();
+      .filter((d) => d !== "joinBy")
```
ptbrowne, line 257: Nit: Maybe, this should be a constant ?

ptbrowne (approved):
LGTM. I do not see any blocker, and I understand the general logic, thanks a lot for the clear explanation.

I understand the nitty gritty details a bit less. I think that we could benefit from a bit more unit test on the joinBy logic and on the parts that we have changed.

bprusinowski:
Thanks for reviewing @ptbrowne, I'll tackle the comments and add tests 👍

app/charts/shared/chart-helpers.tsx
ptbrowne: At least, when changing to multiple joinBy, you can remove the constant and see directly the affected areas.

PR #1285, opened 2023-11-30
fix: Better scope pragmas to fix data cube preview query

bprusinowski:
This PR attempts to better scope pragmas when fetching the cube preview by using a clean Source, without pragmas, as having `#pragma join.hash off` breaks previews for some cubes for yet unknown reasons (see e.g. https://test.visualize.admin.ch/en/browse?previous=%7B%7D&dataset=https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fnfi%2Fnfi_C-94%2Fcube%2F2023-1&dataSource=Prod).

### Context
For more information see [this conversation on Zulip](https://zulip.zazuko.com/#narrow/stream/40-bafu-ext/topic/performance/near/377890).

### Notes
I aimed for a simple solution by overriding `getMainSource`  View's method, which might not be ideal, but:
- I didn't see a better, simple way of doing so (we can't just pass a clean Source into the View constructor, as [this is how it's retrieved internally](https://github.com/zazuko/rdf-cube-view-query/blob/9acb62a2bd5e50a4febf8f871f3e52935c151a62/lib/View.js#L403-L417) – taking the source of the first dimension, instead of the one tied to Source),
- this might be a temporary workaround until Stardog releases a new version (https://zulip.zazuko.com/#narrow/stream/40-bafu-ext/topic/performance/near/378807).

### How to test
1. Go to the deployment preview.
2. Look at the homepage examples to see if they load fast (<5 s).
3. Go to dataset browse page.
4. Change data source to PROD.
5. Search for a `NFI: Topics by ownership` cube.
6. Open the preview.
7. See that only 10 observations are loaded.
8. See [the same cube preview on TEST](https://test.visualize.admin.ch/en/browse?previous=%7B%7D&dataset=https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fnfi%2Fnfi_C-94%2Fcube%2F2023-1&dataSource=Prod) and see that 1530 observations are loaded there.

### DIFF
--- a/app/components/debug-panel/DebugPanel.tsx
+++ b/app/components/debug-panel/DebugPanel.tsx
@@ -125,8 +125,7 @@ const DebugConfigurator = () => {
             size="small"
             href={`${sparqlEditorUrl}#query=${encodeURIComponent(
               `#pragma describe.strategy cbd
-               #pragma join.hash off
-              
+
                DESCRIBE <${cube.iri}>`
             )}&requestMethod=POST`}
             target="_blank"
--- a/app/graphql/context.tsx
+++ b/app/graphql/context.tsx
@@ -10,9 +10,9 @@ import { LRUCache } from "typescript-lru-cache";
 import { SPARQL_GEO_ENDPOINT } from "@/domain/env";
 import { Awaited } from "@/domain/types";
 import { Timings } from "@/gql-flamegraph/resolvers";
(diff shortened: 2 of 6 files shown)


ptbrowne (approved):
Thanks, it is somewhat clear to me, and the comment makes it explicit why this was added 👍

PR #1286, opened 2023-12-01
feat: Landing page can be localized

bprusinowski:
This PR adds support for localized landing pages (e.g. new BLW cubes).

As it looks like these are the only cubes that currently support localized landing pages, I kept support for non-localized ones by falling back to non-translated properties.

Also see https://zulip.zazuko.com/#narrow/stream/40-bafu-ext/topic/blw-cubes/near/379002

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ You can also check the [release page](https://github.com/visualize-admin/visuali
 
 ## Unreleased
 
-Nothing yet.
+- Features
+  - Localized cube landing pages are now supported (dcat:landingPage) 🌎
 
 # [3.24.2] - 2023-11-28
 
--- a/app/rdf/query-cube-metadata.ts
+++ b/app/rdf/query-cube-metadata.ts
@@ -140,7 +140,10 @@ export const getCubeMetadata = async (
       ?contactPoint ${ns.vcard.hasEmail} ?contactPointEmail .
     }
     OPTIONAL { ?iri ${ns.dcterms.publisher} ?publisher . }
-    OPTIONAL { ?iri ${ns.dcat.landingPage} ?landingPage . }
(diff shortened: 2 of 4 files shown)


ptbrowne (approved):
👍 Looks good to me.

Not a blocker but I find that we could add a small unit test and it would help even as a reviewer to understand the effect (for example of the fallbackToNonLocalized). Here I could understand but I cannot clearly build the query in my head.

kronmar:
Really not sure, if `dcat:landingPage` should be localized. At least the [opendata.swiss handbook](https://handbook.opendata.swiss/de/content/glossar/bibliothek/dcat-ap-ch.html) expects a URI and not a string. At least if the cube is expected to be published on opendata.swiss. Maybe there is a different way to deal with that?

bprusinowski:
Hi @kronmar, thanks for your comment! This (localized landing page) was the format in which the most recent BLW cubes were provided to us – in order to support them, we needed to enable this change, otherwise the data couldn't even load.

I think that a change would be needed on the data side, or maybe it could be possible to try to extend ODS schema to support localized links (I think it makes sense, first example to benefit from it would be e.g. [Bathing water quality dataset](https://visualize.admin.ch/en/browse?previous=%7B%22includeDrafts%22%3Afalse%2C%22order%22%3A%22SCORE%22%2C%22search%22%3A%22bath%22%7D&dataset=https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fubd0104%2F6&dataSource=Prod) – see that even though we are looking at English version, the "Learn more about the dataset link" directs to German BAFU page, while there are all languages available).

I guess then the BLW cubes would have problems to be correctly displayed on ODS? Maybe there is a way to include both localized and non-localized property for cubes, so they work with both tools? I feel like this should be discussed on more LINDAS / Zazuko level 👀

Let me know what you think 👍 

cc @sosiology

PR #1291, opened 2023-12-04
feat: Dynamic most recent date filter

bprusinowski:


bprusinowski:
Sure, will do! :)

app/configurator/components/chart-configurator.tsx
```diff
@@ -231,6 +223,21 @@ const useEnsurePossibleFilters = ({
...
+            value.type === "single" &&
+            isDynamicMaxValue(value.value) &&
+            filters[key]
+          ) {
+            filters[key] = {
+              type: "single",
+              value: `${value.value}`,
+            };
+          }
+        }
```
ptbrowne, line 239: Not sure I understand this 😓 Can you give me an example, or add a test 😁 ?

app/configurator/components/field.tsx
```diff
@@ -252,94 +258,132 @@ export const DataFilterSelect = ({
...
+            }
+          />
+          {/* FIXME: adapt to design */}
+          <FormGroup>
+            <FormControlLabel
+              control={
+                <MUISwitch
+                  checked={usesMostRecentDate}
+                  onChange={() =>
+                    fieldProps.onChange({
```
ptbrowne, line 360: Nit: Instead of relying on the state usesMOstRecentDate, I think it is preferrable to rely on the `checked` value coming in the 2nd argument of the callback. This way the function could be stable, and I think it'd be clearer since here you have to think "ok usesMostRecentDate is true, so when it is clicked it shuold be false, and the max date should be used". IN the callback, the checked parameter is already right, so I think it's clearer.

app/rdf/queries.ts
```diff
@@ -784,103 +789,130 @@ export const hasHierarchy = (dim: CubeDimension) => {
...
+            // query, but it doesn't seem to be supported by the cube-view-query
+            const maxValueQuery = SELECT`(MAX(?value) as ?value)`
+              .WHERE` <${cube.term?.value}> <https://cube.link/observationSet> ?observationSet .
+  ?observationSet <https://cube.link/observation> ?source .
+  ?source <${cubeDimension.path?.value}> ?value .`;
+            const maxValueRaw = await maxValueQuery.execute(
+              sparqlClient.query,
+              { operation: "postUrlencoded" }
+            );
+            const maxValue = maxValueRaw?.[0]?.value?.value as string;
```
ptbrowne, line 885: Function could be extracted for clarity ? It could aso be memoized ?

It made me think that if we'd use react-query, we could cache the sparql client results. Then I remember that we are using urql for it's graphql abilities 🤷

ptbrowne (commented):
LGTM 👏 One unit test on buildFilters could be good ?

app/rdf/queries.ts
bprusinowski: Yup, it can be extracted 👍 But for memoization I am not sure, how would you memoize it, I guess when it's extracted it's already "memoized"? (as we are not in the useMemo world here in the backend 😅)

app/configurator/components/chart-configurator.tsx
bprusinowski: I'll add a test, but just to explain: possible filters coming from the backend do not return `VISUALIZE_MAX_VALUE`, but rather an actual date (e.g. 2023), so if we didn't replace it back with `VISUALIZE_MAX_VALUE`, we would potentially trigger an update that shouldn't happen (as in fact, the pinned date didn't change, but is still equal to `VISUALIZE_MAX_VALUE`). Maybe it would make sense to already do this replacement on the backend side? 🤔

bprusinowski:
Thanks for the review @ptbrowne 🙇‍♂️ I'll try to tackle the changes this week, otherwise I'll make a note to myself to do it in the new year 😅

PR #1292, opened 2023-12-05
feat: Add layout step

bprusinowski:
This PR introduces the layouting step to the application when user created more than one chart at once in the editor mode (hidden behind `flag__layoutStep=true`). Also see [this Notion document](https://www.notion.so/interactivethings/Compositions-218e4f5e1b514a3e8930a7a8cab2092a).

- Added layouting step, conditionally enabled when there are at least two charts
- Separated panel layout for every step, as they do not share the same "left-middle" layout anymore and made it more configurable (`L`, `M`, `R`)
- Introduced `LayoutConfigurator`
  - Based on `ConfiguratorState`'s `layout` property
  - Made it possible to set layout annotations (title and description) by moving the meta property from `ConfiguratorState` to `layout`
  - Made it possible to select either Tab or Dashboard layout (with Vertical and Tall sub layouts for the Dashboard layout)
- Adapted paddings and general styles to match the design

### Next steps
- [ ] Tweak annotations fonts and background color in published charts
- [ ] Add translations for "Vertical", "Tall"
- [ ] To be confirmed: use `PanelHeader` in `CONFIGURING_CHART` step instead of keeping top buttons in drawer & chart tabs
- [ ] Check how to handle responsiveness of tall layout with @KerstinFaye
- [ ] Make axis titles responsive
- [ ] Made it possible to drag & drop charts in Dashboard layout to change their positions
- [ ] Add single URLs layout option

### How to test
1. Go to [deployment preview](https://vercel.live/open-feedback/visualization-tool-git-feat-add-layout-step-ixt1.vercel.app?via=pr-comment-visit-preview-link&passThrough=1).
2. Add `&flag__layoutStep=true` to the end of the browser URL and hit Enter.
3. Select a dataset and start making a chart.
4. See that you can publish the chart if you only made one chart.
5. Add another chart.
6. See that you have to proceed to layout options to publish your creation.
7. See that it's possible to add an overarching title and description to your creation.
8. See that it's possible to select from Tab or Dashboard layouts.
9. See that it's possible to select Vertical or Tall layout subtypes in the Dashboard layout.
10. Try to make more than 3 charts to see that Tall layout "recycles" its layout.


app/configurator/components/configurator.tsx
```diff
@@ -99,71 +131,170 @@ const ConfigureChartStep = () => {
...
+
+  useAssureCorrectDataSource("LAYOUTING");
+
+  if (state.state !== "LAYOUTING") {
+    return null;
+  }
+
+  return (
+    <InteractiveFiltersProvider>
+      <PanelLayout type="M" sx={{ background: (t) => t.palette.muted.main }}>
```
bprusinowski, line 217: Why `muted.main` didn't work?

app/configurator/components/icon-button.tsx
```diff
@@ -0,0 +1,82 @@
...
+    display: "flex",
+    flexDirection: "column",
+    justifyContent: "center",
+    alignItems: "center",
+    width: 91,
+    height: 64,
+    borderRadius: 8,
+    transition: theme.transitions.create(["background-color", "color"], {
+      duration: theme.transitions.duration.shorter,
+    }),
```
ptbrowne, line 21: TIL theme transitions. I liked your way of using shorthands in swiss federal ci. I now use 

```
({ palette: p, spacing: s, shadows: e, }) => {}
```

Maybe transitions could `t` :)

app/configurator/components/icon-button.tsx
```diff
@@ -0,0 +1,82 @@
...
+      value={value}
+      onClick={onClick}
+      disabled={disabled}
+      className={clsx(
+        classes.root,
+        disabled ? classes.disabled : null,
+        checked ? classes.checked : null
+      )}
+    >
+      <Icon size={24} name={getIconName(label)} />
```
ptbrowne, line 73: ~~Why not using startIcon on the Button ?~~ Understood now what it was :)

app/configurator/components/layout.tsx
```diff
@@ -1,13 +1,71 @@
...
+    position: "fixed",
+    top: HEADER_HEIGHT,
+    width: "100%",
+    height: `calc(100vh - ${HEADER_HEIGHT}px)`,
+    display: "grid",
+    gridTemplateRows: "auto minmax(0, 1fr)",
+    marginLeft: "auto",
+    marginRight: "auto",
+  },
+  MPanelLayout: {
```
ptbrowne, line 19: Maybe add just tiny docstrings ?

app/configurator/components/annotators.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+const useStyles = makeStyles<Theme, { interactive?: boolean }>({
+  text: {
+    marginBottom: 4,
+    cursor: ({ interactive }) => (interactive ? "pointer" : "text"),
+
+    "&:hover": {
+      textDecoration: ({ interactive }) => (interactive ? "underline" : "none"),
+    },
+  },
```
ptbrowne, line 32: Nit: not very important but I personally tend not to use params for this kind of thing, but only CSS classes. So have an `interactive` class making the change, and applying it conditionally. I find that it is closer to normal CSS, thus maybe more future proof.

app/configurator/components/drawer.tsx
```diff
@@ -10,8 +10,8 @@ export const ConfiguratorDrawer = styled(MuiDrawer)(({ theme }) => ({
   "& > .MuiPaper-root": {
     top: 96,
     bottom: 0,
+    width: DRAWER_WIDTH,
```
ptbrowne, line 13: Nit: It seems this Drawer width could be a CSS variable.

ptbrowne (commented):
👏 Very clear, no blocker on the code. The layout classes make sense, and it's good to have extracted some shared component like the IconButton.

When using the functionality, I had 1 wonder. I wonder if there is a good reason to switch to another "layouting step", instead of staying in the same view with a "Layout" part in the panel. I wonder about that since making the transition involves the whole page, which makes you lose a bit your context, and you lose for example your tab. 

I have seen that you updated the docs and was wondering on whether we should move to storybook, since I think in general we should use more Storybook :)

Also I think that the accordion for title/description should be directly opened in the layouting step.

app/locales/it/messages.po
```diff
@@ -1298,8 +1312,8 @@ msgid "publication.share.mail.subject"
 msgstr "visualize.admin.ch"
 
 #: app/configurator/components/chart-options-selector.tsx
-msgid "right"
-msgstr "Diritto"
+#~ msgid "right"
+#~ msgstr "Diritto"
```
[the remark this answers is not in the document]
bprusinowski: Yes, this would be really nice :D

bprusinowski:
Thanks for the review @ptbrowne 🙇‍♂️ 

Yes, I think the introduction of Storybook would be a nice improvement, maybe we can think about this during Maintenance Quarterly.

When it comes to adding a new step vs adding a new section in the current editor, I think we decided to go with this approach to make the separation of concerns clear, as the goal of the editing step is to always focus on one chart at the time, while the layout step is not about editing the charts anymore, but just about how to present them. I think we could even make it more clear that we are not in a "tab layout" mode when we edit multiple charts, but rather are just in edit single chart at the time mode and change a bit the UI of switching between the charts.

I see at least one challenge of how to communicate the addition of conditional layout section, as it should only appear when we have more than one chart – in the current mode this is enforced by replacing the Publish button with Layout button (basically you can't skip or miss this step when having multiple charts).

As the editing section is already quite populated, I think it makes sense to introduce another step (I think losing the tab selection is acceptable and implemented purposely) to further indicate we are not in a single-chart context anymore. But this could be changed to not clear `activeChartKey` when making the transition 👍 

@KerstinFaye I'll also let you jump in here, maybe I missed something 😅

PR #1299, opened 2023-12-12
perf: Optimize dataset preview performance

bprusinowski:


app/rdf/query-cube-preview.ts
```diff
@@ -0,0 +1,278 @@
...
+  ?dimensionDataKind rdf:type ?dimensionDataKindType .
+  ?dimensionDataKind time:unitType ?dimensionTimeUnitType .
+  ?dimension schema:name ?dimensionLabel .
+  ?dimension schema:description ?dimensionDescription .
+
+  ?observation ?observationPredicate ?observationValue .
+  ?observation ?observationPredicate ?observationValueLabel .
+} WHERE {
+  VALUES ?cube { <${iri}> }
+  FILTER(EXISTS { ?cube a cube:Cube . }) {}
```
Rdataflow, line 80: nit: in the current case you do not query for cube metadata within this query - thus you can simply omit the part `{} UNION` as it only adds empty set to the `WHERE` clause.
and I guess `FILTER(EXISTS { ?cube a cube:Cube . })` isn't needed neither ....

Rdataflow (changes requested):
- [ ] for some unclear reason the metadata GQL for the preview page of BLW cubes
  - is broken with this PR https://visualization-tool-git-perf-optimize-dataset-previe-69ea06-ixt1.vercel.app/browse?previous=%7B%7D&dataset=https%3A%2F%2Fagriculture.ld.admin.ch%2Ffoag%2Fcube%2FMilkDairyProducts%2FConsumption_Price_Month&dataSource=Test
  - while is used to work in main branch https://test.visualize.admin.ch/browse?previous=%7B%7D&dataset=https%3A%2F%2Fagriculture.ld.admin.ch%2Ffoag%2Fcube%2FMilkDairyProducts%2FConsumption_Price_Month&dataSource=Test

- [ ] sorting order is broken with this PR i.e. ordering by danger level for waldbrand
  - `schema:position` isn't respected https://visualization-tool-git-perf-optimize-dataset-previe-69ea06-ixt1.vercel.app/browse?previous=%7B%22order%22%3A%22SCORE%22%2C%22search%22%3A%22waldbrandg%22%7D&dataset=https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fgefahren-waldbrand-warnung%2F1&dataSource=Test
  - which used to be https://test.visualize.admin.ch/browse?previous=%7B%22order%22%3A%22SCORE%22%2C%22search%22%3A%22waldbra%22%7D&dataset=https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fgefahren-waldbrand-warnung%2F1&dataSource=Test
  - nb: just guessing: ordinal time in the BLW time might suffer from the same issue related to `schema:position` ...

app/rdf/query-cube-preview.ts
```diff
@@ -0,0 +1,278 @@
...
+      locale,
+    })}
+    ${buildLocalizedSubQuery(
+      "dimension",
+      "schema:description",
+      "dimensionDescription",
+      { locale }
+    )}
+    FILTER(?dimensionIri != cube:observedBy && ?dimensionIri != rdf:type)
+  } UNION {
```
Rdataflow, line 118: this part of the query leads to many duplicate triples in the resultset. avoiding duplicates reduces the burden of post filtering (i.e. L253)

app/rdf/query-cube-preview.ts
```diff
@@ -0,0 +1,278 @@
...
+            ...baseDimension,
+            __typename: dimensionType,
+          };
+          dimensions.push(dimension);
+        }
+      }
+    }
+  });
+
+  const qsObs = qs.filter(({ p }) => p.equals(ns.cube.observation));
```
Rdataflow, line 250: observations can also be filtered by `o.equals(ns.cube.Observation)` - which will relax the SPARQL query.
- you don't need to reconstruct the `?observationSet cube:observation ?observation` 
- thus can avoid to query `?observationSet` by stepping over ist using `cube:observationSet/cube:observation`
- which avoids duplicates and lowers the post query filtering

app/rdf/query-cube-preview.ts
```diff
@@ -0,0 +1,278 @@
...
+  ?dimensionUnit qudt:currencyExponent ?dimensionUnitCurrencyExponent .
+  ?dimension sh:order ?dimensionOrder .
+  ?dimension meta:dataKind ?dimensionDataKind .
+  ?dimensionDataKind rdf:type ?dimensionDataKindType .
+  ?dimensionDataKind time:unitType ?dimensionTimeUnitType .
+  ?dimension schema:name ?dimensionLabel .
+  ?dimension schema:description ?dimensionDescription .
+
+  ?observation ?observationPredicate ?observationValue .
+  ?observation ?observationPredicate ?observationValueLabel .
```
Rdataflow, line 77: this leads to IRI and Label both present in the SPARQL results. OTOH in the GQL results only Labels are present. would it be an option to drop unnecessary results here in this query, thus no need to treat it downstream?

app/rdf/query-cube-preview.ts
```diff
@@ -118,130 +126,108 @@ export const getCubePreview = async (
...
+    const qDesc = qsDim.find((q) => q.p.equals(ns.schema.description));
+    const qOrder = qsDim.find((q) => q.p.equals(ns.sh.order));
+    const qsType = qsDim.filter((q) => q.p.equals(ns.rdf.type));
+    const qScaleType = qsDim.find((q) => q.p.equals(ns.qudt.scaleType));
+    const scaleType = getScaleType(qScaleType?.o);
+    const qDataType = qsDim.find((q) => q.p.equals(ns.sh.datatype));
+    const qUnit = qsDim.find((q) => q.p.equals(ns.qudt.unit));
+    const qUnitLabel = qs.find((q) => q.s.equals(qUnit?.o));
+    const qDataKind = qsDim.find((q) => q.p.equals(ns.cube("meta/dataKind")));
+    const qDataKindType = qs.find(
```
ptbrowne, line 168: Can't we use `const qlabel = qsDim.out(ns.schema.name)` ? Here, we iterate over the predicates of the dimension for each dimension for each dimesion predicate.

app/rdf/query-cube-preview.ts
```diff
@@ -15,6 +16,12 @@ import * as ns from "./namespace";
 import { getDataKind, getScaleType, timeFormats, timeUnits } from "./parse";
 import { buildLocalizedSubQuery } from "./query-utils";
 
+type ShortQuad = Omit<Quad, "subject" | "predicate" | "object"> & {
+  s: Quad["subject"];
+  p: Quad["predicate"];
+  o: Quad["object"];
+};
```
ptbrowne, line 33: I am bit surprised that it's necessary to have this while we can destructure and rename at the levels of functions, I wonder if by using in/out on the quads, we would not manage to write shorter code with no need for ShortQuad.

app/rdf/query-cube-preview.ts
bprusinowski: We can't use `.out` here, as it's a method of a [`Dataset`](https://github.com/zazuko/clownface/blob/1f1a8a6991f9a506942eeec1a4afc113a43cc2e4/lib/Context.js#L30-L40), not `Quad` that we found above. As in this query we fetch several not related things (e.g. data related to cube, dimensions, observations), we would need to create several instances of `Dataset` class to be able to use `.out` – I tried to use this approach at first, but I had problems to make it work correctly (and it [seems to be doing the same thing](https://github.com/zazuko/clownface/blob/1f1a8a6991f9a506942eeec1a4afc113a43cc2e4/lib/Context.js#L121-L148), just wrapped). I could try to rewrite the logic, but I'm not 100% sure it would be more performant 🤔

app/rdf/query-cube-preview.ts
bprusinowski: That was done for convenience and to force the same naming across several places we used the `Quad`s – it could be done in each individual map, but to use `.out` we would need to completely change the structure of the logic 🥲

app/rdf/query-cube-preview.ts
bprusinowski: For some reason omitting this part leads to worse performance (around 0.5s-0.7s vs 0.3s-0.5s when it's there) – as I am not sure why it's like this, and both queries lead to the same results, I think it could be better to keep it 👍

app/rdf/query-cube-preview.ts
ptbrowne: OK, I did not know that in/out were iterating over all the current context predicates, thanks for pointing this out.

app/rdf/query-cube-preview.ts
```diff
@@ -0,0 +1,278 @@
...
+
+  ?observation ?observationPredicate ?observationValue .
+  ?observation ?observationPredicate ?observationValueLabel .
+} WHERE {
+  VALUES ?cube { <${iri}> }
+  FILTER(EXISTS { ?cube a cube:Cube . }) {}
+  UNION {
+    ?cube cube:observationConstraint/sh:property ?dimension .
+    ?dimension sh:path ?dimensionIri .
+    OPTIONAL { ?dimension sh:datatype ?dimensionDataType . }
```
Rdataflow, line 84: in case datatype is effectively needed, it can be in two places - the other one being
`    OPTIONAL { ?dimension sh:or/rdf:rest*/rdf:first/sh:datatype ?dimensionDataType . }`

app/rdf/query-cube-preview.ts
bprusinowski: Thanks for pointing this our @Rdataflow, I think we still need to have `uniqBy` to get unique observations, as they would be duplicated by the number of dimensions they have, but we can drop `?observationSet cube:observation ?observation` 👍

app/rdf/query-cube-preview.ts
bprusinowski: Good catch, fixed 👍

app/rdf/query-cube-preview.ts
Rdataflow: @bprusinowski in order to get unique results the SPARQL query needs to be reworked first... 😄

bprusinowski:
@Rdataflow dataset preview for BLW cubes is broken due to errors in the data, see that the landing page is localized while it shouldn't be: https://s.zazuko.com/2ionzw1. It works in TEST, because there we queries for cube metadata using the old way, while now we switched to a dedicated query (`DataCubeMetadata`).

This problem will be addressed in the near future by Zazuko 👍

app/rdf/query-cube-preview.ts
bprusinowski: @Rdataflow I tried to include the additional line to SPARQL query and it seems to not affect the performance in noticeable way 👍 I think we might stay with it for the moment if that's fine with you :)

bprusinowski: In fact, when trying with a different, bigger cube, the impact feel more significant. I'll take a look tomorrow to retrieve data type directly from observations 👀

app/rdf/query-cube-preview.ts
```diff
@@ -146,6 +151,66 @@ CONSTRUCT {
...
+      const dimensionValue: DimensionValue = {
+        value: qDimValue.value,
+        label: `${observation[qDimIri.value]}`,
+        position: position !== undefined ? +position : undefined,
+      };
+      dimValuesByDimIri[dimIri].push(dimensionValue);
+    });
+
+    observations.push(observation);
+  });
```
ptbrowne, line 207: Can you index the quads by object, then subject ? Should be then more efficient than find() ?

app/rdf/query-cube-preview.ts
```diff
@@ -82,6 +82,7 @@ CONSTRUCT {
     ?cube cube:observationConstraint/sh:property ?dimension .
     ?dimension sh:path ?dimensionIri .
     OPTIONAL { ?dimension sh:datatype ?dimensionDataType . }
+    OPTIONAL { ?dimension sh:or/rdf:rest*/rdf:first/sh:datatype ?dimensionDataType . }
```
ptbrowne: What does it mean here ?

bprusinowski: Datatype could also be stored in such way, we used to retrieve it in a similar way [here](https://github.com/visualize-admin/visualization-tool/blob/2b77e5226adbdca7c43d592d57614a4791e696e7/app/rdf/parse.ts#L145-L149) – but this was cancelled, as we can avoid querying additional parameters and get data type directly from dimension values we fetch anyway 👍

app/rdf/query-cube-preview.ts
bprusinowski: Done :)

app/rdf/query-cube-preview.ts
```diff
@@ -211,76 +202,88 @@ CONSTRUCT {
...
+        q.subject.equals(qUnit?.object) &&
+        q.predicate.equals(ns.qudt.currencyExponent)
+    );
+    const isKeyDimension = qsType.some((q) =>
+      q.object.equals(ns.cube.KeyDimension)
     );
-    const isKeyDimension = qsType.some((q) => q.o.equals(ns.cube.KeyDimension));
     const isMeasureDimension = qsType.some((q) =>
-      q.o.equals(ns.cube.MeasureDimension)
+      q.object.equals(ns.cube.MeasureDimension)
```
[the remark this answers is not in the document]
bprusinowski: I reduced the number of finds in newest commit, thanks for pointing this out 😄

ptbrowne (approved):
Thanks !

Rdataflow:
@bprusinowski well done 🎉

PR #1337, opened 2024-02-14
chore: Automatic GQL performance regression tests

bprusinowski:


bprusinowski:
@ptbrowne I am going to merge to be able to test the GitHub Action and iterate in the next PR. Feel free to review :)

.github/workflows/performance-tests-pr.yml
```diff
@@ -0,0 +1,33 @@
...
+            k6 run --tag testid=PossibleFilters --env ENV=PR --env ENDPOINT=${{ github.event.deployment_status.target_url }}/api/graphql --env CUBE_IRI=https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/9 --env CUBE_LABEL=Photovoltaikanlagen/9 - </root/k6/performance-tests/graphql/PossibleFilters.js &&
+            k6 run --tag testid=PossibleFilters --env ENV=PR --env ENDPOINT=${{ github.event.deployment_status.target_url }}/api/graphql --env CUBE_IRI=https://environment.ld.admin.ch/foen/nfi/nfi_C-20/cube/2023-3 --env CUBE_LABEL=NFI/2023-3 - </root/k6/performance-tests/graphql/PossibleFilters.js &&
+            k6 run --tag testid=PossibleFilters --env ENV=PR --env ENDPOINT=${{ github.event.deployment_status.target_url }}/api/graphql --env CUBE_IRI=https://energy.ld.admin.ch/elcom/electricityprice --env CUBE_LABEL=Elcom - </root/k6/performance-tests/graphql/PossibleFilters.js
```
ptbrowne, line 33: Can you extract this part to a script ? This way we have less reliance on github actions & we can run the script locally.

ptbrowne (approved):
I am wondering if we could not have the k6 commands as part of our usual tests done via jest (maybe using k6 programmatically) ? This way we stay within our test framework, and maybe we would not have the intermediate step of generating the yml files ?

ptbrowne:
Sorry I was unclear. I was meaning that we would put the test in jests but outside of the test suite required to pass, I think using the "projects" feature from jest would enable to have the main "project" being used for CI and another "performance-e2e" project that could be run only if we want it.

I see the following advantages of using jest:

- Integration with vscode to run performance test from the editor or the CLI
- Ability to only focus on one performance test
- Abilty to have reports (test / fail) per performance test

.github/workflows/performance-tests-pr.yml
bprusinowski: I think still we would have problems to run it locally, as it references `github.event.deployment_status.target_url`, right? Let me know if I missed something again 😅

PR #1340, opened 2024-02-15
Save draft

ptbrowne:
- [x] Create new chart and save as draft
- [x] List drafts as a separate list inside the profile
- [x] Ability to publish a draft from profile
- [x] Ability to turn a published chart into a draft from profile
- [x] Show the draft button inside the "configure" view
- [x] Show the draft button inside the "layout" view
- [x] When viewing a draft, show a warning 
- [X] When viewing a draft, ability for the owner to go to edit page

### DIFF
--- a/app/configurator/components/layout.tsx
+++ b/app/configurator/components/layout.tsx
@@ -47,7 +47,7 @@ const useStyles = makeStyles<Theme>((theme) => ({
   LMRPanelHeaderLayout: {
     width: "100%",
     display: "grid",
-    gridTemplateColumns: `${DRAWER_WIDTH}px minmax(22rem, 1fr) ${DRAWER_WIDTH}px`,
+    gridTemplateColumns: `${DRAWER_WIDTH}px minmax(22rem, 1fr) auto`,
     gridTemplateAreas: `
     "left middle right"
     `,
--- a/app/configurator/configurator-state.spec.tsx
+++ b/app/configurator/configurator-state.spec.tsx
@@ -10,13 +10,13 @@ import {
   ConfiguratorStateConfiguringChart,
   DataSource,
   Filters,
-  MapConfig,
   getChartConfig,
+  MapConfig,
 } from "@/config-types";
 import {
-  ConfiguratorStateAction,
   applyNonTableDimensionToFilters,
   applyTableDimensionToFilters,
+  ConfiguratorStateAction,
   deriveFiltersFromFields,
   getFiltersByMappingStatus,
   getLocalStorageKey,
@@ -153,7 +153,11 @@ describe("initChartFromLocalStorage", () => {
   it("should initialize from localStorage if valid", async () => {
     localStorage.setItem(
       getLocalStorageKey("viz1234"),
-      JSON.stringify({ state: "CONFIGURING_CHART", ...fakeVizFixture })
+      JSON.stringify({
+        state: "CONFIGURING_CHART",
+        published_state: "PUBLISHED",
+        ...fakeVizFixture,
+      })
     );
     const state = await initChartStateFromLocalStorage("viz1234");
     expect(state).not.toBeUndefined();
--- a/app/login/components/profile-tables.tsx
+++ b/app/login/components/profile-tables.tsx
@@ -1,26 +1,38 @@
 import { t, Trans } from "@lingui/macro";
+import { LoadingButton, TabContext, TabList, TabPanel } from "@mui/lab";
 import {
   Box,
   Button,
+  Dialog,
+  DialogActions,
+  DialogContent,
+  DialogProps,
+  DialogTitle,
+  Divider,
   Link,
   Skeleton,
+  Tab,
   Table,
   TableBody,
   TableCell,
   TableHead,
   TableRow,
+  TextField,
   Typography,
+  useEventCallback,
 } from "@mui/material";
 import { PUBLISHED_STATE } from "@prisma/client";
 import sortBy from "lodash/sortBy";
 import NextLink from "next/link";
-import React from "react";
+import React, { FormEvent, useState } from "react";
 
+import useDisclosure from "@/components/use-disclosure";
 import { ParsedConfig } from "@/db/config";
 import { sourceToLabel } from "@/domain/datasource";
 import { truthy } from "@/domain/types";
 import { useUserConfigs } from "@/domain/user-configs";
 import { useDataCubesMetadataQuery } from "@/graphql/hooks";
+import { Locale } from "@/locales/locales";
 import { useRootStyles } from "@/login/utils";
 import { useLocale } from "@/src";
 import { removeConfig, updateConfig } from "@/utils/chart-config/api";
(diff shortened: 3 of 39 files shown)


app/utils/chart-config/api.ts
```diff
@@ -33,31 +37,35 @@ export const createConfig = async (state: ConfiguratorStatePublishing) => {
...
+            version: state.version,
+            dataSource: state.dataSource,
+            layout: state.layout,
+            chartConfigs: state.chartConfigs,
+            activeChartKey: state.activeChartKey,
+          },
+          published_state,
         },
-      },
+        isUndefined
```
bprusinowski, line 63: I guess this could be avoided if published_state was a required property?

app/utils/chart-config/api.ts
```diff
@@ -56,11 +56,7 @@ export const updateConfig = async (
           userId,
           data: {
             key,
-            version: state.version,
-            dataSource: state.dataSource,
-            layout: state.layout,
-            chartConfigs: state.chartConfigs,
-            activeChartKey: state.activeChartKey,
+            ...state,
```
bprusinowski, line 59: Not 100% sure, but I think it was destructured like this because there are more properties in the state that we want to keep in database? It looks like we don't want to have "state" property in database 👀

bprusinowski (commented):
Couldn't test yet due to failed deployment but code looks 💯 👏 Just a few remarks / questions :)

app/utils/chart-config/api.ts
ptbrowne: I think this will stay even if I make published_state required as we do not want each time we update a user config, to have to pass published_state.

app/utils/use-fetch-data.ts
```diff
@@ -1,41 +1,125 @@
...
+
+type QueryCacheValue<T> = {
+  data: T | null;
+  error: Error | null;
+  status: Status;
+};
+
+class QueryCache {
+  cache: Map<string, QueryCacheValue<unknown>>;
+  listeners: [string, () => void][];
```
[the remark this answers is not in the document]
ptbrowne: Same idea but here the QueryCache only notifies subscribers for the key, while Observable fire for every listeners. Since there is little code, I am not sure it is needed to share the observable behavior here.

app/utils/use-fetch-data.ts
```diff
@@ -1,41 +1,125 @@
...
+  const invalidate = useCallback(() => {
     fetchData();
-  }, [queryFn, enable]);
+  }, [fetchData]);
+
+  return { data: data ?? defaultData, error, status, invalidate };
+};
 
-  return { data, error, status };
+export const useHydrate = <T>(queryKey: QueryKey, data: T) => {
```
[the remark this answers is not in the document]
ptbrowne: I think I will leave it like that, I think that importing it from use-fetch-data would be enough. I like typing `useHyd` and having the autocompletion. I added a docstring though.

image

app/utils/chart-config/api.ts
```diff
@@ -13,69 +14,83 @@ import type apiConfigCreate from "../../pages/api/config-create";
...
-};
+export const updateConfig = async (options: UpdateConfigOptions) => {
+  const { key, user_id, published_state } = options;
 
-export const updateConfig = async (
-  state: ConfiguratorState,
-  options: UpdateConfigOptions
-) => {
-  const { key, userId, published_state } = options;
+  console.log("hello", key, user_id, published_state);
```
bprusinowski: 👋

app/pages/v/[chartId].tsx
```diff
@@ -168,10 +184,38 @@ const VisualizationPage = (props: Serialized<PageProps>) => {
...
+                      alignItems: "center",
+                      flexGrow: 1,
+                      flexDirection: "row",
+                      justifyContent: "space-between",
+                    },
+                  }}
+                >
+                  <div>
+                    <Trans id="hint.publication.draft">
+                      This chart is still in draft.
```
Comment (bprusinowski) [inline app/pages/v/[chartId].tsx:204]: This chart is still _a_ draft?

app/configurator/components/layout.tsx
```diff
@@ -47,7 +47,7 @@ const useStyles = makeStyles<Theme>((theme) => ({
   LMRPanelHeaderLayout: {
     width: "100%",
     display: "grid",
-    gridTemplateColumns: `${DRAWER_WIDTH}px minmax(22rem, 1fr) ${DRAWER_WIDTH}px`,
+    gridTemplateColumns: `${DRAWER_WIDTH}px minmax(22rem, 1fr) auto`,
```
bprusinowski, line 50: I think the idea here was to have a symmetrical layout, not sure if we need to have this change – if you think so, then you should also update `LMRPanelHeaderLayout` class to mimic this 😄 Btw, as it's not used anywhere, maybe we can just remove this layout option :)

app/configurator/configurator-state.spec.tsx
```diff
@@ -153,7 +153,11 @@ describe("initChartFromLocalStorage", () => {
   it("should initialize from localStorage if valid", async () => {
     localStorage.setItem(
       getLocalStorageKey("viz1234"),
-      JSON.stringify({ state: "CONFIGURING_CHART", ...fakeVizFixture })
+      JSON.stringify({
+        state: "CONFIGURING_CHART",
+        published_state: "PUBLISHED",
+        ...fakeVizFixture,
+      })
```
bprusinowski, line 160: Maybe it also would make sense to apply migration here, to forget about updating this when config schema changes?

app/login/components/profile-tables.tsx
```diff
@@ -289,12 +289,12 @@ const ProfileVisualizationsRow = (props: ProfileVisualizationsRowProps) => {
...
       }}
     >
-      <TableCell width={80}>
+      <TableCell width="10%">
         <Typography variant="body2">
           {config.data.chartConfigs.length > 1 ? "multi" : "single"}
         </Typography>
       </TableCell>
-      <TableCell width="auto" sx={{ maxWidth: 320 }}>
+      <TableCell width="35%">
```
bprusinowski, line 472: Not sure if important, but I think the sum of cell widths now equals 110%? :D

bprusinowski (approved):
LGTM 👌

app/utils/chart-config/api.ts
ptbrowne: Removed after ;)

app/pages/v/[chartId].tsx
Comment (ptbrowne) [reply app/pages/v/[chartId].tsx:204]: Fixed.

app/login/components/profile-tables.tsx
ptbrowne: 🧠

app/configurator/configurator-state.spec.tsx
ptbrowne: Will do in another PR.

app/configurator/components/layout.tsx
ptbrowne: The problem here was that the publish options + save draft would be wider.

PR #1349, opened 2024-02-19
fix: PR GQL GitHub Action

bprusinowski:
This PR fixes GQL PR performance tests by correctly passing data between Docker image and GitHub Action. It also adds some documentation on developing GitHub Actions locally, which proved to be of great help when fixing the GQL PR action.


.github/workflows/performance-tests-pr.yml
```diff
@@ -15,14 +12,18 @@ jobs:
...
...E_LABEL=Elcom --env ROOT_PATH=/root/ --env CHECK_TIMING=true --quiet - </root/k6/performance-tests/graphql/DataCubePreview.js && k6 run --tag testid=PossibleFilters --env ENV=PR --env ENDPOINT=${{ github.event.deployment_status.target_url }}/api/graphql --env CUBE_IRI=https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/9 --env CUBE_LABEL=Photovoltaikanlagen/9 --env ROOT_PATH=/root/ --env CHECK_TIMING=true --quiet - </root/k6/performance-tests/graphql/PossibleFilters.js && k6 run --tag testid=PossibleFilters --env ENV=PR --env ENDPOINT=${{ github.event.deployment_status.target_url }}/api/graphql --env CUBE_IRI=https://environment.ld.admin.ch/foen/nfi/nfi_C-20/cube/2023-3 --env CUBE_LABEL=NFI/2023-3 --env ROOT_PATH=/root/ --env CHECK_TIMING=true --quiet - </root/k6/performance-tests/graphql/PossibleFilters.js && k6 run --tag testid=PossibleFilters --env ENV=PR --env ENDPOINT=${{ github.event.deployment_status.target_url }}/api/graphql --env CUBE_IRI=https://energy.ld.admin.ch/elcom/electricityprice --env CUBE_LABEL=Elcom --env ROOT_PATH=/root/ --env CHECK_TIMING=true --quiet - </root/k6/performance-tests/graphql/PossibleFilters.js)" > /root/summary.txt
```
ptbrowne, line 23: I think it would be more readable as a script. In general our Github actions shoud be simple scripts so that we can debug them easily.

ptbrowne: Ah OK this is generated. Point still applies but less important. I think a script could be generated so that it does not live in github actions that are difficult to debug locally.

ptbrowne: Can you add a "Generated" mention at the top of the file (mentioning that this file should not be editted directly, but regenerated through the proper command) ?

k6/performance-tests/graphql/PossibleFilters.js
```diff
@@ -80,7 +80,19 @@ export default function Observations() {
...
+    data.metrics.http_req_duration.values.avg >
+    2 * metadata.queries.PossibleFilters.expectedDuration
+  ) {
+    return {
+      stdout: `${Math.round(
+        (100 * data.metrics.http_req_duration.values.avg) /
+          metadata.queries.PossibleFilters.expectedDuration
+      )}% – PossibleFilters – ${cubeLabel}\n`,
+    };
+  }
```
ptbrowne, line 93: Seems like the check and log could be extracted since the logic is repeated.

ptbrowne (approved):

ptbrowne (approved):
Cool, thanks for adding the `act` documentation. I still think it would be easier to have simple scripts, but also good like that.

.github/workflows/performance-tests-pr.yml
bprusinowski: Good idea, I'll add the note 👌 I think with `act` it's "easy enough" to debug, we could revisit if it becomes a problem to have the script embedded in GitHub Action 👍

PR #1352, opened 2024-02-20
Save draft

ptbrowne:
- Reduce flickering the 1st time we save a draft (not changing the page)
- Correctly saving published state the 1st time when we first create the config : I had not seen this because I had first put published state's default as "DRAFT" in Prisma schema locally,  then had changed it to default=PUBLISHED without running the migration. Then on test, the default was PUBLISHED and I could see that at creation time, published draft was not taken into account. After running prisma migrate locally, I could reproduce the bug and fix it.

### DIFF
--- a/app/components/chart-selection-tabs.tsx
+++ b/app/components/chart-selection-tabs.tsx
@@ -29,10 +29,12 @@ import {
   enableLayouting,
   getChartConfig,
   hasChartConfigs,
+  initChartStateFromChartEdit,
   isConfiguring,
   isLayouting,
   isPublished,
   isPublishing,
+  saveChartLocally,
   useConfiguratorState,
 } from "@/configurator";
 import { ChartTypeSelector } from "@/configurator/components/chart-type-selector";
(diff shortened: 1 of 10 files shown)


bprusinowski (approved):
Code looks good! 👌 

Not related to this PR, but while playing around with the feature on TEST I noticed that I can't edit the charts I created some time ago in the past (see video). It looks like there's a need to apply migrations to chart config for every action we do (View, Edit, ...) 👀 

https://github.com/visualize-admin/visualization-tool/assets/52032047/2bd0dcca-8aa2-419b-bf02-630731adaabb

ptbrowne:
Thanks Bartosz, I created another issue for this. Since people do not have charts yet, I also think it is not too much of an issue 😅 .

bprusinowski:
@ptbrowne yup, but if they start making them now and we introduce a new property in the config, it will become an issue in the future 👀 Thanks for creating an issue 🙇‍♂️

Issue #1354, opened 2024-02-20
Can't edit old charts, need for migrations


ptbrowne:
Not related to this PR, but while playing around with the feature on TEST I noticed that I can't edit the charts I created some time ago (see video). It looks like there's a need to apply migrations to chart config for every action we do (View, Edit, ...) 👀 

https://github.com/visualize-admin/visualization-tool/assets/52032047/2bd0dcca-8aa2-419b-bf02-630731adaabb

_Originally posted by @bprusinowski in https://github.com/visualize-admin/visualization-tool/pull/1352#pullrequestreview-1890912677_

tboeni:
@ptbrowne This feature also does not work anymore on **INT** for the _copy and edit this visualization_ button 
https://int.[visualize.admin.ch](https://int.visualize.admin.ch/de/v/M65niBNzwB3H)/de/v/M65niBNzwB3H

PR #1358, opened 2024-02-22
fix: Remove extra published_state property in the configurator state

ptbrowne:
Made a silly mistake where I put published_state in the config and also
in the configurator state. Published_state should only be in the config,
not the state.

This resulted in old charts not being able to edited/copied/viewed due
to missing property while decoding configurator state.

Fix https://github.com/visualize-admin/visualization-tool/issues/1354

### DIFF
--- a/app/config-types.ts
+++ b/app/config-types.ts
@@ -1134,7 +1134,6 @@ const Config = t.type(
     layout: Layout,
     chartConfigs: t.array(ChartConfig),
     activeChartKey: t.string,
-    published_state: t.string,
   },
   "Config"
 );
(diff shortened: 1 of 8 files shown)


bprusinowski (changes requested):
@ptbrowne I think you also need to remove the migration then if the property is no longer in the config?

I also wonder why this bug appeared when migration was correctly added? Shouldn't the charts be migrated correctly, maybe we miss a `migrateConfiguratorState` call somewhere?

ptbrowne:
Thanks Bartosz, good point, I removed the migration.

> I also wonder why this bug appeared when migration was correctly added? 

Thanks also for wondering this 😅 So I investigated, and I had forgotten to bump the final version in `CONFIGURATOR_STATE_VERSION`. So the last migration had not been applied. With the mixing up of the attribute in configurator state, and config, I thought it was working when it was not. I added a test to make sure we do not forget to update the version.

bprusinowski (approved):
LGTM, thanks 🎉

PR #1360, opened 2024-02-22
perf: Initial DataCubeComponents / dimension values query optimizations

### LINKED ISSUE #1372: Load single value per filter when entering the edit mode first and load unique values afterwards
Currently, we need to wait for the filter panel values to load before a chart is shown. This happens due to fetching unique values per each filter – we could speed up showing of the chart by only loading one value per filter initially and then loading all unique values afterwards, so that the loading would be scoped purely to the filter panel, and not the whole application.

**User story**

I want the chart to not be waiting for the filter values to be loaded.

**Definition of Done**

When initially entering editing mode, filters are loaded with single values – so the chart doesn’t need to wait for them – and only then begin to fetch all available values.

bprusinowski:
Closes #1372.

This PR is a first step towards optimizing the performance situation related to `DataCubeComponents` query.

- Disabled loading of dimension values where applicable
- Adjusted several places in the code to not send unnecessary filters to the server to make sure we use `sh:in` as much as possible
- Fixed an issue with not using optional chaining to access last dimension value
- Refactored the approach to fetching available filter values from one combined `DataCubesComponents` call with filters for every dimension to individual calls directly from filter components. This enables us to more precisely pass filters and rely more on `sh:in`, as for the first filter we won't send `filters` object anymore
- Fixed an issue with fetching dimension values which resulted in fetching values that couldn't be selected (combinations resulting in no observations, fixed after selection by possible filters query)

### Next steps
- Consolidate `DataCubeComponents`-related queries into `CONSTRUCT` query/ies
- Consider using `mobx` to improve manageability of data fetching across the app
- Consider decoupling dimension values fetching from components to make it more clear where we actually need to fetch values (currently it's not ideal, as we might or might not have values fetched)

### How to test
**PR**
1. Go to [this link](https://visualization-tool-git-perf-optimize-data-cube-comp-527a80-ixt1.vercel.app/en/create/new?cube=https://environment.ld.admin.ch/foen/fab_ahst_exp_imp_pipeline/1&dataSource=Test).
2. ✅ See that the filters are loaded in the background (some might fail to load due to Vercel's 15s timeout).
3. ✅ See that the last filter (`4. Zusatzmengencode - EN`) contains five values.

**TEST**
1. Go to [this link](https://test.visualize.admin.ch/en/create/new?cube=https://environment.ld.admin.ch/foen/fab_ahst_exp_imp_pipeline/1&dataSource=Test).
2. ❌ See that the initial view loads much slower compared to the PR link.
3. ❌ See that the last filter (4. Zusatzmengencode - EN) contains seven values.
4. Try to change value of above filter to `Metres`.
5. ❌ See that the chart reloads back to `No supplementary unit`.


app/configurator/configurator-state.tsx
```diff
@@ -359,7 +359,11 @@ const getCachedComponents = (
       sourceType: draft.dataSource.type,
       sourceUrl: draft.dataSource.url,
       locale,
-      cubeFilter: { iri: cubeFilter.iri, joinBy: cubeFilter.joinBy },
+      cubeFilter: {
```
ptbrowne, line 362: Should this be renamed cubeOptions ?

app/charts/shared/chart-data-filters.tsx
```diff
@@ -242,7 +242,7 @@ const DataFilter = (props: DataFilterProps) => {
         {
           iri: cubeIri,
           componentIris: [dimensionIri],
-          filters: keys.length > 0 ? interactiveFilters : undefined,
+          filters: keys.length > 1 ? interactiveFilters : undefined,
```
ptbrowne, line 245: Why ?

app/rdf/query-dimension-values.ts
```diff
@@ -232,39 +232,59 @@ const getQueryFilters = (
...
-      })
-    : "";
+  if (filtersList.length === 0) {
+    return "";
+  }
+
+  let i = 0;
+  const filterDimensionIris = filtersList.map(([iri]) => iri);
+  // Also include other dimensions to make sure we don't return values that could
+  // result in no observations
```
ptbrowne, line 242: I fail to see the previous buggy case ? Could you explain to me why if we do not add those dimensions, there is a bug 😅  ?

ptbrowne (approved):

app/charts/shared/chart-data-filters.tsx
bprusinowski: Explained offline :)

app/rdf/query-dimension-values.ts
bprusinowski: Sure, take a look at these queries:
- before change (7 results): https://s.zazuko.com/4ntKoP,
- after change (5 results): https://s.zazuko.com/xohXms.

Also check the number 5 in `How to test (TEST)` section from the PR description 👍 

Basically if we don't include other dimensions, we risk that there is an observation that contains values for the "core dimensions", but has some undefined values in other dimensions – which leads to a "reload" of possible filters and the bug in the PR description. Inclusion of other dimensions act as a filter to remove undefined values potentially attached to these other dimensions.

Let me know if I can clarify this further 👀

app/configurator/configurator-state.tsx
bprusinowski: I added a new issue (#1376) I think it would make sense to rename 👍

Rdataflow:
@bprusinowski now that you pass `loadValues` this is the condition when to ask for hierarchies
https://github.com/visualize-admin/visualization-tool/blob/9de220438a23a37b88c12f6a364e6688839c2c3b/app/graphql/resolvers/rdf.ts#L228-L229
only if `loadValues` is true

impact: gain speed by skipping unecessary hierachy queries alltogether

app/rdf/query-dimension-values.ts
```diff
@@ -227,44 +227,65 @@ const getFiltersList = (filters: Filters, dimensionIri: string | undefined) => {
...
+  // Also include other dimensions to make sure we don't return values that could
+  // result in no observations. Inclusion of other dimensions is necessary to
+  // filter out unobserved values.
+  const otherDimensionFilters = cube.dimensions
+    .filter(
+      (dim) =>
+        !filterDimensionIris.includes(dim.path?.value ?? "") &&
+        !dim.path?.equals(ns.cube.observedBy)
+    )
+    .map((dim) => `?observation <${dim.path?.value}> ?dimension${i++} .`);
```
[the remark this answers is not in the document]
Rdataflow: @bprusinowski I confirm this is a data issue in this cube - therefore it's safe to drop it 👍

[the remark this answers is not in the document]
bprusinowski: Discussed offline with @Rdataflow, it's a data issue that causes the above problems, we can drop these filters.

PR #1361, opened 2024-02-22
Fix tests

ptbrowne:
- fix: Color mapping test
- fix: Temporal filter test broken
- fixed: After layout adjustments we now use different test-id
- fix: Order of default filters has changed on this cube

### DIFF
--- a/e2e/selectors.ts
+++ b/e2e/selectors.ts
@@ -46,9 +46,9 @@ export const createSelectors = ({ screen, page, within }: Ctx) => {
       },
     },
     panels: {
-      left: () => screen.getByTestId("panel-left"),
+      left: () => screen.getByTestId("panel-body-L"),
       drawer: () => screen.getByTestId("panel-drawer"),
-      middle: () => screen.getByTestId("panel-middle"),
+      middle: () => screen.getByTestId("panel-body-M"),
       metadata: () => screen.getByTestId("panel-metadata"),
     },
     edition: {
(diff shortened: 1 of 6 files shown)


e2e/selectors.ts
```diff
@@ -46,9 +46,9 @@ export const createSelectors = ({ screen, page, within }: Ctx) => {
       },
     },
     panels: {
-      left: () => screen.getByTestId("panel-left"),
+      left: () => screen.getByTestId("panel-body-L"),
```
bprusinowski, line 49: Wondering if we could maybe extract this to a const and reuse it here and in the layout file, so we **can't** forget about updating it in the future? Maybe it could be adapted for all test-ids across the app, what do you think? Or should the "testId" name automatically ring a bell to update an e2e test? 😅

bprusinowski (approved):
Thanks @ptbrowne, the situation looks much better now :)

Would it also make sense to remove the below part of GitHub Action? https://github.com/visualize-admin/visualization-tool/blob/7c05360b59ad0fefe154d5b7ea726bb76f85b9e1/.github/workflows/e2e.yml#L36-L62

It looks like the state is mirrored between this and built-in check status in PRs, or maybe I'm missing something 💭

e2e/selectors.ts
ptbrowne: I think the broken test should automatically ring a bell 😁 I will try to make a const file 👍

ptbrowne:
> Would it also make sense to remove the below part of GitHub Action? 

Yes it seems like so, I don't remember know why we have this 😅

PR #1378, opened 2024-02-28
perf: Optimize DataCubeComponents queries

bprusinowski:
This PR brings optimizations related to `DataCubeComponents` query, along with app-wide improvements to cube fetching (do not fetch twice in case of a need for latest cube, do not fetch whole shape but use tailored queries).

## Performance optimizations
- Refactored dimension values fetching to use one `CONSTRUCT` query instead of several smaller `SELECT` queries
- Introduced a `LightCube` data fetching utility class that's used to fetch _just enough_ data related to different parts of a cube, instead of relying on recursive `DESCRIBE` queries against cube shape
- Applied server-side caching to more queries

## Fixes
- Dimension values are now unversioned when possible (regression introduced by #1360)
- Theme labels are now sorted by label in dataset preview, as in browse page


Rdataflow:
fyi: observed in this cube https://visualization-tool-git-perf-optimize-data-cube-comp-4a3cd4-ixt1.vercel.app/create/new?cube=https://environment.ld.admin.ch/foen/nfi/nfi_T-changes/cube/2024-1&dataSource=Prod

bprusinowski:
@Rdataflow thanks for noticing, I am now switching a bit from pure dimension values to dimensions & cubes fetching, I will take a look at this 👍

Rdataflow:
> @bprusinowski what you probably already know: the same hierarchy query seems to fire multiple times ATM

sorry, to clarify, this wasn't the case - the queries just look very similar.

Rdataflow:
@bprusinowski this is a really good step towards unifying the many queries. 
of course the most important performance gain may be in
- fetching all dimensions with their values in a `CONSTRUCT`'ed cube>dimension>value tree at once in one query
https://github.com/visualize-admin/visualization-tool/blob/90c7a75bb7ee0fd4c27220c162067626bfdc8d79/app/rdf/queries.ts#L187
- omitting unnecessary queries (i.e. hierarchies or even components alltogehter for published mode - unless there is an interactive filter) I assume this might be another PR then...

nit: a side effect (probably due to draft / WIP) is in [this edit view](https://visualization-tool-git-perf-optimize-data-cube-comp-4a3cd4-ixt1.vercel.app/create/new?cube=https://transport.ld.admin.ch/foca/FOCA_Air_traffic_movements/2&dataSource=Prod) where the second filter becomes ineffective ... thus no measures will show up (neiter after trying to switch the filtered value)

app/rdf/light-cube.ts
```diff
@@ -0,0 +1,266 @@
...
+
+    const query = `PREFIX schema: <http://schema.org/>
+
+SELECT ?iri WHERE {
+  ?versionHistory schema:hasPart  <${this.iri}> .
+  ?versionHistory schema:hasPart ?iri .
+  ?iri schema:version ?version .
+}
+ORDER BY DESC(?version)
+LIMIT 1`;
```
Rdataflow, line 61: IIUC this would upgrade to any latest version - even if the later version is only a draft while the current version is published. This arises the question, how can we ensure to upgrade from published to published // or from draft|published to draft|published. (or maybe you ensure this already by other means)

... it might be useful to incorporate such a query (together with a second clause for cubes w/o versionHistory) within the main query itself i.e. with a nested SELECT pattern

you may try to nest i.e. https://s.zazuko.com/3kWrvgW
resulting in i.e. 

```sparql
PREFIX schema: <http://schema.org/>

SELECT ?cube ?value ?etc WHERE {

# main query 

{ #start nested upgrade query
SELECT ?iri WHERE {
  {
    VALUES ?iri { <https://environment.ld.admin.ch/foen/nfi/nfi_C-22/cube/2023-1> }
    ?iri schema:version ?version .
  }
  UNION
  {
    VALUES ?oldIri { <https://environment.ld.admin.ch/foen/nfi/nfi_C-22/cube/2023-1> }
    ?versionHistory schema:hasPart ?oldIri .
    ?versionHistory schema:hasPart ?iri .
    ?iri schema:version ?version .
    ?iri schema:creativeWorkStatus ?status .
    ?oldIri schema:creativeWorkStatus ?oldStatus .
    FILTER(?status IN ( <https://ld.admin.ch/vocabulary/CreativeWorkStatus/Published>, ?oldStatus))
  }
}
ORDER BY DESC(?version)
LIMIT 1
} #end nested upgrade query

# rest of the main query ...

}
```

app/rdf/queries.ts
```diff
@@ -306,175 +301,14 @@ export const getCubeDimensionValuesWithMetadata = async ({
...
-  }
-
-  if (undValues.length > 0) {
-    result.push({
-      value: DIMENSION_VALUE_UNDEFINED, // We use a known string here because actual null does not work as value in UI inputs.
-      label: "–",
-    });
-  }
-
-  return result;
```
ptbrowne, line 477: Nice to be able to remove all this code ! I think it would be cool to have a loadDimensionValues test on a particular dimension where we can be sure that the CONSTRUCT query brings the same result as before.

Should `loadDimensionValues` be renamed `loadDimensionValuesWithMetadata` since I guess we will have places where we do not need the labels etc.. ?

bprusinowski:
@Rdataflow I think it would be good to have dimension metadata and dimension values fetching separate, as in some places we just need to get dimension metadata (iri, label, type) and we actually need to fetch values in fewer places (e.g. filters).

By having this separation we would have clearer types and could more easily scope the loading of values where really necessary and cache dimension metadata much better than is the case now.

app/rdf/query-dimension-values.ts
```diff
@@ -185,12 +192,17 @@ CONSTRUCT {
...
-  const quads = await sparqlClient.query.construct(query, {
-    operation: "postUrlencoded",
-  });
-  const valuesQuads = quads.filter((q) => q.predicate.equals(ns.rdf.first));
-
-  return valuesQuads.map((qValue) => parseDimensionValue(qValue, quads));
+  return await executeWithCache(
+    sparqlClient,
+    query,
+    () => sparqlClient.query.construct(query, { operation: "postUrlencoded" }),
```
ptbrowne, line 197: Nit: Since execute with cache already receives the sparqlClient, I wonder if it would not be simpler to have a mode:'construct'|'select'|(keyof typeof sparqlClient.query) so that the user here does not have to know about the details ? Isn't it possible to infer from the query that it is a CONSTRUCT or SELECT ?

app/rdf/light-cube.ts
```diff
@@ -0,0 +1,63 @@
...
+import { DataCubeMetadata } from "@/domain/data";
+import { getCubeMetadata } from "@/rdf/query-cube-metadata";
+
+type LightCubeOptions = {
+  iri: string;
+  locale: string;
+  sparqlClient: ParsingClient;
+};
+
+export class LightCube {
```
ptbrowne, line 32: Can you add a docstring with the intent of this class ?

app/graphql/resolvers/rdf.ts
```diff
@@ -277,20 +277,13 @@ export const dataCubeComponents: NonNullable<
...
-    const rawCube = await loaders.cube.load(iri);
-
-    if (!rawCube) {
-      throw new Error("Cube not found!");
-    }
-
-    const cube = latest ? await getLatestCube(rawCube) : rawCube;
+    const cube = await new LightCube({ iri, locale, sparqlClient }).promote(
+      !!latest
+    );
```
ptbrowne, line 284: Nit: I read `promote` as like a "write" on the cube,  instead of a "read", I wonder if this way would not be clearer ?

```
const cube = await new LightCube().fetch({ latest: !!latest })
```

ptbrowne (commented):
I have some comments but it is looking good right now 👍 I think it would be important to clearly define the current responsibilities of LightCube = is is a DataFetching class or in the future will it also hold the state and acts as a cache ?

app/graphql/resolvers/rdf.ts
bprusinowski: Maybe "init" would also work? Here in fact we are not fetching any information for the cube, but rather (potentially) return a new, latest instance of `LightCube`. As we have `fetchMetadata`, and will have other methods soon, I wonder if "fetch" wouldn't feel like a combined fetch of everything related to the cube?

bprusinowski:
> I have some comments but it is looking good right now 👍 I think it would be important to clearly define the current responsibilities of LightCube = is is a DataFetching class or in the future will it also hold the state and acts as a cache ?

That's a good question, I think for the first version it could be just a data fetching class; in a second step we could try to somehow use it as a cache, but I was more thinking about making data fetching more consolidated 👀

app/rdf/light-cube.ts
bprusinowski: Thanks for pointing this out @Rdataflow and proving an example query! I am not sure if we should have a fallback for cubes without a version history, as we only considered ones with it in our current, very old logic.

You can see how we were doing this until now: https://github.com/visualize-admin/visualization-tool/blob/90c7a75bb7ee0fd4c27220c162067626bfdc8d79/app/rdf/queries.ts#L54-L114 (I am going to replicate this logic in a subquery in `LightCube`)

bprusinowski: Should be done via https://github.com/visualize-admin/visualization-tool/pull/1378/commits/12bef541c66a3297f44da2542ab5d345d2f0c4ab 🤞

Rdataflow: @bprusinowski the good news is: the provided subquery already works for cubes w/o versionHistory thanks to the first set just before UNION. (and there are plenty of them, i.e. the electricity-price cubes. they work well on visualize TEST currently and will continue to work with the provided pattern while the others with history upgrade well on TEST and the new pattern as well...)

how to test - case w/o versionHistory
1. open [electricity price](https://test.visualize.admin.ch/browse?dataset=https%3A%2F%2Fenergy.ld.admin.ch%2Felcom%2Felectricityprice&dataSource=Prod)
2. observe SPARQL queries
3. observe visualize works perfectly

how to test - case upgrade with versionHistory
1. open expired cube [nfi cube 2023.1](https://visualize.admin.ch/create/new?cube=https://environment.ld.admin.ch/foen/nfi/nfi_C-90/cube/2023-1) in visualize TEST or PROD
2. observe sparql queries
3. observe visualize seamlessly upgrades to version 2023.3

I hope this helps you to understand
- the need for the ongoing support for cubes without versionHistory.
- a working solution with a simple subquery pattern handling upgrades where applicable - while reducing the amounts of queries.

Just ping me, in case you encounter any questions while implementing...

Rdataflow:
> @Rdataflow I think it would be good to have dimension metadata and dimension values fetching separate, as in some places we just need to get dimension metadata (iri, label, type) and we actually need to fetch values in fewer places (e.g. filters).
> 
> By having this separation we would have clearer types and could more easily scope the loading of values where really necessary and cache dimension metadata much better than is the case now.

@bprusinowski I fully agree to only fetch values or hierarchies only if actually needed - so loading values or hierarchies should be an optional part in the query in order to speed up those cases. 
OTOH IIUC you think to fetch separately sequentially in two separate queries? Wouldn't this lead to more delay than fetching in a single query whereever possible (i.e. in published view with well known dimensions). For this reason it would be speedier to fetch everything in a single query - of course it could also be fetched entirely in yet another query i.e. the CubeObservations query.

app/rdf/light-cube.ts
bprusinowski: @Rdataflow I think the current implementation should work in the same manner, as if the query does not find an iri, it keeps the current one which we already have when initializing the cube (https://github.com/visualize-admin/visualization-tool/blob/b6f56ba91e86f6e75e0e7d2d0ef1e230b410ae90/app/rdf/light-cube.ts#L60-L68) – it means that we don't have to do a round trip and try to fetch cube without version history, because we've already done so in a previous step.

Is there some other use case you see where the current approach won't work?

bprusinowski: See that the cube without version history works in the preview deployment: https://visualization-tool-git-perf-optimize-data-cube-comp-4a3cd4-ixt1.vercel.app/en/create/new?cube=https://energy.ld.admin.ch/elcom/electricityprice-canton&dataSource=Prod

bprusinowski:
@Rdataflow the problem with fetching everything related to dimensions in one query is that we can't control the caching mechanism.

In some places inside the application we only need to have dimension metadata (which takes let's say up to a second to fetch, including data type, units, etc). In some other places, we need to have both this and dimension values on top of the metadata, and in other places we just need to fetch dimension values.

If we only had one query, we would need to re-fetch dimension metadata every time we need to fetch only values – and as dimension metadata is immutable (no need to re-fetch it ever again once it's been fetched once), this would mean we would slow down the queries used to fetch dimension values by this ~1s every time (this will happen often, e.g. during changing a filter).

If two queries are separated, we can still fire them asynchronously at the same time, but ensure that we do not fetch dimension metadata more than needed.

I hope it explains the rationale to separate them better :)

app/rdf/query-dimension-values.ts
```diff
@@ -141,80 +131,147 @@ type LoadDimensionValuesProps = {
...
+      ?blankNode rdf:first ?value .
+    }
+  } UNION`
+  } {
+    {
+      SELECT DISTINCT ?value ?dimensionVersion WHERE {
+        <${cubeIri}> cube:observationConstraint/sh:property ?dimension .
+        ${queryFilters ? "" : `FILTER(NOT EXISTS{ ?dimension sh:in ?in . })`}
+        ?dimension sh:path <${dimensionIri}> .
+        OPTIONAL { ?dimension schema:version ?dimensionVersion . }
```
Rdataflow, line 173: in case `?dimensionVersion` won't be used, it could be dropped?

app/rdf/query-dimension-values.ts
```diff
@@ -141,80 +131,147 @@ type LoadDimensionValuesProps = {
...
+} WHERE { 
+  ${
+    queryFilters
+      ? ""
+      : `{
+    SELECT ?value ?dimensionVersion WHERE {
+      <${cubeIri}> cube:observationConstraint/sh:property ?dimension .
+      ?dimension sh:path <${dimensionIri}> .
+      ?dimension sh:in/rdf:rest* ?blankNode .
+      OPTIONAL { ?dimension schema:version ?dimensionVersion . }
```
Rdataflow, line 163: in case `?dimensionVersion` won't be used, it could be dropped?

app/rdf/query-dimension-values.ts
```diff
@@ -141,80 +131,147 @@ type LoadDimensionValuesProps = {
...
+      ?blankNode rdf:first ?value .
+    }
```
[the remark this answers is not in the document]
bprusinowski: Wow, that's a great improvement! Will add this 👍

app/rdf/query-dimension-values.ts
```diff
@@ -165,7 +165,7 @@ CONSTRUCT {
     }
   } UNION`
   } {
-    {
+    { #pragma evaluate on
```
ptbrowne, line 174: What does that do ?

ptbrowne (approved):

app/rdf/query-dimension-values.ts
bprusinowski: This suggestion comes from @Rdataflow and seems to optimize the execution time – exact description is "suggests to evaluate the scope at query optimization time", see https://docs.stardog.com/query-stardog/query-hints

Rdataflow:
@bprusinowski you may further optimize the sub-SELECT - see inline comments here https://s.zazuko.com/2GXw33V

app/rdf/query-dimension-values.ts
Rdataflow: @ptbrowne see https://docs.stardog.com/query-stardog/query-hints#general-query-optimization-hints

<html>
<body>

evaluate | on,off | scope | Suggests to evaluate the scope at query optimization time.
-- | -- | -- | --



</body>
</html>

bprusinowski:
Good catch @Rdataflow, I'll implement the change 👍

app/rdf/query-dimension-values.ts
bprusinowski: Some of the queries use "raw" SPARQL, without SELECT of CONSTRUCT utilities from Zazuko's libraries. I felt it was easier to write them like this, as otherwise I had some problems with passing `ns` objects inside the CONSTRUCT queries, and also I am not 100% sure if we can control placement of pragmas directly in the middle of a query (as we now do with `#pragma evaluate on`). Let me know if that's a deal-breaker, I can investigate then further in a next PR 🙇‍♂️

Rdataflow:
@bprusinowski upon
1. opening the cube https://test.visualize.admin.ch/create/new?copy=qhx8SgHLTpnx 
2. then reloading (Ctrl+F5)

results in a ~80s SPARQL query  - which may be optimized for 5x speed in a subsequent PR

see query with inline comments https://s.zazuko.com/2pLemUo

Rdataflow:
> Do you see some other way we could achieve the same goal of not returning values that might result in no observations? Maybe adding filters not equal to undefined values?

@bprusinowski I described why we can simply drop `otherDimensions`, see https://github.com/visualize-admin/visualization-tool/pull/1360#discussion_r1513956761 for the details

PR #1386, opened 2024-03-05
fix: SingleURLs publishing

### LINKED ISSUE #1384: After publishing a chart with separate URLs the layout option screen is still shown without any options
**Steps to Reproduce**
1. go to https://test.visualize.admin.ch/en/v/BV8Nql433j1R?dataSource=Int
2. click on "Copy and edit this visualization" 
3. Republish the same visualizations but with the option "separate URLs" (select all visualizations to publish)
4. 3 new tabs open each with a visualization as expected. 
5. The penultimate screen before the visualization is published is still shown but with no options to go back or edit the visualizations

Image

Additional context: 
When the page is reloaded the options are shown and the user can go back to the editor 

Image

Tested in Chrome and Safari.

bprusinowski:
Fixes #1384.

### How to test
1. Go to [this link](https://visualization-tool-git-fix-single-urls-ixt1.vercel.app/en/create/new?cube=https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/9&dataSource=Prod).
2. Add a second chart.
3. Proceed to layout options.
4. Switch to Separate URLs layout.
5. Publish.
6. See that the first chart replaced the current page and second chart was opened in a new tab.

### DIFF
--- a/app/configurator/configurator-state.spec.tsx
+++ b/app/configurator/configurator-state.spec.tsx
@@ -8,6 +8,7 @@ import {
   ColumnConfig,
   ComboLineDualConfig,
   ConfiguratorStateConfiguringChart,
+  ConfiguratorStatePublishing,
   DataSource,
   Filters,
   getChartConfig,
@@ -26,6 +27,7 @@ import {
   initChartStateFromCube,
   initChartStateFromLocalStorage,
   moveFilterField,
+  publishState,
   setRangeFilter,
   updateColorMapping,
 } from "@/configurator/configurator-state";
@@ -55,6 +57,7 @@ jest.mock("rdf-cube-view-query", () => ({
 }));
 
 jest.mock("@/utils/chart-config/api", () => ({
+  createConfig: jest.fn(),
   fetchChartConfig: jest.fn(),
 }));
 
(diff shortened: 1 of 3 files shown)


app/configurator/configurator-state.spec.tsx
```diff
@@ -1201,3 +1204,27 @@ describe("filtering", () => {
...
+      chartConfigs: [
+        { key: "A", cubes: [] },
+        { key: "B", cubes: [] },
+        { key: "C", cubes: [] },
+      ],
+    } as any as ConfiguratorStatePublishing;
+
+    const cb = jest.fn();
+    await publishState({} as any, key, state, async () => cb());
+    expect(cb).toHaveBeenCalledTimes(publishableChartKeys.length);
```
ptbrowne, line 1228: Thanks for the test. Suggestion: it seems to me that the fix you introduced is not tested here ? Maybe we can add checks on the calls to make sure they use the proper i.

`expect(cb.calls.map(c => c.parameter[1]).toEqual([1, 2, 3])` ? (Not sure of the exact API of jest spies calls).

ptbrowne (approved):
Nice catch and thanks for test 💯 I just have a minor suggestion, LGTM 👍

app/configurator/configurator-state.spec.tsx
bprusinowski: Oh man 🤦‍♂️ Yes, you're right, I adjusted the test, thanks 😅

PR #1388, opened 2024-03-05
Add dataset section

ptbrowne:
- [x] Behind the flag "add-dataset" (add flag__add-dataset=true, with ? if not already there, to URL to activate flag)
  - [x] Add section in control showing current cubes
  - [x] Add dialog to search and add cubes based on shared dimension
- [x] Ability to add/remove dataset inside chart config
- [x] When removing dataset, the chart config is re-initialized since the current chart could use a joinBy dimension

## Behind the scenes

- [x] The ChartLoadingWrapper now is responsible for loading the data and values
  - [ ] Extra check the values loaded across chart to see if everything working correctly
- [x] useQuery hooks created with makeUseQuery will no longer return the previous loaded data if the query key changes

### DIFF
--- a/app/components/chart-selection-tabs.tsx
+++ b/app/components/chart-selection-tabs.tsx
@@ -504,6 +504,21 @@ type TabsInnerProps = {
   onChartSwitch?: (key: string) => void;
 };
 
+/**
+ * The Tabs component clones its children adding extra properties, expecting that we use Tab components.
+ * Since we either use a Draggable or a div, we must not forward directly the properties.
+ * Forwarding the props directly to a DOM node would result in for example "textColor" not being a valid HTML prop.
+ */
+const PassthroughTab = ({
+  children,
+  value: _value,
+}: {
+  children: React.ReactNode;
+  value?: string;
+}) => {
+  return <>{children}</>;
+};
+
 const TabsInner = (props: TabsInnerProps) => {
   const {
     data,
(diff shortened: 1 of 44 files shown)


app/configurator/configurator-state.tsx
```diff
@@ -1410,6 +1416,68 @@ const reducer: Reducer<ConfiguratorState, ConfiguratorStateAction> = (
...
+            ]) {
+              if (
+                f[fieldName] === action.value.joinBy.left ||
+                f[fieldName] === action.value.joinBy.right
+              ) {
+                f[fieldName] = "joinBy";
+              }
+              if (f.color) {
+                fields.push(f.color);
+              }
```
bprusinowski, line 1403: I am not sure I understand, why only color and not other fields?

app/configurator/configurator-state.tsx
```diff
@@ -1410,6 +1416,68 @@ const reducer: Reducer<ConfiguratorState, ConfiguratorStateAction> = (
...
+
+          // Need to go over fields, and replace any IRI part of the joinBy by "joinBy"
+          const fields = Object.values(chartConfig.fields);
+          while (fields.length > 0) {
+            const f = fields.pop();
+            for (const fieldName of [
+              "componentIri",
+              "leftComponentIri",
+              "rightComponentIri",
+            ]) {
```
bprusinowski, line 1394: Can we have these fields? Were they added in some prior commit? I searched in codebase and I couldn't find these strings 👀

app/configurator/components/add-dataset-dialog.tsx
```diff
@@ -67,26 +84,27 @@ export const DatasetDialog = ({
     pause: !cubesComponentQuery.data || relevantDimensions.length === 0,
   });
   const handleSubmit = (ev: FormEvent<HTMLFormElement>) => {
+    console.log("handle submit");
```
bprusinowski: 🔍

app/graphql/hooks.spec.tsx
```diff
@@ -0,0 +1,199 @@
+import { act, render, waitFor } from "@testing-library/react";
+import { renderHook } from "@testing-library/react-hooks";
+
+import { useDataCubesComponentsQueryVariables } from "@/graphql/hooks.mock";
+import { Response } from "@/test/utils";
+
+import { makeUseQuery, useDataCubesComponentsQuery } from "./hooks"; // Assuming your makeUseQuery function is in a separate file
```
bprusinowski, line 7: Nit: comment not needed?

app/graphql/hooks.ts
```diff
@@ -42,17 +42,23 @@ const makeUseQuery =
...
-          setResult((prev) => ({ ...prev, fetching: true }))
+          setResult((prev) => ({
+            ...prev,
+            fetching: true,
+            data: prev.queryKey === queryKey ? prev.data : null,
+            queryKey,
+          }))
         );
-        setResult(result);
+        setResult({ ...result, queryKey });
```
bprusinowski, line 61: 💯

app/graphql/hooks.ts
```diff
@@ -67,7 +73,9 @@ const makeUseQuery =
       // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [queryKey, options.pause]);
 
-    return [result, executeQuery] as const;
+    const { data, ...rest } = result;
+    const res = { ...rest, data: result.queryKey === queryKey ? data : null };
+    return [res, executeQuery] as const;
```
bprusinowski, line 78: I am wondering if we need to set the data no null and couldn't just make the query correctly re-fetch the data in the background?

Currently it looks that once we add a dataset, whole application goes into the loading state, unmounting filters, chart, etc, while it could persist the last state and indicate loading? I think it would be less blinking on a page :)

Not related to this PR, but we also have this issue currently with changing the filters – whole chart gets unmounted and a loading indicator takes its place, when it could e.g. just change chart opacity and then update it without mount / unmount thing.

Let me know what you think :)

app/components/chart-selection-tabs.tsx
```diff
@@ -504,6 +504,21 @@ type TabsInnerProps = {
...
+const PassthroughTab = ({
+  children,
+  value: _value,
+}: {
+  children: React.ReactNode;
+  value?: string;
+}) => {
+  return <>{children}</>;
+};
+
```
bprusinowski, line 521: 👍

bprusinowski (approved):
LGTM 👏 

I tested the feature and works very well overall 💯 Just a few points to maybe consider to improve:
- adding a new cube could automatically switch to first available chart type instead of keeping old, not available one,
- there can be "NaN" values in the tooltip, maybe they should be shown with "-"?
Screenshot 2024-03-06 at 09 26 53

app/rdf/query-search.ts
```diff
@@ -221,9 +241,9 @@ export const searchCubes = async ({
       }
   `.GROUP().BY`?iri`.THEN.BY`?title`.THEN.BY`?status`.THEN.BY`?datePublished`
     .THEN.BY`?description`.THEN.BY`?publisher`.THEN.BY`?creatorIri`.THEN
-    .BY`?creatorLabel`.prologue`${pragmas}`
-    .prologue`#pragma join.bind off`; // HOTFIX WRT Stardog v9.2.1 bug see https://control.vshn.net/tickets/sbar-1066
+    .BY`?creatorLabel`.prologue`${pragmas}`.prologue`#pragma join.bind off`; // HOTFIX WRT Stardog v9.2.1 bug see https://control.vshn.net/tickets/sbar-1066
 
+  console.log(scoresQuery.build());
```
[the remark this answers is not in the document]
ptbrowne: Removed after, thanks.

app/configurator/components/add-dataset-dialog.tsx
ptbrowne: Removed after, thanks

app/graphql/hooks.ts
bprusinowski: Thanks 💯

bprusinowski:
One additional small thing I noticed related to tooltip: it looks like the position is wrong in [this chart](https://visualization-tool-git-feat-add-dataset-ixt1.vercel.app/en/v/8jq4PNagV_mo?dataSource=Prod) after merging two cubes? 👀 Not sure if related to this PR, just a heads up 👍

app/configurator/configurator-state.tsx
ptbrowne: Thanks, again good catch, the fields had a typo. I consolidated the fields and extracted them in config-types, where they are defined. I still find it brittle though.

ptbrowne:
I am going to look for the tooltip problem in another PR, thanks for the heads up.

PR #1389, opened 2024-03-06
perf: Do not include other dimension iris when fetching dimension values

bprusinowski:
After a discussion with @Rdataflow, the issue that originally led to the inclusion of other dimension iris in the dimension values query was due to the data issue; it turns out the query doesn't need to be changed (due to performance reasons), but rather the data.

See the original issue in [PR description](https://github.com/visualize-admin/visualization-tool/pull/1360) and [here](https://github.com/visualize-admin/visualization-tool/pull/1360#discussion_r1504098838).

### DIFF
--- a/app/rdf/query-dimension-values.ts
+++ b/app/rdf/query-dimension-values.ts
@@ -306,57 +306,39 @@ export const getQueryFilters = (
     return "";
   }
 
-  let i = 0;
-  const filterDimensionIris = filtersList.map(([iri]) => iri);
-  // Also include other dimensions to make sure we don't return values that could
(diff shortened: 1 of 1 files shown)


Rdataflow:
@bprusinowski LGTM

FYI: we managed to patch the cube https://visualization-tool-git-perf-remove-other-dimensions-342a6d-ixt1.vercel.app/create/new?copy=tNfm1XUWS3_q by inserting ""^^cube:Undefined
drawback: currently the cube suffers some datatype inconsistencies with the shape but for simple tests like this it may be less harmful than incomplete cubes...

ptbrowne (approved):

PR #1391, opened 2024-03-06
fix: Redirects to latest cube

### LINKED ISSUE #1390: Can't preview an old version of a cube
**Describe the bug**
It's not possible to preview an old version of a cube (cube not found error).

**To Reproduce**
1. Go to https://test.visualize.admin.ch/en/browse?dataset=https://environment.ld.admin.ch/foen/nfi/nfi_C-20/cube/2023-2&dataSource=Prod
2. See error (cube not found).
3. Go to https://test.visualize.admin.ch/_cube-checker?cubeIri=https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fnfi%2Fnfi_C-20%2Fcube%2F2023-2&dataSource=Prod
4. See that cube exists.

**Expected behavior**
A latest version of the cube should be loaded (2023-03), instead of throwing an error.

- Visualize environment and version: TEST v3.26.3 (9de2204)
- Browser and version: Safari 17.3 (19617.2.4.11.8)

bprusinowski:
Fixes #1390.

I think the problem was that we didn't treat the cube with version "2023-2" as versioned cube and failed to fetch latest iri. This PR solves this on a SPARQL query level (and removes relying on checking if a cube is versioned by a heuristic).

With this change, we try to look for newest version of the cube if an older version was passed in the URL (respecting published status).

## How to test
**PR**
1. Go to [this link](https://visualization-tool-git-fix-redirects-to-latest-cube-ixt1.vercel.app/en/browse?previous=%7B%22order%22%3A%22SCORE%22%2C%22search%22%3A%22nfi%22%7D&dataset=https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fnfi%2Fnfi_C-20%2Fcube%2F2023-2&dataSource=Prod) (NFI/2023-02).
2. ✅ See that the latest version of the cube has been loaded.

**TEST**
1. Go to [this link](https://test.visualize.admin.ch/en/browse?dataset=https://environment.ld.admin.ch/foen/nfi/nfi_C-20/cube/2023-2&dataSource=Prod) (NFI/2023-02).
2. ❌ See that there's a _cube not found_ error.

### DIFF
--- a/app/rdf/query-latest-cube-iri.ts
+++ b/app/rdf/query-latest-cube-iri.ts
@@ -7,6 +7,7 @@ PREFIX schema: <http://schema.org/>
 
 SELECT ?iri WHERE {
   {
+    # Versioned cube.
     SELECT ?iri WHERE {
       VALUES ?oldIri { <${cubeIri}> }
       ?versionHistory schema:hasPart ?oldIri .
@@ -17,11 +18,24 @@ SELECT ?iri WHERE {
       FILTER(NOT EXISTS { ?iri schema:expires ?expires . } && ?status IN (?oldStatus, <https://ld.admin.ch/vocabulary/CreativeWorkStatus/Published>))
     }
   } UNION {
-    SELECT ?iri WHERE {
-      VALUES ?iri { <${cubeIri}> }
-      ?iri cube:observationConstraint ?shape .
+    {
+      # Non-versioned cube.
+      SELECT ?iri WHERE {
+        VALUES ?iri { <${cubeIri}> }
+        ?iri cube:observationConstraint ?shape .
+        ?iri schema:creativeWorkStatus ?status .
+        FILTER(NOT EXISTS { ?iri schema:expires ?expires . } && ?status = <https://ld.admin.ch/vocabulary/CreativeWorkStatus/Published>)      
+      }
+    }
+  } UNION {
+    {
+      # Version history of a cube.
+      VALUES ?versionHistory { <${cubeIri}> }
+      ?versionHistory schema:hasPart ?iri .
+      ?iri schema:version ?version .
       ?iri schema:creativeWorkStatus ?status .
-      FILTER(NOT EXISTS { ?iri schema:expires ?expires . } && ?status = <https://ld.admin.ch/vocabulary/CreativeWorkStatus/Published>)      
+      ?oldIri schema:creativeWorkStatus ?oldStatus .
+      FILTER(NOT EXISTS { ?iri schema:expires ?expires . } && ?status IN (?oldStatus, <https://ld.admin.ch/vocabulary/CreativeWorkStatus/Published>))
     }
   }
 }
--- a/app/rdf/query-latest-cube-iri.ts
+++ b/app/rdf/query-latest-cube-iri.ts
@@ -34,8 +34,7 @@ SELECT ?iri WHERE {
       ?versionHistory schema:hasPart ?iri .
       ?iri schema:version ?version .
       ?iri schema:creativeWorkStatus ?status .
-      ?oldIri schema:creativeWorkStatus ?oldStatus .
-      FILTER(NOT EXISTS { ?iri schema:expires ?expires . } && ?status IN (?oldStatus, <https://ld.admin.ch/vocabulary/CreativeWorkStatus/Published>))
+      FILTER(NOT EXISTS { ?iri schema:expires ?expires . } && ?status = <https://ld.admin.ch/vocabulary/CreativeWorkStatus/Published>)
     }
   }
 }
--- a/app/rdf/query-latest-cube-iri.ts
+++ b/app/rdf/query-latest-cube-iri.ts
@@ -8,7 +8,7 @@ PREFIX schema: <http://schema.org/>
 SELECT ?iri WHERE {
   {
     # Versioned cube.
-    SELECT ?iri WHERE {
+    SELECT ?iri ?version WHERE {
       VALUES ?oldIri { <${cubeIri}> }
       ?versionHistory schema:hasPart ?oldIri .
       ?versionHistory schema:hasPart ?iri .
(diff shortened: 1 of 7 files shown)


Rdataflow:
huh, the current state of this PR breaks unversioned cubes like i.e. eletricity price

bprusinowski:
Thanks for catching this @Rdataflow, I corrected the query 🤞

Rdataflow:
@bprusinowski according to my tests it works as expected now for versioned (superseeded and current) [i.e. NFI] as well as non-versioned [i.e. electricity prices] 👍

app/rdf/query-latest-cube-iri.ts
```diff
@@ -0,0 +1,44 @@
...
+  } UNION {
+    {
+      # Version history of a cube.
+      VALUES ?versionHistory { <${cubeIri}> }
+      ?versionHistory schema:hasPart ?iri .
+      ?iri schema:version ?version .
+      ?iri schema:creativeWorkStatus ?status .
+      ?oldIri schema:creativeWorkStatus ?oldStatus .
+      FILTER(NOT EXISTS { ?iri schema:expires ?expires . } && ?status IN (?oldStatus, <https://ld.admin.ch/vocabulary/CreativeWorkStatus/Published>))
+    }
```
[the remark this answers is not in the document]
bprusinowski: Sure, with this part we also catch cases like `https://culture.ld.admin.ch/sfa/StateAccounts_Function`, where we specify like to a cube's version history :)

We had this defined in one of the E2E tests, I think it might have been used in some links from OpenData.swiss datasets 👀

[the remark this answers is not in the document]
bprusinowski: Good catch, thanks! I corrected this 👍

app/rdf/query-latest-cube-iri.ts
```diff
@@ -0,0 +1,44 @@
...
+  } UNION {
+    {
+      # Non-versioned cube.
+      SELECT ?iri WHERE {
+        VALUES ?iri { <${cubeIri}> }
+        ?iri cube:observationConstraint ?shape .
+        ?iri schema:creativeWorkStatus ?status .
+        FILTER(NOT EXISTS { ?iri schema:expires ?expires . } && ?status = <https://ld.admin.ch/vocabulary/CreativeWorkStatus/Published>)      
+      }
```
Rdataflow, line 28: this subquery would imply to show non-versioned cubes only if they would be Published... (line 27)
this condition may be relaxed to show draft cubes of any status.

... then this will also require an additional `FILTER NOT EXISTS { ?versionHistory schema:hasPart ?iri . } ` to mutually exclusive with the versioned cube query and not interfere by adding draft cubes to the verioned cube case.

app/rdf/query-latest-cube-iri.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+  } UNION {
+    {
+      # Version history of a cube.
+      VALUES ?versionHistory { <${cubeIri}> }
+      ?versionHistory schema:hasPart ?iri .
+      ?iri schema:version ?version .
+      ?iri schema:creativeWorkStatus ?status .
+      FILTER(NOT EXISTS { ?iri schema:expires ?expires . } && ?status = <https://ld.admin.ch/vocabulary/CreativeWorkStatus/Published>)
```
Rdataflow, line 37: this would redirect version history to published cubes exclusively. 
OTOH there might be cube histories with only draft cubes. 
what do you think might this something to ORDER BY DESC ?

app/rdf/query-latest-cube-iri.ts
```diff
@@ -0,0 +1,43 @@
+/** Creates SPARQL query to fetch latest cube iri.
+ * Works for both versioned and unversioned cubes.
+ */
+export const getLatestCubeIriQuery = (cubeIri: string) => {
+  return `PREFIX cube: <https://cube.link/>
+PREFIX schema: <http://schema.org/>
+
+SELECT ?iri WHERE {
```
Rdataflow, line 8: @bprusinowski wouldn't it be useful to also output the `?shape` in this query?
use case: reuse to fetch the shape directly - or potentially `DESCRIBE ?cube ?shape` in a single query...

app/rdf/query-latest-cube-iri.ts
```diff
@@ -0,0 +1,43 @@
...
+      # Version history of a cube.
+      VALUES ?versionHistory { <${cubeIri}> }
+      ?versionHistory schema:hasPart ?iri .
+      ?iri schema:version ?version .
+      ?iri schema:creativeWorkStatus ?status .
+      FILTER(NOT EXISTS { ?iri schema:expires ?expires . } && ?status = <https://ld.admin.ch/vocabulary/CreativeWorkStatus/Published>)
+    }
+  }
+}
+ORDER BY DESC(?version)
```
Rdataflow, line 41: IIUC `?version` is not availble here unless `SELECT`ed in the subqueries

app/rdf/query-latest-cube-iri.ts
bprusinowski: Yes, this would be the case. I think it's something I would expect – if there's a link to an unversioned cube version and somebody clicks it, they would open a dataset preview page. By default we don't show draft datasets in the search but require an explicit action to enable them and this logic would not match that behavior 👀

As we don't have a way to specify status of a cube in the URL, I think it makes sense to default to published cubes, especially in Visualize PROD. What do you think? Otherwise we would need to implement a way to keep cube status in the URL.

app/rdf/query-latest-cube-iri.ts
bprusinowski: Again, I am not sure if we should default to showing draft datasets in Visualize PROD, even if they are newer than published versions. I think it makes perfect sense when it comes to versioned cubes, as we can exactly know its status, if there's a "generic" IRI, I would assume it should led to last published cube. Let me know if that's not the expectation 👍 

How would it work with ordering? I think we would need to somehow know the "position" of a given status, right?

app/rdf/query-latest-cube-iri.ts
bprusinowski: This query is currently used in places where we don't actually need to fetch shape at all (client-side redirect and in `LightCube`) 👍

Rdataflow: just thinking ahead :smile: - no current need (yet)

app/rdf/query-latest-cube-iri.ts
Rdataflow: non-versioned cubes could also be Draft (initially) especially on LINDAS TEST or INT. 
it won't hurt to able to display them (once a user clicked on those links)

app/rdf/query-latest-cube-iri.ts
Rdataflow: Based on the cubeHistory I'd show Draft only in cases no Published are available :+1: 
i.e. subquery
```sparql
SELECT ?iri {
    VALUES ?versionHistory { <${cubeIri}> }
      ?versionHistory schema:hasPart ?iri .
      ?iri schema:version ?version .
      ?iri schema:creativeWorkStatus ?status .
      FILTER(NOT EXISTS { ?iri schema:expires ?expires . } )
} 
ORDER BY DESC (?status) DESC(?version) # order by status then version (to prefer Published over latest)
LIMIT 1
```

app/rdf/query-latest-cube-iri.ts
Rdataflow: @bprusinowski won't this come into play once reused for several cases i.e. for Components or the like?
IIUC at least here in the resolver the shape fetching is triggered in a sequence
https://github.com/visualize-admin/visualization-tool/blob/28fedbd216b6ad1fd41196ab7334f181d44513b3/app/graphql/resolvers/rdf.ts#L96-L97

app/rdf/query-latest-cube-iri.ts
bprusinowski: I hope the new query should correctly catch all the cases 🤞

app/rdf/query-latest-cube-iri.ts
```diff
@@ -17,27 +17,31 @@ SELECT ?iri WHERE {
...
-        VALUES ?iri { <${cubeIri}> }
-        ?iri cube:observationConstraint ?shape .
+      # Version history of a cube.
+      SELECT ?iri ?status ?version WHERE {
+        VALUES ?versionHistory { <${cubeIri}> }
+        ?versionHistory schema:hasPart ?iri .
+        ?iri schema:version ?version .
         ?iri schema:creativeWorkStatus ?status .
-        FILTER(NOT EXISTS { ?iri schema:expires ?expires . } && ?status = <https://ld.admin.ch/vocabulary/CreativeWorkStatus/Published>)      
+        FILTER(NOT EXISTS { ?iri schema:expires ?expires . })
```
Rdataflow, line 29: @bprusinowski this subquery is also responding in versioned cases - this should be prevented to not interfere and spill over Draft in versioned cases.

i.e. adding `FILTER NOT EXISTS { ?versionHistory schema:hasPart ?iri . }` should do the trick.

Rdataflow: @bprusinowski sorry - I've overlooked the new order of the 3 subqueries...

ptbrowne (approved):
Thanks !

PR #1392, opened 2024-03-07
Add dataset improvements in filter and metadata panel

ptbrowne:
Improvements on the add dataset functionality.

- refactor: Extract out make search query
- refactor: Use enum for SearchCubeFilterType
- fix: Use correct value for TemporalDimension
- fix: Use correct component iri when tweaking iri to be joinBy
- feat: ControlTabField handle multi components field
- fix: Missing localized string
- wip fix nan
- feat: Show multiple for vertical axis in case of combo chart + show merged temporal dimension according to time unit
- feat: Use chart spec to known where are the iri attributes
- feat: Show correct dimension label in metadata panel

### DIFF
--- a/app/configurator/components/chart-controls/control-tab.tsx
+++ b/app/configurator/components/chart-controls/control-tab.tsx
@@ -20,7 +20,7 @@ import useEvent from "@/utils/use-event";
 
 type ControlTabProps = {
   chartConfig: ChartConfig;
-  component?: Component;
+  fieldComponents?: Component[];
   value: string;
   onClick: (x: string) => void;
   labelId: string | null;
@@ -31,7 +31,7 @@ type ControlTabProps = {
 export const ControlTab = (props: ControlTabProps) => {
   const {
     chartConfig,
-    component,
+    fieldComponents,
     value,
     onClick,
     checked,
@@ -40,6 +40,8 @@ export const ControlTab = (props: ControlTabProps) => {
     warnMessage,
   } = props;
   const handleClick = useEvent(() => onClick(value));
+
+  const component = fieldComponents?.[0];
   const isActive = overrideChecked(chartConfig, value) ? true : !!component;
   const { upperLabel, mainLabel } = getLabels(
     chartConfig,
@@ -84,9 +86,10 @@ const getLabels = (
   switch (value) {
     case "y":
       if (isComboChartConfig(chartConfig)) {
+        console.log(chartConfig);
         return {
-          upperLabel: null,
-          mainLabel: getFieldLabel("y"),
+          upperLabel: getFieldLabel("y"),
+          mainLabel: null,
         };
       }
     default:
--- a/app/configurator/components/chart-controls/control-tab.tsx
+++ b/app/configurator/components/chart-controls/control-tab.tsx
@@ -11,7 +11,10 @@ import {
   overrideChecked,
 } from "@/configurator";
 import { getFieldLabel } from "@/configurator/components/field-i18n";
-import { getIconName } from "@/configurator/components/ui-helpers";
+import {
+  getIconName,
+  getLabelFromDimension,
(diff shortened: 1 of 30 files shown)


app/rdf/query-search.ts
```diff
@@ -97,7 +97,110 @@ export const searchCubes = async ({
...
+  const scoresQuery = mkScoresQuery(
+    locale,
+    filters,
+    creatorValues,
+    themeValues,
+    includeDrafts,
+    query
+  ); // HOTFIX WRT Stardog v9.2.1 bug see https://control.vshn.net/tickets/sbar-1066
+
+  console.log(scoresQuery.build());
```
bprusinowski, line 110: 🔍

app/graphql/schema.graphql
```diff
@@ -318,8 +318,15 @@ type DataCubeOrganization {
   label: String
 }
 
+enum SearchCubeFilterType {
+  TemporalDimension
+  DataCubeTheme
+  DataCubeOrganization
+  DataCubeAbout
+}
+
```
bprusinowski, line 327: 👍

app/configurator/components/chart-controls/control-tab.tsx
```diff
@@ -84,9 +86,10 @@ const getLabels = (
   switch (value) {
     case "y":
       if (isComboChartConfig(chartConfig)) {
+        console.log(chartConfig);
```
bprusinowski: 👁️

bprusinowski (commented):
Nice changes! 👏 

I noticed two things when checking out the preview deployment:
- changing data filter reloads the whole page and "scrolls up", even when we didn't add a second cube (and we kind of lose track of the filter we changed and its new value),
- see https://visualization-tool-git-feat-add-dataset-ixt1.vercel.app/create/new?edit=aZzqcj_ca00d (log in as `ixt` user first) and remove the second cube – year key dimension is not selected as filter, and it should be 👀

ptbrowne:
Thanks for this Bartosz, the flash of filters is due to previous data not being kept in the filters (the useQuery hook not returning the data if the key changes), previous data behavior should be tweaked for the filters. I'll do this in another PR. I'll also check the year filter not coming back there.

ptbrowne:
In fact, I took a look and saw that if I added the keepPreviousData attribute handling (like in react-query), I could manage not to have the filters "reload" and show a loading indicator. The change of loading the multi-dataset data prior to actually changing the chart config must help there.

app/graphql/hooks.ts
```diff
@@ -40,21 +46,23 @@ export const makeUseQuery =
...
+    const { keepPreviousData } = options;
     const executeQuery = React.useCallback(
       async (options: T) => {
         const result = await executeQueryFn(options.variables, () =>
           setResult((prev) => ({
             ...prev,
             fetching: true,
-            data: prev.queryKey === queryKey ? prev.data : null,
+            data:
+              prev.queryKey === queryKey || keepPreviousData ? prev.data : null,
```
bprusinowski, line 65: 😍

bprusinowski (approved):
Thanks for the changes, the code looks 💯 :)

When testing the deployment preview, I've noticed one more thing which breaks the published chart when having two cubes merged 👀 

1. Go to https://visualization-tool-git-feat-add-dataset-ixt1.vercel.app/en/create/new?cube=https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/9&dataSource=Prod
2. Add "Greenhouse gas emissions by sector and by gas" dataset
3. See that Vertical Axis is empty (no label shown)
4. Publish the chart
5. See `No dimension <joinBy> in cube!` error

As these are not directly related to this PR (the same behavior on TEST), it's good to merge (but would be good to fix before INT deployment) 🙏 🤞

PR #1393, opened 2024-03-07
fix: Keep locale in user actions

bprusinowski:
This PR keeps the currently selected locale when using profile actions (view, edit, copy, share) from the user profile.

## How to test
1. Go to [this link](https://visualization-tool-git-fix-keep-locale-in-user-actions-ixt1.vercel.app/en?dataSource=Prod).
2. Log in (e.g. as `ixt` test user).
3. Navigate to user profile.
4. Edit an example chart.
5. See that the page opened in English (on TEST it opens in German).

### DIFF
--- a/app/login/components/profile-tables.tsx
+++ b/app/login/components/profile-tables.tsx
@@ -199,7 +199,7 @@ const ProfileVisualizationsRow = (props: ProfileVisualizationsRowProps) => {
     const actions: (ActionProps | null)[] = [
       {
         type: "link",
-        href: `/v/${config.key}`,
+        href: `/${locale}/v/${config.key}`,
         label: t({ id: "login.chart.view", message: "View" }),
         iconName: "eye",
         priority:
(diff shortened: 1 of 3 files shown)


app/login/components/profile-tables.tsx
```diff
@@ -199,29 +199,29 @@ const ProfileVisualizationsRow = (props: ProfileVisualizationsRowProps) => {
...
+        href: `/${locale}/create/new?edit=${config.key}`,
         label: t({ id: "login.chart.edit", message: "Edit" }),
         iconName: "edit",
         priority:
           config.published_state === PUBLISHED_STATE.DRAFT ? 0 : undefined,
       },
       {
         type: "link",
-        href: `/v/${config.key}`,
+        href: `/${locale}/v/${config.key}`,
```
ptbrowne, line 224: Thnks for the fix, I thought it should be handled directly by NextLink ?

ptbrowne (approved):

e2e/selectors.ts
```diff
@@ -84,7 +84,7 @@ export const createSelectors = ({ screen, page, within }: Ctx) => {
         screen
           .getByTestId("panel-drawer")
           .within()
-          .findByText("Use abbreviations"),
+          .findByText("Use abbreviations", {}, { timeout: 10_000 }),
```
ptbrowne, line 87: 🎉 for the test being fixed but do you know why it comes so slow ?

bprusinowski: Yes, this is because now the whole sidebar gets unmounted when segmentation component is changed, while previously this was not the case (see PROD) 👀  Maybe this could also be fixed by `keepPreviousData`, but for now it looks like we need to await this checkbox's presence 🤔

ptbrowne: OK thanks, I am going to take a look.

bprusinowski: Thank you! 🙇‍♂️

PR #1399, opened 2024-03-08
Support for multi dataset in metadata panel

ptbrowne:
* Showing information for all the cubes in the metadata panel
* Grouping the dimensions by cube in the metadata panel


app/components/metadata-panel.tsx
```diff
@@ -406,17 +412,47 @@ const Header = ({ onClose }: { onClose: () => void }) => {
...
+  const cubesMetadata = data?.dataCubesMetadata;
+  if (!cubesMetadata) {
+    return null;
+  }
   return (
     <TabPanel className={classes.tabPanel} value="general">
-      <DataSetMetadata dataSetIri={datasetIri} dataSource={dataSource} />
+      {fetching ? <Loading /> : null}
+      {error ? (
+        <Error>{`${error instanceof Error ? error.message : error}`}</Error>
```
bprusinowski, line 447: So `error` could be something different than `Error`? 🤯

app/components/dataset-metadata.tsx
```diff
@@ -19,7 +19,7 @@ import { Icon } from "@/icons";
 import { useLocale } from "@/locales/use-locale";
 import { makeOpenDataLink } from "@/utils/opendata";
 
-export const DataSetMetadata = ({
+export const DatasetMetadata = ({
```
bprusinowski, line 22: 👍

app/components/metadata-panel.tsx
```diff
@@ -470,6 +478,37 @@ const TabPanelData = ({ dimensions }: { dimensions: Component[] }) => {
     return dimensions.map((d) => ({ label: d.label, value: d }));
   }, [dimensions]);
 
+  const locale = useLocale();
+  const grouped = groupBy(
+    dimensions.flatMap((d): Component[] => {
+      if ("originalIris" in d) {
+        return (
+          d.originalIris?.map((o) => ({
```
bprusinowski, line 486: Nit: do we need `?` here, when `originalIris` property exists?

bprusinowski (approved):
LGTM! 👌 

Two things noticed when reviewing deployment preview:
- it looks that dimension values are not shown in the metadata panel (I guess it happened when I introduced `loadValues` property some time ago, it also doesn't work on TEST 😨). Let me know if I should take a look at this – maybe it would make sense to only fetch them when user actually goes to the dimension preview, to not fire potentially expensive queries upfront?
- `joinBy` dimension is not translated 👀

app/components/metadata-panel.tsx
ptbrowne: The property exists but could be undefined ;)

ptbrowne: I moved the check to the if ;)

app/components/metadata-panel.tsx
ptbrowne: Yes, every thing can be throw ;)

```
> const hello = () => { try { throw 'world' } catch (e) { console.log('got thrown', e) } }
undefined
> hello()
got thrown world
```

app/components/metadata-panel.tsx
```diff
@@ -418,7 +422,33 @@ const TabPanelData = ({
...
+        d
+      ): {
+        label: string;
+        value: Component;
+        isJoinByDimension: boolean;
+      }[] => {
+        if (
+          "isJoinByDimension" in d &&
+          d.isJoinByDimension &&
+          "originalIris" in d
```
bprusinowski, line 437: Nit: I think we have this check in several places now, maybe would make sense to extract to something like `isJoinByDimension`?

app/locales/de/messages.po
```diff
@@ -1445,32 +1449,32 @@ msgid "time-units."
 msgstr ""
 
 #: app/configurator/components/ui-helpers.ts
-msgstr "time-units.Day"
-msgstr "Tag"
+msgid "time-units.Day"
+msgstr ""
```
bprusinowski, line 1453: Should this be removed? 🤔

bprusinowski (approved):
LGTM 🎉 

Just one small thing from deployment preview: it looks like the joined dimension label is missing translation?
Screenshot 2024-03-12 at 15 12 29

app/charts/combo/combo-line-dual-state.tsx
```diff
@@ -53,6 +53,8 @@ export type ComboLineDualState = CommonChartState &
     maxRightTickWidth: number;
   };
 
+const mean = (arr: number[]) => arr.reduce((a, b) => a + b, 0) / arr.length;
```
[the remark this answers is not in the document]
ptbrowne: Done after ;)

app/components/metadata-panel.tsx
ptbrowne: Thanks, good idea, I extracted the guard in [`3e01901` (#1399)](https://github.com/visualize-admin/visualization-tool/pull/1399/commits/3e01901b169690c5e25a83fb324cd4089908b15c)

app/locales/de/messages.po
ptbrowne: Thanks, removed.

ptbrowne:
Thanks @bprusinowski , I've added the missing translations.

PR #1410, opened 2024-03-13
perf: Optimize PossibleFilters query

### LINKED ISSUE #1406: Rewrite the PossibleFilters query to fetch everything in one go
Currently, we sequentially fire queries to get possible filters if a given combination doesn't yield at least one observation – this could be improved by using a different approach as suggested by @Rdataflow here:
- https://s.zazuko.com/2pAcDjW
- https://s.zazuko.com/3g37YE9
- https://s.zazuko.com/3kFuPLa

bprusinowski:
Closes #1406

---

This PR rewrites the `PossibleFilters` query so it is much more performant and:
- no longer fetches the whole cube shape,
- uses only one query instead of "cascading" queries that removed dimensions one after the other when no observation was found (thanks @Rdataflow for the idea 🙂),
- unversions found observation in the same query in which possible filters are determined.


app/rdf/query-possible-filters.ts
```diff
@@ -0,0 +1,123 @@
...
+    })
+    .join("\n")}
+}
+ORDER BY ${
+    // Order by the boolean `d` variables, so that the first result is the one
+    // with the most matching dimensions, keeping the order of the dimensions
+    // in mind, to mirror the cascading behavior of the filters.
+    queryFilters.map(({ i }) => `DESC(?d${i})`).join(" ")
+  }
+LIMIT 1`;
```
ptbrowne, line 99: Cool 👍 

Would it be possible to have a unit test for the query ? (checking the string). I think it's helpful so that for example I can copy/paste the query in Lindas to check it/understand better the result.

ptbrowne (approved):
LGTM 👏 👏  Just a suggestion, I saw the unit test, and wonder if we could maybe use a real cube so that we can copy/paste the query and run it against a live cube ?

app/rdf/query-possible-filters.ts
bprusinowski: Yes, done in later commit! 😄

bprusinowski:
Good point @ptbrowne, done :)

Rdataflow:
@bprusinowski nit: k6 tests fail now due to `SingleFilters` I guess

bprusinowski:
Fixed on `main`, thanks for noticing!

PR #1411, opened 2024-03-14
perf: Various optimizations

### LINKED ISSUE #1407: Check why dimensions fetching appears to not happen in parallel
Currently, there's an impression that somehow multiple dimension fetching is not done in parallel; we should investigate if that's the case and fix if appropriate, either by modifying the code or using a separate query to get dimension values for multiple dimensions in one go.

Example query that works when we have no filters: https://s.zazuko.com/3d9D51g

### LINKED ISSUE #1408: Geo-related performance optimizations
- Currently, we still use `cubeDimensionByIri -> geoShapes` resolver chain; we could rewrite this to be a single query to eliminate the resolver chain completely
- `geoCoordinates` might be fetched along with dimension values, if makes sense
- Check the TopoJSON-related parts; do we need to compute `mesh`?

bprusinowski:
Closes #1407, closes #1408

---

This PR:
- cleans up some of the code related to the map chart,
- removes unnecessary `mesh` computation,
- optimizes dimensions values fetching by using `Promise.all` instead of `for ... await`,
- refactors geo-related queries to remove resolver chains,
- updates several packages.


app/rdf/query-geo-shapes.ts
```diff
@@ -62,110 +60,105 @@ export const createGeoShapesLoader =
...
-        })) as RawGeometry[];
-      } catch (e) {
-        console.error(e);
-      }
+    try {
+      rawGeometries = (await geometriesQuery.execute(sparqlClient.query, {
+        operation: "postUrlencoded",
+      })) as RawGeometry[];
+    } catch (e) {
+      console.error(e);
```
ptbrowne, line 97: Nit: this try/catch can be a bit simplified by using catch() directly. This way you do not have to declare rawGeometries before hand.

```
const rawGeometries = await stuff().catch(e => {
        console.error(e)
	return []
})
```

app/rdf/queries.ts
```diff
@@ -174,23 +174,18 @@ export const createCubeDimensionValuesLoader =
...
           cache,
-        }
-      );
-      result.push(dimensionValues);
-    }
-
-    return result;
+        });
+      })
+    );
```
ptbrowne, line 188: 👍

CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ You can also check the [release page](https://github.com/visualize-admin/visuali
 
 ## Unreleased
 
+- Performance
+  - Dimension values are now fetched in parallel in case of fetching for several dimensions at once
+  - Refactored shapes and coordinates fetching to not rely on resolver chains
```
ptbrowne, line 14: Was it not performant because we would first load the dimension, and then the shapes ? And now we do not have to wait for the dimension ?

ptbrowne (approved):
👍 Thanks !

CHANGELOG.md
bprusinowski: Technically the improvement is not really noticeable, but now we have one resolver for cube, dimension and shapes, instead of the situation visible in the screenshot. We used to have lots of resolver chains, but at some point removed them and switched to scalars to improve things a bit, this was the last remaining piece of code to adapt the same pattern :)

Screenshot 2024-03-15 at 13 55 41

Here it's a first step towards optimizing it further and again, not fetch the whole shape, but rather "picked" properties to really gain visible improvement. But this was more complex to refactor, as we need the whole shape in several methods that are used in this query, so could be something to focus on in the next sprint 👍

bprusinowski:
Merging after running and passing the E2E tests locally. We might need to look into why they keep failing (maybe still some tests are affected by filters panel loading) – or introduce some retrials? The problem seems to mostly appear when several tests are run together (`yarn run e2e`) – when running single tests locally (`yarn run e2e e2e/tooltip.spec.ts`), they usually work without failures.

cc @ptbrowne

PR #1414, opened 2024-03-15
Storybook

ptbrowne:
## Goals

- Extract component documentation out of catalog
- Catalog is only kept for "application user guide" right now. It could be completely moved to storybook in a second step.

## Comments

- Tried to use vite but it did not work well
- Used where applicable .mock.ts files that are used inside stories. In the future, we should take a look at how we could
  share more mocks between components
- Storybook can be sometimes a bit unstable and hang, see "Common errors" section
- Could reuse the "theme.stories.tsx" from "uis" :+1:
- I tried to follow the components / organisms / pages nomenclature
- Storybook is built separate from the main application. It is built into the public/ folder so that it is available via /storybook.

## Common errors when developing on Storybook

- Providers do not work
  - Need to wrap to use Story or Canvas so that they are wrapped by the decorators configured in preview.ts
- Canvas' `of` does not work
  - Need to use <Meta /> in the docs file
  - Story need to be loaded in main.ts (beware of tsx/mdx extension). This error was because during developement, I manually added one by one each story, otherwise there were errors. Now that there is a glob for all *.docs.mdx and *.stories.tsx, it should be OK.

- At least one component must be exported otherwise the story file can be reported missing ()
- Sometimes, loading a story hangs. Restarting storybook, and sometimes the browser helps.


app/docs/chart-selection-tabs.docs.mdx
```diff
@@ -0,0 +1,44 @@
...
+      {
+        ...palmerPenguinsFixture.data,
+        state: "PUBLISHED",
+      }
+    }
+    allowDefaultRedirect={false}
+  >
+    <ChartSelectionTabs />
+  </ConfiguratorStateProvider>
+</ReactSpecimen> */}
```
bprusinowski: 👀

app/docs/fixtures.ts
```diff
@@ -475,55 +475,91 @@ export const tableObservations = [
   },
 ];
 
-export const tableMeasures = [
+export const tableMeasures: Measure[] = [
```
bprusinowski, line 478: 👍

bprusinowski (approved):
Super cool, thanks! 🥇 I'll also take a look at the deployment preview once it boots up 👍

adintegra:
Very cool @ptbrowne 🎉 Thanks for this!

PR #1420, opened 2024-03-19
fix: Include all columns in full data download

### LINKED ISSUE #1416: Download Full xls : Dimensions are missing
Some Dimensions (dimensions that are not Key dimensions) are missing by download Full dataset on Visualize INT 

**To Reproduce**
1. Open https://www.visualize.admin.ch/en/v/t6ee4XGN_Ofz
2. Download Full datset in csv and in xls
3. look at Columns : They are 8 Columns
4. Opne same cube on INT : https://int.visualize.admin.ch/en/v/8m_Dl0WtYvjx
5. Download Full datset in csv and in xls
6. look at Columns : 4 columns are missing

**Expected behavior**
By Download "Full Dataset" all Dimensions which are integreted in the cube must be downloaded. See Behaviour on PROD with same Cube. 

**Screenshots or video**
Download with PROD :
image

Download same Cube with INT
image

- Visualize environment and version:v3.27.1 [8c3e7f9]
- Browser and version : Edge

**Additional context**

bprusinowski:
Closes #1416


app/components/chart-footnotes.tsx
```diff
@@ -211,7 +211,10 @@ export const ChartFootnotes = ({
                     <DataDownloadMenu
                       dataSource={dataSource}
                       title={dataCubeMetadata.title}
-                      filters={filters}
+                      filters={
+                        filters ??
+                        chartConfig.cubes.map((cube) => ({ iri: cube.iri }))
+                      }
```
ptbrowne, line 217: Not sure I understand this ? Shouldn't we also "mix" the filters with the cubes ?

ptbrowne: Here we do not use getFullDataDownloadFilters  ?

bprusinowski: Hmmm, here we either pass `filters` if they are present, or "construct" them in case they are missing. In `getFullDataDownloadFilters` we need to pass `filters` as an argument, so couldn't apply this to `chartConfig.cubes.map((cube) => ({ iri: cube.iri }))`. Let me know if it makes sense or if I missed something 👀

bprusinowski: I refactored a bit to introduce a new function, let me know if it makes more sense now @ptbrowne 🙇‍♂️

app/components/data-download.tsx
```diff
@@ -252,12 +252,9 @@ const DataDownloadInnerMenu = ({
...
           subheader={<Trans id="button.download.data.all">Full dataset</Trans>}
           fileName={`${fileName}-full`}
-          filters={filters?.map((d) => {
-            return {
-              iri: d.iri,
-              componentIris: d.componentIris,
-            };
-          })}
+          // We need to include every cube column in full dataset download (client's
+          // request), so we do not pass any filters here.
```
ptbrowne: `we do not pass any componentIris here` ?

app/components/data-download.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { RDFCubeViewQueryMock } from "@/test/cube-view-query-mock";
+RDFCubeViewQueryMock;
+
+// eslint-disable-next-line import/order
+import get from "lodash/get";
```
ptbrowne, line 5: For my information, why do we need to disable import/order here ?

bprusinowski: That's a good question, it had to do with the order of mocking things – in some tests, this mock had to be the first thing that's mocked, which clashes with the default import order we define via eslint. In fact, I believe it makes more sense to include this mock in jest setup – I adjusted this in https://github.com/visualize-admin/visualization-tool/pull/1420/commits/7dfce84e672dab5e816e434037d797e98581b200 👀

ptbrowne: Looks good, thanks 👍

ptbrowne (approved):
Thanks 👍

bprusinowski:
Merging to `main`, as failed CI E2E tests work fine locally.

PR #1428, opened 2024-03-21
docs: Initialize developer docs

bprusinowski:
This PR is a proposition to introduce a new type of _closer-to-the-code_ docs (developer docs). For rationale, see https://github.com/visualize-admin/visualization-tool/blob/docs/developer-docs/docs/home.md

### Next steps, if accepted
- [ ] Integrate with Storybook (or consider other system like [quartz](https://github.com/jackyzha0/quartz)). It would make sense not to introduce another tool, but on the other hand, if Obsidian would be the tool of choice, there are e.g. some custom link blocks that do not seem to render correctly in regular Markdown 👀


bprusinowski:
@ptbrowne I decided to open a PR for the proposal for developer docs as discussed offline; let me know what you think, we can also discuss next week 👍

docs/dimensions/dimension.md
```diff
@@ -0,0 +1,2 @@
+
+
```
ptbrowne, line 2: To remove ?

docs/dimensions/values/temporal entity.md
```diff
@@ -0,0 +1,7 @@
+A special type of [[dimension value]], which has a character-based [[schema > position]] and uses shared values (e.g. https://ld.admin.ch/time/month/2000-02).
+
+Can appear in [[TemporalEntityDimension]].
+
+### assumptions
+
+- [[schema > position]] contains the full, formatted date – according to its [[time unit]]
```
ptbrowne, line 7: Nit: In general, I think prefer a bit larger documents with a table of contents.

docs/refactor ideas.md
```diff
@@ -0,0 +1,2 @@
+- [ ] Convert dimensions to classes or introduce new properties to their types to consolidate their behavior and characteristics, instead of having a lot of separated functions w/o real connection
```
ptbrowne, line 1: I would put optimization and refactor in the same file

ptbrowne (approved):
Thanks for starting this Bartosz. I think we should consolidate a bit more into files so that it not too scattered, and so that we do not have the "in which file do I put this thing" problem.

^ This is a nit. LGTM 👍

docs/refactor ideas.md
bprusinowski: I updated the file title to contain "performance optimizations", I think they should still be separate from technical refactors?

bprusinowski:
Thanks for reviewing @ptbrowne!

I think the original idea I had in mind was to rely on smaller files linked together by `[[xyz]]` tags, which allows you to quickly jump between files and having them open in several tabs, instead of scrolling back and forth in larger documents.

My slight concern with larger documents with table of contents would be more maintenance needed to keep the TOC in sync with content, as I didn't find a way to auto-generate it, like in Notion (there as some Obsidian plugins, but still you'd need to install and run when contents changes). On the other hand, I also see the issue with "in which file did I put this", but maybe if we link the things properly, that problem would not materialize badly 😅 

I'll merge the PR in its current state, and maybe we can see how they work, and if we find this valuable, and adjust as needed 🙏

PR #1436, opened 2024-04-05
style: UI improvements

### LINKED ISSUE #1426: Block the temporal range on axis with option "Use most recent" (de-activated)
It is currently possible to reduce the temporal range of a graph at x-axis level, but not to lock the temporal range at the maximum available at the time of publication. If new data with a wider temporal range are published, the graph will automatically show the full time ramnge.

As a web editor, I would like to be able to embed a graphic which doesn’t display the most recent year on the x-axis when the cube is updated with a wider temporal range. 

Definition of Done :
- In visualize, in Edit mode, the option "Use most recent" is not only available for filters (already implemented) but also for axes if a temporal dimension can be selected, and for all chart types.
- This option is activated per default

To test :
- Choose a dataset with a temporal Dimension.  
- Check if the option “use most recent date” is available and activated per default wherever the temporal dimension can be selected (axes, filters -> already implemented for filters).
- Publish the visualisation and save the URL to share
- Import data with one more year and publish the Cube again
- Test the URL -> the last available year should be displayed
- Edit the visualization and deactivate the option “use most recent year”  
- Publish the visualisation and save the URL
- Import data with one more year and publish the Cube again
- Test the URL -> the last available year should not be displayed
- Repeat for each chart Type and wherever a temporal dimension can be selected (axes, filters).

[Figma Prototype
](https://www.figma.com/proto/Hn9wvuEYLUmdJtg4KwZYUx/visualize.admin.ch---design?page-id=8%3A145669&type=design&node-id=3803-125459&viewport=5091%2C-4790%2C0.91&t=ij8ravmJGEIKgGdF-1&scaling=contain&starting-point-node-id=3803%3A125459&mode=design)

[Design Section
](https://www.figma.com/file/Hn9wvuEYLUmdJtg4KwZYUx/visualize.admin.ch---design?type=design&node-id=3803-125279&mode=design&t=wVtTQmKrVDyyulWM-4)

bprusinowski:
This PR introduces some UI improvements to match the implementation with [the design](https://www.figma.com/file/Hn9wvuEYLUmdJtg4KwZYUx/visualize.admin.ch---design?type=design&node-id=3803-125279&mode=design&t=e8xUca8wfq1dtnoe-0).

- `ColorRamp` simplification,
- icon updates,
- filter panel drag and drop controls alignment,
- missing temporal filter index fix.


app/charts/map/map-legend.tsx
```diff
@@ -705,7 +705,7 @@ const ContinuousColorLegend = ({
           width={width - MARGIN.left - MARGIN.right}
           height={COLOR_RAMP_HEIGHT}
           colorInterpolator={getColorInterpolator(palette as any)}
-          nbClass={width - MARGIN.left - MARGIN.right}
+          nSteps={width - MARGIN.left - MARGIN.right}
```
ptbrowne, line 708: Cannot see any steps at https://visualization-tool-git-feat-most-recent-x-axis-ixt1.vercel.app/storybook/index.html?path=/story/components-colorramp--discrete, is this because nSteps is not changed there ?

app/themes/federal.tsx
```diff
@@ -614,6 +614,24 @@ theme.components = {
...
+          }}
+        >
+          <Icon
+            name="chevronDown"
+            size={16}
+            style={{ transform: "scale(1.5)" }}
+          />
+        </span>
+      ),
+    },
```
ptbrowne, line 637: I checked it at https://visualization-tool-git-feat-most-recent-x-axis-ixt1.vercel.app/storybook/index.html?path=/story/design-system-theme--components and it seems a bit weird ?  In the app it is good 🤔

app/components/move-drag-button.tsx
```diff
@@ -0,0 +1,23 @@
+import { Box, BoxProps } from "@mui/material";
+
+import { Icon } from "@/icons";
+
+type MoveDragButtonProps = {
+  className?: string;
+  dragButtonProps?: BoxProps;
```
ptbrowne, line 7: I would not have an inner dragButtonProps, but have an icon className; since the MoveDragButton is the container, I expect props to be passed to the parent component.

app/components/form.tsx
```diff
@@ -373,90 +378,93 @@ export const Select = ({
...
+                    {opt.label}
+                  </span>
+                  {opt.disabledMessage && (
+                    <DisabledMessageIcon message={opt.disabledMessage} />
+                  )}
+                </MenuItem>
+              );
+            })}
+          </MUISelect>
+          {sideControls}
```
ptbrowne, line 466: I have the impression that sideControls should be wrapped into a flexShrink: 0 container.

image

Created from : https://visualization-tool-git-feat-most-recent-x-axis-ixt1.vercel.app/en/browse?dataset=https%3A%2F%2Fenvironment.ld.admin.ch%2Ffoen%2Fnfi%2Fnfi_T-changes%2Fcube%2F2024-1

app/configurator/components/field.tsx
```diff
@@ -254,61 +266,62 @@ export const DataFilterSelect = ({
...
+        topControls={topControls}
+        sideControls={sideControls}
+        open={isOpen}
+        onClose={handleClose}
+        onOpen={handleOpen}
+        loading={loading}
+        {...fieldProps}
+        value={usesMostRecentValue ? maxValue : fieldProps.value}
+      />
+    </div>
```
ptbrowne, line 324: A bit difficult for me to check the diff because of the ident changes, I am wondering if we should have chart-configurator field examples in the storybook ? I think it can be used to check the "extreme" cases, for example with the screenshot I posted where a long label makes the drag handle disapppear.

ptbrowne (changes requested):
Thanks for aligning the design.

Unfortunately, I now see an horizontal scroll bar in the side panel (I have the Show scroll bars set to Always in mac os settings, to detect this kind of problem).

This should be fixed before merging ;) Otherwise I think this is good. I think an improvement could be to have stories for the components with topControl and sideControl.

image

bprusinowski:
Thanks for the review @ptbrowne! I'll go though the detailed comments, but the issue you mentioned with vertical scroll was fixed in another PR, that could be merged to this one (https://github.com/visualize-admin/visualization-tool/pull/1437/commits/ae5cf439301681986b2c2787087e112210ee0c8d) 👀

app/charts/map/map-legend.tsx
bprusinowski: Looks like it, I updated the story to infer the types based on the component in https://github.com/visualize-admin/visualization-tool/pull/1436/commits/bec64dcf70c3a3b599b198b4005c8ee90811fa10 👍

app/configurator/components/field.tsx
bprusinowski: Yes, I think it would be handled in #1439 👍

app/themes/federal.tsx
bprusinowski: Not super sure why it's like that, I assume it's due to a fact that we render select element in a custom way – and generally we don't use native MUI elements, but have our own wrappers in the `Form` file. Maybe we could have a note somewhere to use "our" components, and not native ones? I tried to fix this problem, but then there are some padding issues when select element is expanded 🤔

bprusinowski:
@ptbrowne should be good to review again 🙏

ptbrowne (approved):
👍 Thanks !

PR #1438, opened 2024-04-09
Label docker image

ptbrowne:
Right now, it is difficult given an image retrieved via docker pull registry.ldbar.ch/visualize-admin/visualization-tool:release
to know which is the version. Using labels, we should be able to know exactly the version by doing docker inspect on the image.

### DIFF
--- a/scripts/docker-build-push.sh
+++ b/scripts/docker-build-push.sh
@@ -16,6 +16,8 @@ set -x
 
 # Build docker image, passing arguments defined in the CI 
 docker build \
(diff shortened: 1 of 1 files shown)


bprusinowski (approved):
LGTM! Wondering why the `yarn.lock` has changed? 👀

ptbrowne:
> Wondering why the yarn.lock has changed? 👀

I ran npx yarn deduplicate (trying to reduce the size of node modules) but it seems it triggers Typescript warnings then, I removed this from this PR.

Issue #1439, opened 2024-04-09
Review and add missing Storybook stories


bprusinowski:
It would be good to review the current state of our Storybook stories and add them and the documentation for missing components.

              I checked quickly for a storybook for the legend and could not find any, maybe something for later ?

_Originally posted by @ptbrowne in https://github.com/visualize-admin/visualization-tool/pull/1435#discussion_r1557676957_

PR #1443, opened 2024-04-10
refactor: Use CSS instead of JavaScript to align chart elements in dashboards

### LINKED ISSUE #1442: Use CSS subgrid to align charts in dashboard
Currently, we rely on JavaScript logic to align the chart elements across different columns in a dashboard layout. This could be achieved in a simpler and more efficient way with the use of CSS subgrid display.

bprusinowski:
Closes #1442

This PR swaps JS logic to keep chart elements in sync in the two columns of the Tall dashboard layout for [CSS subgrid](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_grid_layout/Subgrid), which is both simpler and more performant.

As we have a nested layout and elements we want to align are not direct descendants of the grid layout that contains them (unavoidable due to e.g. drag and drop functionality), some of the container elements now use the [contents](https://caniuse.com/css-display-contents) display to make them "invisible" to the grid itself.


ptbrowne (approved):
👍 Code looked good to me, do we have a story for a dashboard ?

bprusinowski:
Thanks! We don't, I briefly tried to add one but had some problems with local GQL server not being accessible from Storybook. Maybe this could be a part of #1439?

PR #1445, opened 2024-04-11
chore: Reduce Docker image size

### LINKED ISSUE #1444: Reduce Docker image size
Implement some optimizations to make the Docker image smaller and faster to build.

bprusinowski:
Closes #1444

This PR switches to a [multi-stage Docker build](https://docs.docker.com/build/building/multi-stage/) to drastically reduce Docker image size: 9.78GB -> 520MB (-95%), which should make the TEST / INT CI pipelines faster.

Approach was taken from the official Next.js Dockerfile example from [here](https://github.com/vercel/next.js/blob/canary/examples/with-docker/Dockerfile?rgh-link-date=2024-04-12T10%3A30%3A31Z).


Dockerfile
```diff
@@ -39,13 +45,30 @@ ENV NEXTAUTH_SECRET=$NEXTAUTH_SECRET
...
+COPY --from=builder /src/app/.next/standalone ./
+COPY --from=builder /src/app/.next/static ./app/.next/static
+
+USER nextjs
 
 EXPOSE 3000
+
+ENV PORT 3000
+
+CMD HOSTNAME="0.0.0.0" node app/server.js
```
ptbrowne, line 74: Why not npm start ?

.dockerignore
```diff
@@ -1,4 +1,3 @@
 .git
 .next
-*Dockerfile*
-node_modules
\ No newline at end of file
+*Dockerfile*
```
ptbrowne, line 3: Why do we keep the node_modules as part of the context ? I thought that we should not keep it so that we do not upload node_modules to the docker image (which could be system specific), and so that we have a clean install inside the Docker build.

Dockerfile
bprusinowski: Might be the case, I adapted the official Next.js Dockerfile example from [here](https://github.com/vercel/next.js/blob/canary/examples/with-docker/Dockerfile). Let me know if you see some issues with this approach 👀

.dockerignore
bprusinowski: Good point, I'll investigate, I think I had some kind of error and modified this, this might in fact might not be needed 🤔

Dockerfile
ptbrowne: Here it seems we would miss the db:migrate script and potentially some environment variables (although I think I am not sure GLOBAL_AGENT and NO_PROXY are needed in case of visualize; I know they are in Elcom though), and the $PORT ? Thanks for the link to the doc, I know understand what is this app/server.js, I think we should have a docker:start command in npm scripts, so that we can handle additional scripts and make sure "start" and "docker:start" are aligned.

bprusinowski: Ah yes, I see, thanks for pointing this out! I thought we kind of have an equivalent of db:migrate script in the Dockerfile, when we `RUN yarn prisma generate`, but this might not be the case. I'll try to add the docker:start command and align both 👍

bprusinowski: @ptbrowne done in https://github.com/visualize-admin/visualization-tool/pull/1445/commits/3527df544d2ea0ee86722acef2237bec24ba96d6. Not sure if we need to be able to control $PORT as we also exposed port 3000 in the old Docker file? Or did I miss something? 😄

ptbrowne: Yep, not sure about that, let's try without port then ;)

ptbrowne (approved):
👏 Great !

PR #1450, opened 2024-04-12
Shared dimension querying

ptbrowne:
## Goal

Ability to query datasets by termset. I use termset as this is the word used in SPARQL, and I find that
Shared Dimension might be a little confusing, since in a single dimension, we can have several termsets.

In this PR, I add the querying of cubes by termsets. This is not yet directly done in the app UI, but
I added stories. In the future, we will be able to use this querying for

- Joined cubes
- Initial search page

The story is available here : https://visualization-tool-q9hub87q3-ixt1.vercel.app/storybook/index.html?path=/story/data-search--search


app/rdf/query-search.ts
```diff
@@ -342,4 +436,5 @@ const mkScoresQuery = (
       }
   `.GROUP().BY`?iri`.THEN.BY`?title`.THEN.BY`?status`.THEN.BY`?datePublished`
     .THEN.BY`?description`.THEN.BY`?publisher`.THEN.BY`?creatorIri`.THEN
-    .BY`?creatorLabel`.prologue`${pragmas}`.prologue`#pragma join.bind off`; // HOTFIX WRT Stardog v9.2.1 bug see https://control.vshn.net/tickets/sbar-1066
+    .BY`?creatorLabel`.prologue`${pragmas}`.prologue`#pragma join.bind off`;
```
bprusinowski, line 418: There's not a need to group by `termsetIri` and `termsetLabel` when `searchingSharedDimensions`? 🤔

app/rdf/query-search.ts
```diff
@@ -342,4 +436,5 @@ const mkScoresQuery = (
...
+}; // HOTFIX WRT Stardog v9.2.1 bug see https://control.vshn.net/tickets/sbar-1066
```
bprusinowski, line 419: Nit: maybe it's possible to reformat the chaining above to apply this comment just above the prologue`#pragma join.bind off`, as it's connected to this (now it looks like it applies to the whole function, as it's outside of it)?

codegen.yml
```diff
@@ -26,6 +26,7 @@ generates:
         DataCubeMetadata: "../domain/data#DataCubeMetadata"
         DataCubeObservations: "../domain/data#DataCubeObservations"
         DataCubePreview: "../domain/data#DataCubePreview"
+        Termset: "../domain/data#Termset"
```
bprusinowski, line 29: I think you also need to add this to `app/graphql/resolver-types.ts` down below? Not sure if the current behavior would cause some problems, but it looks like the scalars are defined in both places, I remember having some `any` types when adding them only to one file.

app/rdf/query-cube-termsets.ts
```diff
@@ -0,0 +1,39 @@
...
+PREFIX sh: <http://www.w3.org/ns/shacl#>
+
+SELECT DISTINCT ?termsetIri ?termsetLabel WHERE {
+    VALUES (?iri) {(<${iri}>)}
+
+    ?iri cube:observationConstraint/sh:property ?dimension .
+    ?dimension a cube:KeyDimension .
+    ?dimension sh:in/rdf:rest*/rdf:first ?value.
+    ?value schema:inDefinedTermSet ?termsetIri .
+    ?termsetIri a meta:SharedDimension .
```
ptbrowne, line 28: In the end, the query will evolve in the end so that we do not have to check each dimension values to find termsets. The dimension to termsets triplets will be directly materialized, so that querying can be faster.

bprusinowski (approved):
LGTM 💯 I like the approach of displaying the termsets directly as search result tags 👌 

One note after checking out the story: I think I'd expect the filters to be `AND` and not `OR`, as it the case now (looks like if I select `Cantons` termset, I get 5 results; when adding yearly temporal dimension, it goes up to 25).

I think when we'll merge the cubes, we need to apply the `AND` condition? Maybe I missed something, or it's like this only in the story? 😁

codegen.yml
ptbrowne: Thanks !

app/domain/data.ts
```diff
@@ -393,6 +393,10 @@ export type SearchCube = {
     iri: string;
     label: string;
   }[];
+  termsets: {
+    iri: string;
+    label: string;
+  }[];
```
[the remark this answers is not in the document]
ptbrowne: OK, will keep it in mind. Not sure if it is that useful though, I think in this case I prefer to be explicit and not rely on "inheritance".

app/rdf/query-search.ts
ptbrowne: No, actually since we concat them in the select clause, we should not group by.

PR #1451, opened 2024-04-12
feat: Do not unmount the chart when data is being loaded

### LINKED ISSUE #1126: Do not completely hide the chart when the data is loading
Right now, we completely hide the chart when the data is loading (both in editor mode and when using interactive filters in published mode). As we introduced animations, we are missing the movement between previous values and current values, as everything jumps when the data is loading. I wonder if it wouldn't make sense to indicate this state is some different way, e.g. by applying an loading overlay with reduced opacity, but not remove the chart from the DOM.

bprusinowski:
Closes #1126

Currently we unmount the whole chart component and use `<Loading />` when data is being fetched via the `ChartDataWrapper` component.

In my opinion, this is suboptimal user experience, as we lose track of the old chart state when the data is being loaded, and immediately are presented with new state, losing the sense of the difference between prior and current charts.

As we take advantage of smooth transitions between chart states in several chart types, it should result in less obtrusive behavior of the chart component, especially when manipulating chart filters.

For an initial implementation I applied a semi-transparent `LoadingOverlay` component, this might be changed after @KerstinFaye's review 😁

I also discovered a bug in the hook for data loading we use, that results in missing loading indication of any kind for the charts – this is also fixed by this PR.

### PR
1. Go to [this link](https://visualization-tool-git-feat-do-not-unmount-chart-wh-2a4991-ixt1.vercel.app/en/create/new?cube=https://environment.ld.admin.ch/foen/nfi/nfi_C-20/cube/2023-3&dataSource=Prod).
2. Play around with the filters.
3. ✅ See that there is a semi-transparent loading overlay and that the chart is not unmounted.

### PROD
1. Go to [this link](https://visualize.admin.ch/en/create/new?cube=https://environment.ld.admin.ch/foen/nfi/nfi_C-20/cube/2023-3&dataSource=Prod).
2. Play around with the filters.
3. ❌ See that there's no indication of the data being fetched (bug in the data fetching hooks, previously the whole chart was being unmounted).

### DIFF
--- a/app/charts/chart-data-wrapper.spec.tsx
+++ b/app/charts/chart-data-wrapper.spec.tsx
@@ -61,9 +61,9 @@ jest.mock("@/components/hint", () => ({
 
 describe("ChartDataWrapper", () => {
   it("should render the LoadingOverlay if prop is still being loaded", () => {
-    const Chart = jest.fn();
-    const LoadingOverlay = jest.fn();
-    render(
+    const Chart = () => <div>My chart</div>;
+    const LoadingOverlay = () => <div>Loading overlay</div>;
+    const root = render(
       <LoadingStateProvider>
         <ChartDataWrapper
           chartConfig={{ cubes: [] } as any as ChartConfig}
@@ -75,7 +75,11 @@ describe("ChartDataWrapper", () => {
         />
       </LoadingStateProvider>
     );
-    expect(Chart).toHaveBeenCalled();
-    expect(LoadingOverlay).toHaveBeenCalled();
+    expect(
+      root.getByText("My chart").innerHTML.includes("My chart")
+    ).toBeTruthy();
+    expect(
+      root.getByText("Loading overlay").innerHTML.includes("Loading overlay")
+    ).toBeTruthy();
   });
 });
--- a/app/charts/chart-data-wrapper.spec.tsx
+++ b/app/charts/chart-data-wrapper.spec.tsx
@@ -56,15 +56,18 @@ jest.mock("@/graphql/hooks", () => ({
 
 jest.mock("@/components/hint", () => ({
   Loading: jest.fn(() => null),
+  NoDataHint: jest.fn(() => null),
 }));
 
 describe("ChartDataWrapper", () => {
-  it("should not render the component if prop is still being loaded", () => {
+  it("should render the LoadingOverlay if prop is still being loaded", () => {
     const Chart = jest.fn();
+    const LoadingOverlay = jest.fn();
     render(
       <LoadingStateProvider>
         <ChartDataWrapper
           chartConfig={{ cubes: [] } as any as ChartConfig}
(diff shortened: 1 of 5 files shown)


app/graphql/hooks.ts
```diff
@@ -81,13 +85,7 @@ export const makeUseQuery =
...
     }, [queryKey, options.pause]);
 
-    const { data, ...rest } = result;
-    const res = {
-      ...rest,
-      data: result.queryKey === queryKey || keepPreviousData ? data : null,
-    };
-
-    return [res, executeQuery] as const;
+    return [result, executeQuery] as const;
```
ptbrowne, line 88: Looks like a tricky bug, is there an easy way to add a test for it, I see we already tests in hooks.spec.tsx, maybe it'd be easy, not sure.

app/charts/chart-data-wrapper.spec.tsx
```diff
@@ -56,22 +56,26 @@ jest.mock("@/graphql/hooks", () => ({
...
           Component={Chart}
           dataSource={{ type: "sparql", url: "url" }}
           observationQueryFilters={[]}
           fetching
         />
       </LoadingStateProvider>
     );
-    expect(Chart).not.toHaveBeenCalled();
+    expect(Chart).toHaveBeenCalled();
+    expect(LoadingOverlay).toHaveBeenCalled();
```
ptbrowne, line 79: In testing-library style (testing what the user sees), I think we would have text.
This way for example if the LoadingOverlay ends up being in a display: none container, the test would fail.

```
const Chart = () => <div>My chart</div>
const LoadingOverlay = () => <div>Loading overlay</div>
const root = render()
expect(root.getByText('My chart')).toBeVisible()
expect(root.getByText('Loading overlay')).toBeVisible()
```

ptbrowne (approved):
Nice, thanks for this change, it will greatly improve the experience 💯 

I find the LoadingOverlay text to be a little too dim, maybe Kerstin will find a way to make it a little more visible.

KerstinFaye:
Thanks @bprusinowski. It will very much improves user experience. That's really great. 

I would shorten the loading information that appears after some seconds. Maybe "Processing large data sets. Please wait.", or"Optimizing large data processing. Please wait."?

Is there any way of not dimming the text but dimming the chart underneath? It seams like the text has the same color as its background but with more opacity. Maybe we could improve the contrast by increasing opacity of the chart so the text is darker than the chart?

sosiology:
@KerstinFaye @bprusinowski just a small heads up. The text for the current loading state (see screenshot) was defined by Thomas Bettler. If we intend on changing it, i would suggest we quickly check in with him. Please also do not forget the translations. Thanks!


image

app/charts/chart-data-wrapper.spec.tsx
bprusinowski: Thanks, didn't think about it in this way :) I adjusted the test 👍

PR #1457, opened 2024-04-18
fix: Cached endpoint variables

bprusinowski:
This PR fixes `ENDPOINT` variable to correctly reference cached LINDAS endpoint, which should fix missing data source selection in preview deployments.

### Next steps
- [ ] Reach out to Abraxas once we have a new PROD deployment to update the ENDPOINT env variable on their side


ptbrowne (approved):
Looks good, did Zazuko they give their go for this ? I remember there were some challenges before ? But I also know that we use the cached endpoint for Elcom.

bprusinowski:
Yes, we mostly weren't sure how long the data is cached for, if I remember correctly it's around 2 mins 🤔 We got a green light from Thomas to change it 👍

PR #1461, opened 2024-04-19
feat: Improve resizing behavior

bprusinowski:
This PR improves resize behavior by disabling transitions during resizing. It also makes the Observer container a flex one, so the `Loading` indicator is centered when the chart is being loaded.

### PR
1. Go to [this link](https://visualization-tool-git-feat-improve-resizing-behavior-ixt1.vercel.app/en/create/new?cube=https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/14&dataSource=Int).
2. Disable browser fullscreen mode if applicable.
3. Grab the edge of the browser window and resize it.
4. ✅ See that the chart resizes immediately.

### TEST
1. Go to [this link](https://test.visualize.admin.ch/en/create/new?cube=https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/14&dataSource=Int).
2. Disable browser fullscreen mode if applicable.
3. Grab the edge of the browser window and resize it.
4. ❌ See that the chart resizing is laggy.

### DIFF
--- a/app/charts/shared/use-width.tsx
+++ b/app/charts/shared/use-width.tsx
@@ -1,45 +1,60 @@
-import { createContext, ReactNode, useContext } from "react";
+import React, { createContext, ReactNode, useContext } from "react";
 
+import { useTransitionStore } from "@/stores/transition";
 import { useResizeObserver } from "@/utils/use-resize-observer";
 
-export interface Margins {
+export type Margins = {
   top: number;
   right: number;
   bottom: number;
   left: number;
-}
-export type Width = number;
-export interface Bounds {
+};
+
+export type Bounds = {
   width: number;
   height: number;
   margins: Margins;
   chartWidth: number;
   chartHeight: number;
-}
+};
 
-const INITIAL_WIDTH: Width = 1;
+const INITIAL_WIDTH = 1;
 
 export const Observer = ({ children }: { children: ReactNode }) => {
-  const [resizeRef, width] = useResizeObserver<HTMLDivElement>();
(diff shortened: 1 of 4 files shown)


app/charts/shared/use-width.tsx
```diff
@@ -1,46 +1,60 @@
...
+  React.useEffect(() => {
+    setEnableTransition(false);
+    const timeout = setTimeout(() => {
+      setEnableTransition(true);
+    }, 500);
+
+    return () => {
+      clearTimeout(timeout);
+    };
+  }, [width, setEnableTransition]);
```
ptbrowne, line 37: Super nit: I think that maybe the effect is firing a bit too much ? Alternative way.


```
const usePrevious = <T,>(value: T, duration: number): T => {
  const [previousValue, setPreviousValue] = useState<T>(value);

  useEffect(() => {
    const timeoutId = setTimeout(() => {
      setPreviousValue(value)
    }, duration);

    return () => {
      clearTimeout(timeoutId);
    };
  }, [value, duration]);

  return previousValue;
};


const prev = usePrevious(width, 500)
const isResizing = prev !== width
useEffect(() => setEnableTransition(isResizing), [isResizing])
```

ptbrowne (approved):
👍

app/charts/shared/use-width.tsx
bprusinowski: Really like the idea, done, thanks 💯

PR #1464, opened 2024-04-19
perf: Fire DataCubesObservations query as soon as possible

### LINKED ISSUE #1423: Load optimized possible filters upfront // apply available information as early as possible
From Gitlab Issue: Please check [there](https://gitlab.ldbar.ch/bafu/visualize/-/issues/650) before tackling the issues incase Thomas adds additional feedback or input

**Context**

Now that `PossibleFilters` were vastly optimized, this offers the possiblity to rely on those results to trigger the later queries more specifically and improve the query order and result processing

**Overall Goal**

Show the chart as early as possible and load the active UI as fast as possible.

**User Story**

As a User I'd like to have the UI loaded as early as possible (based on the possibilites given by the already improved query)

**Details**

<details>

Stored chart config
image

```
#Observations query expected to be fired earlier - based on the given chart config
{
    "query": "query DataCubeObservations($sourceType: String!, $sourceUrl: String!, $locale: String!, $cubeFilter: DataCubeObservationFilter!) {\n  dataCubeObservations(\n    sourceType: $sourceType\n    sourceUrl: $sourceUrl\n    locale: $locale\n    cubeFilter: $cubeFilter\n  )\n}\n",
    "operationName": "DataCubeObservations",
    "variables": {
        "locale": "de",
        "sourceType": "sparql",
        "cubeFilter": {
            "iri": "https://environment.ld.admin.ch/foen/nfi/nfi_C-90/cube/2023-3",
            "componentIris": [
                "https://environment.ld.admin.ch/foen/nfi/Topic/116",
                "https://environment.ld.admin.ch/foen/nfi/classificationUnit",
                "https://environment.ld.admin.ch/foen/nfi/evaluationType",
                "https://environment.ld.admin.ch/foen/nfi/grid",
                "https://environment.ld.admin.ch/foen/nfi/inventory",
                "https://environment.ld.admin.ch/foen/nfi/unitOfEvaluation",
                "https://environment.ld.admin.ch/foen/nfi/unitOfReference"
            ],
            "filters": {
                "https://environment.ld.admin.ch/foen/nfi/unitOfReference": {
                    "type": "single",
                    "value": "https://ld.admin.ch/country/CHE"
                },
                "https://environment.ld.admin.ch/foen/nfi/classificationUnit": {
                    "type": "single",
                    "value": "https://environment.ld.admin.ch/foen/nfi/ClassificationUnit/Total"
                },
                "https://environment.ld.admin.ch/foen/nfi/unitOfEvaluation": {
                    "type": "single",
                    "value": "https://environment.ld.admin.ch/foen/nfi/UnitOfEvaluation/2382"
                },
                "https://environment.ld.admin.ch/foen/nfi/grid": {
                    "type": "single",
                    "value": "https://environment.ld.admin.ch/foen/nfi/Grid/410"
                },
                "https://environment.ld.admin.ch/foen/nfi/evaluationType": {
                    "type": "single",
                    "value": "https://environment.ld.admin.ch/foen/nfi/EvaluationType/1"
                }
            }
        }
    }
}
```

</details>

**DoD**
- [x] When starting creating a new chart from a cube, possible filters are fetched upfront, to avoid situation where the initial, random filter combination doesn't make sense and triggers unnecessary observations & dimension values queries
- [ ] When loading a published chart, CubeObservations query is fired as early as possible

bprusinowski:
Closes #1423

This PR removes a need to pass dimensions to get query filters, as it looks like this behavior is not needed. It was used to filter out interactive data filters that are a part of dimensions, which should always be true (except when a dimension was completely removed from a cube, which would most probably break the chart anyway).

With this change we fire `DataCubesObservations` query in parallel to `DataCubesComponents` fetching, not waiting for it to finish first.

### PR
1. Go to [this link](https://visualization-tool-git-perf-fire-observations-query-asap-ixt1.vercel.app/en/v/ENrCApdv5Cbo?dataSource=Prod).
2. Open Developer Tools (e.g. [see how to do so on Chrome](https://developer.chrome.com/docs/devtools/open)).
3. Switch to the Network tab.
4. Reload the page.
5. ✅ See that the `graphql` queries were fired (almost) at the same time.

Screenshot 2024-04-19 at 13 30 54

### TEST
1. Go to [this link](https://test.visualize.admin.ch/en/v/Zq6ql3-tONlE?dataSource=Prod).
2. Open Developer Tools (e.g. [see how to do so on Chrome](https://developer.chrome.com/docs/devtools/open)).
3. Switch to the Network tab.
4. Reload the page.
5. ❌ See that some graphql queries were fired sequentially (waited for other queries to finish first).

Screenshot 2024-04-19 at 13 31 11

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ You can also check the [release page](https://github.com/visualize-admin/visuali
   - Filters are now correctly applied when adding a new chart
 - Style
   - Introduced smaller UI improvements in the options panel
+- Performance
+  - DataCubesObservations query no longer waits for DataCubesComponents query to finish in order to start executing
 
 # [4.1.0] - 2024-04-10
 
--- a/app/browse/datatable.tsx
+++ b/app/browse/datatable.tsx
@@ -269,17 +269,13 @@ export const DataSetTable = ({
       ...componentsData.dataCubesComponents.measures,
     ]);
   }, [componentsData?.dataCubesComponents]);
-  const queryFilters = useQueryFilters({
-    chartConfig,
-    dimensions: componentsData?.dataCubesComponents?.dimensions,
-    componentIris,
-  });
+  const queryFilters = useQueryFilters({ chartConfig, componentIris });
   const [{ data: observationsData }] = useDataCubesObservationsQuery({
     variables: {
       ...commonQueryVariables,
-      cubeFilters: queryFilters ?? [],
+      cubeFilters: queryFilters,
     },
-    pause: fetchingComponents || !queryFilters,
+    pause: fetchingComponents,
   });
 
   return metadataData?.dataCubesMetadata &&
--- a/app/browse/datatable.tsx
+++ b/app/browse/datatable.tsx
@@ -272,7 +272,6 @@ export const DataSetTable = ({
   const queryFilters = useQueryFilters({
     chartConfig,
     dimensions: componentsData?.dataCubesComponents?.dimensions,
-    measures: componentsData?.dataCubesComponents?.measures,
     componentIris,
   });
   const [{ data: observationsData }] = useDataCubesObservationsQuery({
--- a/app/charts/area/chart-area.tsx
+++ b/app/charts/area/chart-area.tsx
@@ -28,7 +28,7 @@ export const ChartAreasVisualization = ({
 }: {
   dataSource: DataSource;
   chartConfig: AreaConfig;
-  queryFilters?: DataCubeObservationFilter[];
+  queryFilters: DataCubeObservationFilter[];
   componentIris: string[] | undefined;
 }) => {
   return (
--- a/app/charts/area/chart-area.tsx
+++ b/app/charts/area/chart-area.tsx
@@ -15,31 +15,14 @@ import { Ruler } from "@/charts/shared/interaction/ruler";
 import { Tooltip } from "@/charts/shared/interaction/tooltip";
 import { LegendColor } from "@/charts/shared/legend-color";
 import { InteractionHorizontal } from "@/charts/shared/overlay-horizontal";
-import { AreaConfig, DataSource } from "@/config-types";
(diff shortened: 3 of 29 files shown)


ptbrowne:
I see that you added this code in https://github.com/visualize-admin/visualization-tool/pull/1273/files, so I think you added that to only apply the filters to the correct cubes. So I think you still need to know which cubes has which filters.

Maybe we should add the cube iri to each filter, so that this would be instant ? But then we'd have the problem of versioned/unversioned iri ? Maybe hear we need a query that returns only dimension iri and measure iri for a cube (without getting unit, values, description etcc...), that would not prevent the waterfall query though.

bprusinowski:
Thanks for catching this @ptbrowne. I dug a little deeper and you're right that we still need to narrow down interactive data filters to a given cube – interestingly, this is not the case with current logic on the `main` branch (see that `queryFilters` contain mixed values from two cubes).

Screenshot 2024-04-19 at 15 12 07

This is due to the fact that we simply pass `dimensions` which come from both cubes, and the logic in its current form would in fact always return all dimension unless one of them disappeared from a cube.

But to answer the need of querying dimensions to get this information, I think that's not the case, as interactive data filters are only enabled for filters from the left panel and we keep their IRIs in the chart config – so we can already narrow them down correctly per cube without filters (see newest commit).

There still could be a problem if two cubes contain dimension with the same IRI, for it I think we should refactor interactive filters store to be cube-based. But this PR should already improve performance and fix wrong assignment of filters in most of the cases compared to actual implementation 🤞 

Screenshot 2024-04-19 at 15 24 36

ptbrowne:
Cool to have seen this request waterfall 🚀 🚀 
Code looks good, thanks again for walking me through it :)

ptbrowne (approved):

Rdataflow:
@bprusinowski well done :rocket: :rocket: :rocket: a clear speed gain for published charts

bprusinowski:
Thanks for catching this issue @Rdataflow 💯

PR #1473, opened 2024-04-24
perf: Upgrade to latest cube iri once

### LINKED ISSUE #1472: Upgrade cube version initially and once only on client-side
**Situation**
- Each GQL query re-initiates the version upgrade on it's own, even after the version information is already known at the client
- Version upgrade is not streamlined across all the different GQL queries (i.e. some use LightCube, others won't)

**Motivation**
- Firing less queries to the SPARQL endpoint helps to improve the general performance. Therefore omitting double queries (i.e. for version upgrade) will help to improve.

**DoD**

- [ ] The cube version information initially queried by the client is reused - suceeding GQL queries rely on that dynamically queried information, therefore no need to re-query for version upgrades.
- [ ] The version upgrade functionality is streamlined among all the GQL queries. The recent improved query pattern from getLatestCubeIriQuery is reused therefore.

bprusinowski:
Closes #1472

This goal of this PR is to make the server-side not aware of a "latest cube" concept and to always just pass `cubeIri`, without the `latest` property. Cube upgrade will happen once on the client side when initializing the chart (edit mode) or fetching the chart config (embed and published modes).

The rationale for it is that there is no need to constantly ask for latest cube version in every GQL query. This will improve performance and reduce the number of queries we send to LINDAS.

This change should also make it easier to control when an upgrade should happen by having centralized place to define that behavior, instead of having to pass the latest property in every GQL query (`upgradeConfiguratorState` that could be called conditionally if at some we introduce an ability to pin cube version).

## Implementation logic
I introduced a new, `publishIri` property to chart config cube, which stores cube IRI at publish time. I decided to leave the current `iri` property to make it much easier to handle the change throughout the application without additional custom JS logic to keep the mapping somewhere and access upgraded iri in every query – because of this, majority of the code stays the same. We just keep the currently upgraded cube iri in the `iri` property.

This should make it easy to control the upgrade logic – if the cube is not meant to be upgraded, we assign `publishIri` to `iri`; if it's meant to be upgraded, we keep upgraded iri in `iri`, without affecting rest of the application.

### Questions
- Does the current implementation look ok @ptbrowne?
- Should we maybe override `iri` with `publishIri` when updating the cube, at technically it's a "runtime" property?
- Maybe it would make more sense to create something like `runtimeConfig` that would be a part of configurator state that is removed at publish time?


Rdataflow:
@bprusinowski unfortunately the upgrade of published charts does not to happen anymore. This needs to be fixed.

repro case:
1. open https://visualization-tool-git-perf-upgrade-cube-version-once-ixt1.vercel.app/v/vyJ1Kt0I9Wbb?dataSource=Test
2. see version 5 is loaded while version 6 is published (nb: hence dateModified is from the past)

bprusinowski:
Thanks for catching a problem with published charts @Rdataflow, should be fixed now 😄

app/utils/chart-config/versioning.ts
```diff
@@ -854,9 +854,12 @@ export const chartConfigMigrations: Migration[] = [
   },
 ];
 
-export const migrateChartConfig = makeMigrate(chartConfigMigrations, {
+export const migrateChartConfig = makeMigrate<ChartConfig>(
```
ptbrowne, line 887: 👍

app/configurator/configurator-state.tsx
```diff
@@ -1975,16 +1975,13 @@ export const isPublished = (
 export const hasChartConfigs = (
   s: ConfiguratorState
 ): s is ConfiguratorStateWithChartConfigs => {
-  return (
-    isConfiguring(s) || isLayouting(s) || isPublishing(s) || isPublished(s)
-  );
+  return "chartConfigs" in s;
```
ptbrowne, line 1978: 👍

ptbrowne (approved):
Thanks a lot @bprusinowski 💯

PR #1474, opened 2024-04-24
Beta tag for merging cubes

ptbrowne:
- fix: No shadow on tag if not clickable
- fix: Missing translations
- feat: Add beta tag

### DIFF
--- a/app/configurator/components/add-dataset-dialog.tsx
+++ b/app/configurator/components/add-dataset-dialog.tsx
@@ -49,6 +49,7 @@ import {
   isConfiguring,
   useConfiguratorState,
 } from "@/configurator";
+import { BetaTag } from "@/configurator/components/beta-tag";
 import { Termset, isTemporalDimensionWithTimeUnit } from "@/domain/data";
 import { truthy } from "@/domain/types";
 import {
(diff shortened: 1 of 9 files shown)


app/configurator/components/add-dataset-dialog.tsx
```diff
@@ -69,6 +69,8 @@ import SvgIcSearch from "@/icons/components/IcSearch";
 import { useLocale } from "@/locales/use-locale";
 import { exhaustiveCheck } from "@/utils/exhaustiveCheck";
 
+import { BetaTag } from "./beta-tag";
```
bprusinowski, line 72: Super-nit: I like the consistency of having the same "style" for imports, maybe we could somehow enforce aliased ones? I think I changed this in my VS Code settings, but maybe there's some configuration option in `settings.json`. Not for this PR, just a thing that I noticed :)

bprusinowski (approved):
👌

app/configurator/components/add-dataset-dialog.tsx
ptbrowne: I also like consistency :) I think when I auto import, I automatically get aliases, but here it's because I extracted the component, and when extracting the imports gets aliased.

bprusinowski: Thank you 🙇‍♂️

sosiology:
Very nice @ptbrowne the tooltip with the info is a nice addition. Could we get this on INT soonish (latest End of Week) as the feature is already on INT?

Issue #1470, opened 2024-04-24
Rework Components query to query multiple dimensions at once


sosiology:
**Situation**        

- The Components GQL query queries multiple dimension individually as of now.
- The Elements with duplicate predicates (i.e. schema:identifier) figure twice in the result list.

**Motivation**
- The SPARQL endpoint is generally better performing upon recieving less queries - but optimized ones.

**DoD**     

- [ ]  The Components Query makes one SPARQL query over all dimensions at once (instead of multiple)
- [ ] The resultset has no duplicates regardless of duplicate predicates

bprusinowski:
Hi @Rdataflow, I merged the #1487 a while ago to be able to compare the performance after the latest query improvements on the same infrastructure (INT data source on Visualize TEST and Visualize INT, with server-side cache disabled).

After a bit of testing, these are the rough results (times for the longest components query from the debug panel):
- BLW cubes: ~20s on [TEST](https://test.visualize.admin.ch/en/create/ZyOQW6uExUux?dataSource=Int&flag__server-side-cache.disable=true) and ~5s on [INT](https://int.visualize.admin.ch/en/create/7BCTGS3YSoFK?dataSource=Int&flag__server-side-cache.disable=true),
- Traffic noise pollution: ~1s on [TEST](https://test.visualize.admin.ch/en/create/new?cube=https://environment.ld.admin.ch/foen/ubd0037/9&dataSource=Int) and ~0.5s to [INT](https://int.visualize.admin.ch/en/create/new?cube=https://environment.ld.admin.ch/foen/ubd0037/9&dataSource=Int),
- Stapfer: ~10s on [TEST](https://test.visualize.admin.ch/en/create/new?cube=https://culture.ld.admin.ch/sfa/stapfer-enquete-1799/3&dataSource=Int) and ~25s on [INT](https://int.visualize.admin.ch/en/create/new?cube=https://culture.ld.admin.ch/sfa/stapfer-enquete-1799/3&dataSource=Int),
- Elcom: ~20s on [TEST](https://test.visualize.admin.ch/en/create/new?cube=https://energy.ld.admin.ch/elcom/electricityprice&dataSource=Int) and ~40s on [INT](https://int.visualize.admin.ch/en/create/new?cube=https://energy.ld.admin.ch/elcom/electricityprice&dataSource=Int).

[...]

PR #1483, opened 2024-04-30
chore: GraphQL cleanup

bprusinowski:
This PR is a preparation for #1470 along with a cleanup of old GraphQL stuff that we no longer use 🧹


ptbrowne (approved):
LGTM 👏

PR #1485, opened 2024-04-30
drop legacy hasHierarchy

Rdataflow:
see also https://github.com/visualize-admin/visualization-tool/commit/ac4db2ebf39ef57b7afc4cefd0b3f87831f01ee5

WRT https://github.com/zazuko/cube-creator/issues/1243

cc @ptbrowne

### DIFF
--- a/app/rdf/queries.ts
+++ b/app/rdf/queries.ts
@@ -592,8 +592,7 @@ const unversionServerFilters = async (
 
 export const hasHierarchy = (dim: CubeDimension) => {
   return (
-    dim.out(ns.cubeMeta.inHierarchy).values.length > 0 ||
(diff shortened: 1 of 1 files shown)


bprusinowski:
Hi @Rdataflow, thanks for opening a PR! I wonder if [this list](https://s.zazuko.com/2HP3KkG) should be empty before merging it? 👀

Rdataflow:
@bprusinowski thank you for double checking the data - I also investigated this with a little difference in the query https://s.zazuko.com/3ju5W8V - which prooves none of the dimensions feature solely hasHierarchy - we can rely inHierarchy or both are present 👍

bprusinowski:
I see, thanks for the enhanced query 🎉

bprusinowski (approved):

PR #1487, opened 2024-05-02
refactor: Combined query to fetch dimension values

### LINKED ISSUE #1470: Rework Components query to query multiple dimensions at once

bprusinowski:
Closes #1470

This PR is an exploration of the feasibility to fetch values for multiple dimensions in a single SPARQL query.

### Constrains

- We need to be able to filter each dimension individually, due to cascading filters behavior. Could be achieved with ```FILTER(IF(?dimensionIri = <A>, ?dimensionIri = <a> && ?dimensionIri = <a>, ?dimensionIri))```
- We need to be able to unversion dimension values per individual dimension (can't unversion values outside of individual SELECT queries, as sometimes `schema:sameAs` is not used to indicate unversioned values, this is only the case when a dimension is versioned).

we need to combine individual queries into a big one (at least that's my current assumption).

### DIFF
--- a/app/rdf/queries.ts
+++ b/app/rdf/queries.ts
@@ -196,20 +196,63 @@ const getCubeDimensionsValues = async (
     cache: LRUCache | undefined;
   }
 ) => {
+  const { minMaxDimensions, regularDimensions } = resolvedDimensions.reduce<{
+    minMaxDimensions: ResolvedDimension[];
+    regularDimensions: ResolvedDimension[];
+  }>(
+    (acc, dimension) => {
+      if (shouldLoadMinMaxValues(dimension)) {
+        acc.minMaxDimensions.push(dimension);
+      } else {
+        acc.regularDimensions.push(dimension);
+      }
+
+      return acc;
+    },
+    { minMaxDimensions: [], regularDimensions: [] }
+  );
+
+  const result = await Promise.all([
+    getMinMaxDimensionsValues(minMaxDimensions, {
+      sparqlClient,
+      cache,
+    }),
+    getRegularDimensionsValues(regularDimensions, {
+      sparqlClient,
+      filters,
+      cache,
+    }),
+  ]);
+
+  return result
+    .flat()
+    .sort(
+      (a, b) =>
+        resolvedDimensions.indexOf(a.resolvedDimension) -
+        resolvedDimensions.indexOf(b.resolvedDimension)
+    )
+    .map(({ values }) => values);
+};
+
+const getMinMaxDimensionsValues = async (
+  resolvedDimensions: ResolvedDimension[],
+  {
+    sparqlClient,
+    cache,
+  }: {
+    sparqlClient: ParsingClient;
+    cache: LRUCache | undefined;
+  }
+) => {
   return await Promise.all(
     resolvedDimensions.map(async (resolvedDimension) => {
-      if (shouldLoadMinMaxValues(resolvedDimension)) {
-        return await getMinMaxDimensionValues(resolvedDimension, {
+      return {
+        resolvedDimension,
+        values: await getMinMaxDimensionValues(resolvedDimension, {
           sparqlClient,
           cache,
-        });
-      } else {
-        return await getRegularDimensionValues(resolvedDimension, {
-          sparqlClient,
-          filters,
-          cache,
-        });
-      }
+        }),
+      };
     })
   );
 };
@@ -287,6 +330,32 @@ const getMinMaxDimensionValues = async (
   }
 
   return [];
+};
+
+const getRegularDimensionsValues = async (
+  resolvedDimensions: ResolvedDimension[],
+  {
+    sparqlClient,
+    filters,
+    cache,
+  }: {
+    sparqlClient: ParsingClient;
+    filters?: Filters;
+    cache: LRUCache | undefined;
+  }
+) => {
+  return await Promise.all(
+    resolvedDimensions.map(async (resolvedDimension) => {
+      return {
+        resolvedDimension,
+        values: await getRegularDimensionValues(resolvedDimension, {
+          sparqlClient,
+          filters,
+          cache,
+        }),
+      };
+    })
+  );
 };
 
 const getRegularDimensionValues = async (
--- a/app/rdf/queries.ts
+++ b/app/rdf/queries.ts
@@ -9,7 +9,6 @@ import { LRUCache } from "typescript-lru-cache";
 
 import { FilterValueMulti, Filters } from "@/config-types";
 import {
-  DimensionValue,
   Observation,
   ObservationValue,
   parseObservationValue,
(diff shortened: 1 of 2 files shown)


Rdataflow:
@bprusinowski can you share an example of a less performing query? thank you

bprusinowski:
@Rdataflow sure, this is a [query](https://s.zazuko.com/3bdGoDd) to fetch dimension values for every dimension in the Bathing water quality cube (takes ~12s). Compare this to queries fired on PROD: https://visualize.admin.ch/en/create/lmbY5klvYJAm?dataSource=Prod&flag__debug=true&flag__server-side-cache.disable=true (takes 1.4s for 5 queries, one per each dimension, fired in parallel).

The above example is for a non-filtered query (fetch all dimension values), but we also need to be able to filter and unversion each dimension differently (see PR description). This is why I think we need to combine individual queries like this – let me know if it's enough information or if you see some other direction to try.

Rdataflow:
@bprusinowski it seems those `OPTIONAL`s for `schema:version` have a negative impact. You may try to further parallelize using a query pattern of 
```sparql
{ 
    # versioned case
    ...
    ?dimension schema:version ?version 
    ... 
} UNION { 
    # nonversioned case
    ... 
    FILTER NOT EXISTS { ?dimension schema:version ?version } 
    ... 
}
```

 i.e. like https://s.zazuko.com/HenZ6K

HTH for self guiding the next steps 👍

bprusinowski:
Hey @Rdataflow, I optimized the query based on your suggestion and fixed some issues, it seems to fetch the correct values now.

However it looks like it's less performing than the current approach of parallel queries, both for smaller and bigger cubes. You can see some examples for Photovoltaikanlagen ([TEST](https://test.visualize.admin.ch/en/create/YsL_13MC9adM?dataSource=Int&flag__server-side-cache.disable=true), [PR](https://visualization-tool-git-refactor-combined-components-450f24-ixt1.vercel.app/en/create/w99DtM3Oo4Ga?dataSource=Int&flag__server-side-cache.disable=true)) and NFI: Change ([TEST](https://test.visualize.admin.ch/en/create/a9PjvTNJQXxP?dataSource=Int&flag__server-side-cache.disable=true), [PR](https://visualization-tool-git-refactor-combined-components-450f24-ixt1.vercel.app/en/create/M0IGCxKjSo2Q?dataSource=Int&flag__server-side-cache.disable=true)) cubes (both using INT data source to not rely on cached endpoint and with disabled server-side cache).

It looks like the new approach is ~2x slower than what we currently have on TEST. I will take a deeper look to see if there's something obvious to optimize, but it would be great if you could also take a look in case you have a bit of time.

Let me know what you about the whole direction of combining the queries 👀

Rdataflow:
@bprusinowski unfortunately those /create/ links won't endure... do you have some permalinks maybe?

Rdataflow:
i.e. those /create/new?from=<published> would work IIUC

bprusinowski:
@Rdataflow of course – I updated the links, should be ok now (I blame it on Monday morning 🤦 😅)

bprusinowski:
@Rdataflow I did some tests to fire the query "for the first time", to avoid some apparent caching on LINDAS side (did this by modifying the query to e.g. remove retrieval of color, so the query looks like _new_); it still looks like `#pragma` improves the situation a bit (1.8s for [pragma](https://s.zazuko.com/2rwGBKS) vs 2.2s [without](https://s.zazuko.com/26Hr1fZ)) for the Traffic noise pollution cube. The timings change if I "comment out" other properties, but the delta seems to always be in favor of `#pragma`...

Rdataflow:
@bprusinowski the query pattern still would profit of some minor optimization steps...
the 3rd and 4th UNION block of each dimension may be optimized using a inner SELECT to prevent `?obs ?dimensionIri ?versionedValue` from evaluation in case of empty due to `FILTER NOT EXISTS { ... }` returning empty set

this might look like  i.e.
```sparql
# 3rd UNION block on a dimension
  {
    SELECT DISTINCT ?dimensionIri ?versionedValue ?unversionedValue
    WHERE {
      {
        SELECT ?observation
        WHERE {
          VALUES ?dimensionIri { <https://environment.ld.admin.ch/foen/nfi/inventory> }
          <https://environment.ld.admin.ch/foen/nfi/nfi_T-changes/cube/2024-1> cube:observationConstraint/sh:property ?dimension .
          ?dimension sh:path ?dimensionIri .
          ?dimension schema:version ?version .
          FILTER NOT EXISTS {
            ?dimension sh:in ?in .
          }
          <https://environment.ld.admin.ch/foen/nfi/nfi_T-changes/cube/2024-1> cube:observationSet/cube:observation ?observation .
        }
      }
      VALUES ?dimensionIri { <https://environment.ld.admin.ch/foen/nfi/inventory> }
      ?observation ?dimensionIri ?versionedValue .
      ?versionedValue schema:sameAs ?unversionedValue .
    }
  }
  UNION
  {
# 4th UNION block on a dimension
    SELECT DISTINCT ?dimensionIri ?versionedValue ?unversionedValue
    WHERE {
      {
        SELECT ?observation
        WHERE {
          VALUES ?dimensionIri { <https://environment.ld.admin.ch/foen/nfi/inventory> }
          <https://environment.ld.admin.ch/foen/nfi/nfi_T-changes/cube/2024-1> cube:observationConstraint/sh:property ?dimension .
          ?dimension sh:path ?dimensionIri .
          FILTER NOT EXISTS {
            ?dimension sh:in ?in .
          }
          FILTER NOT EXISTS {
            ?dimension schema:version ?version .
          }
          <https://environment.ld.admin.ch/foen/nfi/nfi_T-changes/cube/2024-1> cube:observationSet/cube:observation ?observation .
        }
      }
      ?observation ?dimensionIri ?versionedValue .
      BIND(?versionedValue AS ?unversionedValue)
    }
  }
```

then as the innermost `SELECT ?observation` becomes emtpy (true for many dimensions) the UNION blocks are faster now 😄

bprusinowski:
Thanks again @Rdataflow for optimizing the query 💯

Unfortunately it looks that it's still significantly less performing that the ones we have on TEST / INT / PROD. See [this combined query](https://s.zazuko.com/2Pz9ktb) that takes 12-13s – the same cube on TEST takes ~7-8s to load values for every dimension when fired separately.

I think we might reach out to Zazuko, seeing that the approach we currently try doesn't seem to improve things – does it sound good? Maybe I miss some additional context, but knowing that we'll use a cached endpoint that will already offload a load of computing power from Stardog, I am not sure if it's worth it to sacrifice 50% of performance just to send a smaller number of queries.

Let me know what you think @Rdataflow :)

cc @sosiology @adintegra

Rdataflow:
@bprusinowski the proposed query obviously misses to constrain the dimensionIri to the relevant dimension only - therefore it suffers heavily degraded performance

https://s.zazuko.com/2xuxSbq

nb: you may not rely on TEST w/ v10.1 ATM due to VSHN SBAR-1122 ... at least w/o `#pragma join.hash off`

cc @sosiology @adintegra

PR #1488, opened 2024-05-02
feat: Add Save button to annotations panel

### LINKED ISSUE #1427: Add "Save" button to the title and description overlay
Implement "Save" button as documented in the design refinements from Sprint 5.6. 

Image

bprusinowski:
Closes #1427

This PR adds a "Save" button to the annotations panel. As we save the changes dynamically anyway, it acts as a "close panel" button.

### DIFF
--- a/app/configurator/components/annotation-options.tsx
+++ b/app/configurator/components/annotation-options.tsx
@@ -1,4 +1,3 @@
-import { Trans } from "@lingui/macro";
 import { Box, Button, Typography } from "@mui/material";
 import { useEffect, useRef } from "react";
 
@@ -108,9 +107,7 @@ const AnnotationOptions = (props: AnnotationOptionsProps) => {
             onClick={handleClosePanel}
             sx={{ alignSelf: "flex-end", mt: 3, px: 5 }}
           >
-            <Typography component="span">
-              <Trans id="save">Save</Trans>
-            </Typography>
+            <Typography component="span">Ok</Typography>
           </Button>
         </ControlSectionContent>
       </ControlSection>
--- a/app/configurator/components/annotation-options.tsx
+++ b/app/configurator/components/annotation-options.tsx
@@ -1,4 +1,5 @@
-import { Box } from "@mui/material";
+import { Trans } from "@lingui/macro";
+import { Box, Button, Typography } from "@mui/material";
 import { useEffect, useRef } from "react";
 
 import { Meta, getChartConfig } from "@/config-types";
@@ -17,11 +18,11 @@ import { MetaInputField } from "@/configurator/components/field";
 import { getFieldLabel } from "@/configurator/components/field-i18n";
 import { locales } from "@/locales/locales";
 import { useLocale } from "@/locales/use-locale";
+import useEvent from "@/utils/use-event";
 
 export const ChartAnnotationsSelector = () => {
   const [state] = useConfiguratorState(isConfiguring);
   const chartConfig = getChartConfig(state);
-
   return (
     <AnnotationOptions
       type="chart"
@@ -50,11 +51,21 @@ type AnnotationOptionsProps = {
 
 const AnnotationOptions = (props: AnnotationOptionsProps) => {
   const { type, activeField, meta } = props;
+  const [_, dispatch] = useConfiguratorState();
   const locale = useLocale();
   // Reorder locales so the input field for
   // the current locale is on top
   const orderedLocales = [locale, ...locales.filter((l) => l !== locale)];
   const panelRef = useRef<HTMLDivElement>(null);
+  const handleClosePanel = useEvent(() => {
+    dispatch({
+      type:
+        type === "chart"
+          ? "CHART_ACTIVE_FIELD_CHANGED"
+          : "LAYOUT_ACTIVE_FIELD_CHANGED",
+      value: undefined,
+    });
+  });
 
   useEffect(() => {
     if (panelRef?.current) {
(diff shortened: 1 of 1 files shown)


bprusinowski:
@KerstinFaye I see that in the current design file we also have a `Cancel` button, contrary to what's inside the #1427 issue 😱 

As we have a "live mode editing" across the whole application, where changes are visible on a chart as soon as we make them (except for some data loading related actions, like updating multi-filters where we need to confirm the selection to avoid constant data changes and loading), I wonder if we should change this paradigm for annotations – because if I understand correctly, in the [current design](https://www.figma.com/file/Hn9wvuEYLUmdJtg4KwZYUx/visualize.admin.ch---design?type=design&node-id=3718-229003&mode=design&t=YISUZKohm2MLxOrX-4), user should be able to update title and description without immediate feedback on a chart, but only applied after clicking "Save", and discarded when clicking "Cancel")?

This is a personal preference, but I find it more user-friendly to have this live feedback as we do currently, just wanted to check with you if we should change this (and if so, should we also apply it in other parts of the application)? If you think it's a good idea to change this, maybe then we could also have a small tag above, like "Pending changes", so we know if we actually made a change? I think it might be a bit confusing now, as we can save the changes immediately after entering the panel, without actually updating the title or description yet 👀 

Let me know what you think about this :)

app/configurator/components/annotation-options.tsx
```diff
@@ -92,6 +103,14 @@ const AnnotationOptions = (props: AnnotationOptionsProps) => {
               />
             </Box>
           ))}
+          <Button
+            onClick={handleClosePanel}
+            sx={{ alignSelf: "flex-end", minHeight: 32, mt: 3, px: 5 }}
```
ptbrowne, line 108: Nit: Why do we have to override minHeight and px, this should be handled via the button right ?

app/configurator/components/annotation-options.tsx
```diff
@@ -92,6 +103,14 @@ const AnnotationOptions = (props: AnnotationOptionsProps) => {
...
+          >
+            <Typography component="span">
+              <Trans id="save"> Save</Trans>
```
ptbrowne, line 111:
```suggestion
              <Trans id="save">Save</Trans>
```

ptbrowne:
Code looks good but I am wondering about the behavior here. Here I wonder, there was no modification to how changing a field changes the chart, so Bartosz if I am not mistaken, here in this PR, even if the user does not click "Save", it will have been saved already right ?
I completely agree with Bartosz, the addition of a Save button feels like a change of paradigm. What is exactly the intention there ?

app/configurator/components/annotation-options.tsx
bprusinowski: Thanks for catching, for height I could use size="small" 👍 For px, it looks different for different buttons in the design currently, see e.g. [this button](https://www.figma.com/file/Hn9wvuEYLUmdJtg4KwZYUx/visualize.admin.ch---design?type=design&node-id=3718-229434&mode=design&t=MBXaeZBQkyiFi3RM-4) and [this button](https://www.figma.com/file/Hn9wvuEYLUmdJtg4KwZYUx/visualize.admin.ch---design?type=design&node-id=3718-229077&mode=design&t=MBXaeZBQkyiFi3RM-4) – in fact maybe this should be minWidth applied in the theme, but I didn't want to do a full review of design to find a rule in this PR, maybe we could check with @KerstinFaye?

bprusinowski:
@ptbrowne yes, it's a "fake" Save button now, as I didn't want to completely change the behavior before discussing this 👍

KerstinFaye:
@bprusinowski Thanks a lot for the feedback. I agree, it's definitely better to have the live feedback. I reviewed the implemented version and I think the solution to have the save button but still the live feedback is best.

KerstinFaye:
@ptbrowne I think this ideas is even based on your feedback that you gave ones in a document that you did for UI improvements. I thought it was a good idea to give the user the opportunity to confirm their action to approve the title input. But if you are all agreeing that it's a change of paradigm then we can also leave it like we had.

app/configurator/components/annotation-options.tsx
KerstinFaye: [Here are again all Button sizes in the library](https://www.figma.com/file/kYsVD11OnsPAcDGYCHz5jH/Federal-Projects-Global-Library-%E2%80%93%C2%A0CURRENT?type=design&node-id=97-8822&mode=design&t=FYQj7douDXJw5OXL-4)
In this case I would apply the small button size.

bprusinowski:
Thanks for the review @KerstinFaye! From my side it would be fine to leave the current implementation from this PR, let me know what you think @ptbrowne 👍

ptbrowne:
> I think this ideas is even based on your feedback that you gave ones in a document that you did for UI 

Maybe, I have tried to think more about it, and I think my feedback was  to answer to the need of the user "after inputting a title, what do I do now ". I think we want to 

* respect the current paradigm of auto saving (1)
* have an action at the bottom so that the user knows what to do, and does not look around until seeing the close button (2)

I think this PR solution with the "save" button solves (2) but could be confusing to the user, and somehow does not respect (1). So maybe a good solution would be to have the text of the button only "OK".

KerstinFaye:
> I think this PR solution with the "save" button solves (2) but could be confusing to the user, and somehow does not respect (1). So maybe a good solution would be to have the text of the button only "OK".

That works. Let's use "ok".

ptbrowne:
Thanks 👍

ptbrowne (approved):

PR #1489, opened 2024-05-02
Termset search filter

ptbrowne:
Adds a section in the search page to filter cubes per termsets.

image

- [ ] I could not find yet why the counts are not correct, it seems like my query is correct ?
   The visualize filter is there. I have added the `distinct` in the count clause when
   counting by cube, since multiple dimensions of the same cube might have the termset.

```
PREFIX cube: <https://cube.link/>
  PREFIX meta: <https://cube.link/meta/>
  PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
  PREFIX schema: <http://schema.org/>
  PREFIX sh: <http://www.w3.org/ns/shacl#>
  
  SELECT DISTINCT (COUNT(distinct ?iri) as ?count) ?termsetIri ?termsetLabel WHERE {
      ?iri cube:observationConstraint/sh:property ?dimension .
      ?dimension a cube:KeyDimension .
      ?dimension sh:in/rdf:rest*/rdf:first ?value.
      ?value schema:inDefinedTermSet ?termsetIri .
      ?termsetIri a meta:SharedDimension .
      OPTIONAL { ?termsetIri schema:name ?termsetLabel_en . FILTER(LANG(?termsetLabel_en) = "en") }
OPTIONAL { ?termsetIri schema:name ?termsetLabel_de . FILTER(LANG(?termsetLabel_de) = "de") }
OPTIONAL { ?termsetIri schema:name ?termsetLabel_fr . FILTER(LANG(?termsetLabel_fr) = "fr") }
OPTIONAL { ?termsetIri schema:name ?termsetLabel_it . FILTER(LANG(?termsetLabel_it) = "it") }
OPTIONAL { ?termsetIri schema:name ?termsetLabel_ . FILTER(LANG(?termsetLabel_) = "") }
BIND(COALESCE(?termsetLabel_en, ?termsetLabel_de, ?termsetLabel_fr, ?termsetLabel_it, ?termsetLabel_) AS ?termsetLabel)
  
      
    ?iri schema:workExample <https://ld.admin.ch/application/visualize> .
    ?iri schema:creativeWorkStatus <https://ld.admin.ch/vocabulary/CreativeWorkStatus/Published> .
    ?iri cube:observationConstraint ?shape .
    FILTER NOT EXISTS { ?iri schema:expires ?expiryDate }
  
    }      GROUP BY ?termsetIri ?termsetLabel
```

### DIFF
--- a/app/browser/dataset-browse.tsx
+++ b/app/browser/dataset-browse.tsx
@@ -211,7 +211,16 @@ export const SearchDatasetControls = ({
   );
 };
 
-const defaultNavItemTheme = {
+type NavItemTheme = {
+  activeBg: string;
+  activeTextColor: string;
+  textColor: string;
+  countColor: string;
+  countBg: string;
+  closeColor?: string;
+};
+
+const defaultNavItemTheme: NavItemTheme = {
   activeTextColor: "white",
   activeBg: "primary.main",
   textColor: "text.primary",
@@ -219,7 +228,7 @@ const defaultNavItemTheme = {
   countBg: "grey.300",
 };
 
-const themeNavItemTheme = {
+const themeNavItemTheme: NavItemTheme = {
   activeBg: "category.main",
   activeTextColor: "white",
   textColor: "text.primary",
@@ -227,7 +236,7 @@ const themeNavItemTheme = {
   countBg: "category.light",
 };
 
-const organizationNavItemTheme = {
+const organizationNavItemTheme: NavItemTheme = {
   activeBg: "organization.main",
   activeTextColor: "white",
   textColor: "text.primary",
@@ -235,12 +244,13 @@ const organizationNavItemTheme = {
   countBg: "organization.light",
 };
 
-const termsetNavItemTheme = {
+const termsetNavItemTheme: NavItemTheme = {
   activeBg: "warning.light",
   activeTextColor: "text.primary",
   textColor: "text.primary",
   countColor: "warning.main",
   countBg: "warning.light",
+  closeColor: "warning.main",
 };
 
 const useStyles = makeStyles(() => ({
@@ -391,7 +401,10 @@ const NavItem = ({
         className={classes.removeFilterButton}
         sx={{
           backgroundColor: level === 1 ? theme.activeBg : "transparent",
-          color: level === 1 ? theme.activeTextColor : theme.activeBg,
+          color:
+            level === 1
+              ? theme.closeColor ?? theme.activeTextColor
+              : theme.activeBg,
         }}
       >
         <SvgIcClose width={24} height={24} />
@@ -532,7 +545,7 @@ const NavSection = ({
   icon: React.ReactNode;
   items: (DataCubeTheme | DataCubeOrganization | Termset)[];
   theme: { backgroundColor: string; borderColor: string };
-  navItemTheme: typeof themeNavItemTheme;
+  navItemTheme: NavItemTheme;
   currentFilter?: DataCubeTheme | DataCubeOrganization | Termset;
   filters: BrowseFilter[];
   counts: Record<string, number>;
--- a/app/browser/dataset-browse.tsx
+++ b/app/browser/dataset-browse.tsx
@@ -617,23 +617,12 @@ const NavSection = ({
 
 const TermsetNavSection = ({
   currentFilter,
+  termsetCounts,
 }: {
   currentFilter: Termset | undefined;
+  termsetCounts: { termset: Termset; count: number }[];
 }) => {
-  const locale = useLocale();
-  const [configState] = useConfiguratorState();
-  const { dataSource } = configState;
-
-  const [searchPageQuery] = useSearchPageQuery({
-    variables: {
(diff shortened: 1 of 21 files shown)


app/rdf/query-termsets.ts
```diff
@@ -0,0 +1,82 @@
...
+    termset: {
+      iri: result.termsetIri.value,
+      label: result.termsetLabel.value,
+    },
+  }));
+};
+
+export const getCubeComponentTermsets = async (
+  iri: string,
+  options: {
```
bprusinowski, line 51: Not 100% sure, but looks like it's not used anywhere? Is it for debugging purposes? 👀

app/browser/dataset-browse.tsx
```diff
@@ -290,9 +300,23 @@ const NavChip = ({
...
+      case "DataCubeTheme":
+        return "theme";
+      case "DataCubeOrganization":
+        return "organization";
+      case "DataCubeAbout":
+        throw new Error("Should not happen");
+      case "Termset":
+        return "termset";
+      default:
+        // eslint-disable-next-line
```
bprusinowski, line 314: Wouldn't an exhaustive check work here without disabling eslint?

app/browser/dataset-browse.tsx
```diff
@@ -508,10 +532,10 @@ const NavSection = ({
 }: {
   label: React.ReactNode;
   icon: React.ReactNode;
-  items: (DataCubeTheme | DataCubeOrganization)[];
+  items: (DataCubeTheme | DataCubeOrganization | Termset)[];
```
bprusinowski, line 535: Nit: As we have `DataCube` prefixes for themes and organizations, would it make sense to have e.g. `DimensionTermset`? I think they are related to dimensions and not cubes, right?

app/browser/dataset-browse.tsx
```diff
@@ -592,6 +616,66 @@ const NavSection = ({
...
+      items={termsets}
+      theme={{
+        backgroundColor: "warning.light",
+        borderColor: "warning.main",
+      }}
+      navItemTheme={termsetNavItemTheme}
+      currentFilter={currentFilter}
+      icon={<SvgIcOrganisations width={20} height={20} />}
+      label={<Trans id="browse-panel.termsets">Termsets</Trans>}
+      extra={null}
```
bprusinowski, line 664: Nit: can be omitted?

app/browser/select-dataset-step.tsx
```diff
@@ -143,8 +143,15 @@ const prepareSearchQueryFilters = (filters: BrowseFilter[]) => {
           d.__typename !== SearchCubeFilterType.DataCubeAbout
       )
       .map((d) => {
-        const type = SearchCubeFilterType[d.__typename];
-        return { type, label: d.label, value: d.iri };
+        const type =
+          d.__typename === "Termset"
+            ? SearchCubeFilterType.Termset
+            : SearchCubeFilterType[d.__typename];
```
bprusinowski, line 149: Wouldn't `const type = SearchCubeFilterType[d.__typename]` work here?

app/browser/dataset-browse.tsx
```diff
@@ -592,6 +616,66 @@ const NavSection = ({
...
+      key="orgs"
+      items={termsets}
+      theme={{
+        backgroundColor: "warning.light",
+        borderColor: "warning.main",
+      }}
+      navItemTheme={termsetNavItemTheme}
+      currentFilter={currentFilter}
+      icon={<SvgIcOrganisations width={20} height={20} />}
+      label={<Trans id="browse-panel.termsets">Termsets</Trans>}
```
bprusinowski, line 663: 🇩🇪 🇫🇷 🇮🇹

bprusinowski (approved):
LGTM, cool thing to be able to see termsets here :)

I wonder if the tags shouldn't also be included in dataset result row, as is the case for organizations and themes? But might also be a bit "too much" information, not sure 🤔

app/browser/dataset-browse.tsx
ptbrowne: Yep, it does, it was disabled because previously it was unused, now I use it in the Error. Thanks.

app/browser/select-dataset-step.tsx
ptbrowne: It does now, thanks I missed it.

app/rdf/query-termsets.ts
ptbrowne: It was there because I copied paste a previous file to start this file :facepalm:.

ptbrowne:
Ah thanks for noticing, then counts are okay, we'll just have to revisit the query when the shared dimensions are materialised.

app/browser/dataset-browse.tsx
ptbrowne: Not sure I want to go in this direction, I think I would even want to go in the other direction and have Theme and Organization be without prefix.

PR #1490, opened 2024-05-02
feat: Only display relevant data points in map legend for non-ordinal scale components

### LINKED ISSUE #1476: Legend: Only display data points relevant to the visualization
While testing the improvements to the legends and symbols (issue https://github.com/visualize-admin/visualization-tool/issues/1425), the question arose from Tobi B. if it is possible to only show the relevant items in the legend. 

see example here: https://int.visualize.admin.ch/de/v/Ne4JcJneHn8p

bprusinowski:
Closes #1476


app/charts/shared/legend-color.tsx
```diff
@@ -142,7 +149,7 @@ const useLegendGroups = ({
   const segmentValues =
     segmentFilters?.type === "multi" ? segmentFilters.values : emptyObj;
 
-  return useMemo(() => {
+  return React.useMemo(() => {
```
ptbrowne, line 152: Ah we should settle on one style 😁 I personally do not like to see the code full of "React", is there a specific reason why you like this style ?

app/charts/shared/legend-color.tsx
```diff
@@ -190,45 +199,52 @@ export const LegendColor = memo(function LegendColor(props: LegendColorProps) {
...
     chartConfig,
     title: component.label,
-    values: sortedValues.map((d) => `${d.value}`),
+    values:
+      // If the component is a measure or an ordinal dimension, we want to show all
+      // values for comparison purposes
+      isOrdinalDimension(component) || isOrdinalMeasure(component)
+        ? sortedStringValues
+        : sortedStringValues.filter((v) =>
+            observationLabels.includes(getLabel(v))
```
ptbrowne, line 246: observationLabels could be a Set, so that we can use has instead of includes. Should be more efficient (O(n) vs O(1))

ptbrowne: I think it would be a bit clearer to have legendValues inside it's own memo. sortedStringValues could be computed inside the memo.

ptbrowne (approved):
Cool 👍  I have some suggestions but overall LGTM.

app/charts/shared/legend-color.tsx
bprusinowski: I remember that in the past we needed to `import React from "react"` anyway, so having `import React, { useMemo } from "react"` felt like doing the same thing twice for me, but it looks like [since React 17 it's not needed anymore](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#whats-different-in-the-new-transform) 😁

I am fine with using named imports, maybe we can then have an eslint rule, similar to lodash and d3 to enforce this?

ptbrowne: OK, I can do that. It is fine in this PR, there will be a refactor PR.

app/charts/shared/legend-color.tsx
bprusinowski: Interesting, I wasn't aware that there's such a performance difference between set and array 💡

KerstinFaye:
@bprusinowski sorry, was not there last week to take a look. Just tested it. It's a good solution.

bprusinowski:
Thanks for testing @KerstinFaye 🙇‍♂️

PR #1502, opened 2024-05-06
fix: Add dataSource in case it's missing

bprusinowski:
This PR:
- adds `dataSource` in case it's missing (might be the case for very old chart configs), defaulting to PROD,
- migrates non-cached PROD `dataSource` to cached one.


ptbrowne (approved):
💯 Thanks

PR #1507, opened 2024-05-07
feat: Add safety net for missing key dimension filters

### LINKED ISSUE #1504: multiple std errors are displayed where only one is expected
**Describe the bug**
multiple std errors are displayed where only one is expected

**To Reproduce**

1. Go to https://visualize.admin.ch/v/iyq3m4Ok43Re?dataSource=Prod
3. Scroll down to 3rd chart
4. See multiple std errors

**Expected behavior**
same amount of standard errors as bars - (in this case one)

**Screenshots or video**
image

- Visualize environment and version: PROD 4.0.0

bprusinowski:
Closes #1504

This PR introduces a safety net to not allow missing key dimension filters. Such situation could happen e.g. by manually modifying the `localStorage` state; in the current implementation, we do not add back key dimensions, but rather keep them as addable filters, which breaks the charts.

### DIFF
--- a/app/configurator/components/chart-configurator.tsx
+++ b/app/configurator/components/chart-configurator.tsx
@@ -452,30 +452,44 @@ const useFilterReorder = ({
     state,
   });
   const fetching = possibleFiltersFetching || componentsFetching;
-
-  const { filterDimensions, addableDimensions } = useMemo(() => {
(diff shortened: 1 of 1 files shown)


ptbrowne (approved):
👍 Thanks

PR #1508, opened 2024-05-07
fix: Keep errors in one container

bprusinowski:
This PR adds a container to encompass all errors, so the `subgrid` layout doesn't break in case of errors. To see what's broken currently, see [this link](https://blw-agricultural-market-data-platform.vercel.app/blog/test).


app/components/chart-published.tsx
```diff
@@ -248,49 +248,51 @@ const ChartPublishedInner = (props: ChartPublishInnerProps) => {
...
+              <Box sx={{ mb: 4 }}>
+                <HintBlue iconName="hintWarning">
+                  <Trans id="dataset.hasImputedValues">
+                    Some data in this dataset is missing and has been
+                    interpolated to fill the gaps.
+                  </Trans>
+                </HintBlue>
+              </Box>
+            )}
+          </div>
```
ptbrowne, line 295: Maybe we could have a ChartWarnings here that would have the logic to extract errors, display, and wrap them ?

This way we would not have to repeat the logic ? I guess we would need to have a way to filter some errors ?

```
<ChartWarnings ignore={['imputation']} /> // in preview
```

ptbrowne:
@bprusinowski Do you think you could take a little time to rebase this PR and fix the error 🙏 ? I guess we could forgo the refactor I was proposing, in the interest of getting this PR out 😅

bprusinowski:
Sure, @ptbrowne you mean the merge conflict?

ptbrowne (approved):

PR #1510, opened 2024-05-08
ensure status in Draft|Published

Rdataflow:
ensure only Draft|Published are shown
WRT https://github.com/zazuko/cube-creator/wiki/LINDAS-Specifics#needed-attributes-that-a-cube-shows-up-on-visualizeadminch

there are rare cases with different status i.e. https://s.zazuko.com/m6iuJB which shouldn't be shown

nb: this PR also makes results more consistent with https://github.com/visualize-admin/visualization-tool/blob/eef59af39594123bacc535484df6654550830fa1/app/rdf/queries.ts#L69-L74

cc @bprusinowski

### DIFF
--- a/app/rdf/query-utils.ts
+++ b/app/rdf/query-utils.ts
@@ -58,7 +58,8 @@ export const makeVisualizeDatasetFilter = (options?: {
     ${cubeIriVar} schema:workExample <https://ld.admin.ch/application/visualize> .
     ${
       includeDrafts
-        ? ""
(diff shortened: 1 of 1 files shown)


bprusinowski:
Thanks @Rdataflow, LGTM! Side note: I think this situation technically shouldn't happen, looking at the below screenshot? Maybe it could be possible with some custom pipeline? 🤔 

Screenshot 2024-05-08 at 08 54 11

bprusinowski (approved):

Rdataflow:
@bprusinowski it's likely due to a combination of legacy project and code in C-C that doesn't expire deprecated cubes properly :+1:

PR #1512, opened 2024-05-08
Joined dimensions improvements

ptbrowne:
Metadata panel

- Show additional information for joined dimensions in metadata panel
- Add info icon tooltip to joined dimensions in metadata panel

Add dataset dialog

- Show caution alert (state saved in local storage)

### DIFF
--- a/app/configurator/components/add-dataset-dialog.tsx
+++ b/app/configurator/components/add-dataset-dialog.tsx
@@ -130,6 +130,16 @@ const useStyles = makeStyles((theme: Theme) => ({
   newAnnotation: {
     color: theme.palette.success.main,
   },
+  listTag: {
+    whiteSpace: "break-spaces",
+  },
+  listItemSecondary: {
+    display: "flex",
+    flexWrap: "wrap",
+    alignItems: "center",
+    gap: 1,
+    paddingTop: theme.spacing(1),
+  },
 }));
 
 const NewAnnotation = (props: TypographyProps) => {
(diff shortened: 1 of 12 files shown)


app/configurator/components/chart-type-selector.tsx
```diff
@@ -27,9 +20,10 @@ import { IconButton } from "@/configurator/components/icon-button";
 import { useChartType } from "@/configurator/config-form";
 import { ConfiguratorStateWithChartConfigs } from "@/configurator/configurator-state";
 import { useDataCubesComponentsQuery } from "@/graphql/hooks";
-import { Icon } from "@/icons";
 import { useLocale } from "@/locales/use-locale";
 
+import { InfoIconTooltip } from "./info-icon-tooltip";
```
bprusinowski, line 25: Super-nit: I think the default behavior of VS Code is to use relative imports up to 2 levels high, and aliased is available for longer ones. I added this setting to VS Code settings to always use aliased one:

```
"typescript.preferences.importModuleSpecifier": "non-relative",
```

maybe it could be added to workspace settings so it's shared?

app/configurator/components/add-dataset-dialog.tsx
```diff
@@ -156,6 +171,54 @@ const NewAnnotation = (props: TypographyProps) => {
   );
 };
 
+const useCautionAlert = () => {
+  const [isOpen, setIsOpen] = useLocalState<boolean>(
```
bprusinowski, line 175: Nit: I think you don't have to use `<boolean>`, as the type should be inferred by looking at the implementation of `useLocalState`?

app/utils/use-local-state.ts
```diff
@@ -0,0 +1,23 @@
...
+  useEffect(() => {
+    localStorage.setItem(key, JSON.stringify(state));
+  }, [key, state]);
+
+  const updateState = useCallback((value: T) => {
+    setState(value);
+  }, []);
+
+  return [state, updateState];
+};
```
bprusinowski, line 21: 👌

bprusinowski (approved):
LGTM 💯

app/configurator/components/chart-type-selector.tsx
ptbrowne: I already have it 🤷 I just noticed that info-icon-tooltip should be in app/components directly, not in configurator/, I am changing this.

PR #1516, opened 2024-05-16
Preparation dashboard

ptbrowne:
InteractiveFilters state has been lifted up, out of the chart themselves, inside a global InteractiveFilters provider, which creates one zustand store for each chart, and then provides it to the charts through InteractiveFiltersChartProvider. This will enable control of multiple interactive filters by one component.


app/components/chart-published.tsx
```diff
@@ -292,82 +295,84 @@ const ChartPublishedInner = (props: ChartPublishInnerProps) => {
...
-                sx={{
-                  justifyContent: meta.title[locale]
-                    ? "space-between"
-                    : "flex-end",
-                  gap: 2,
-                }}
-              >
-                {meta.title[locale] ? (
-                  <Title text={meta.title[locale]} />
+            <InteractiveFiltersProvider chartConfigs={state.chartConfigs}>
```
bprusinowski, line 298: I think now we have two nested `InteractiveFiltersProviders`, one in `ChartPublished` and one here, in `ChartPublishedInner`. Shouldn't there only be one?

app/stores/interactive-filters.tsx
```diff
@@ -133,15 +135,69 @@ const interactiveFiltersStoreCreator: StateCreator<State> = (set) => {
...
 >(undefined);
 
+export const getSharedFilters = (
+  chartConfigs: ChartConfig[]
+): SharedFilters => {
+  const allActiveFilters = chartConfigs.flatMap((c) => {
+    const interactiveFiltersConfig = c.interactiveFiltersConfig;
+    if (!interactiveFiltersConfig) {
+      return [];
+    }
```
bprusinowski, line 164: Is it possible that the `interactiveFiltersConfig` is undefined? I think we have it for every chart type 🤔

app/package.json
```diff
@@ -38,7 +38,8 @@
     "@rdfjs/namespace": "^2.0.1",
     "@reach/auto-id": "^0.15.3",
     "@sentry/nextjs": "^7.112.2",
-    "@storybook/nextjs": "^8.0.9",
+    "@storybook/nextjs": "^8.1.0",
+    "@storybook/react-vite": "^8.1.0",
```
bprusinowski, line 42: Should we have two Storybook frameworks?

bprusinowski (approved):
LGTM 💯 

When testing, I noticed one issue (which is also there on TEST, so not a regression, but related to this area I think):
1. Go to https://visualization-tool-git-feat-preparation-dashboard-ixt1.vercel.app/en/create/new?cube=https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/9&dataSource=Prod
2. Add a second chart.
3. Go back to first chart.
4. Go to X axis.
5. Enable interactive filter.
6. Proceed to layout options.
7. See that the chart changes filter range to 2014-2014.

I can also open a new issue, let me know 👍

app/charts/shared/overlay-voronoi.tsx
```diff
@@ -26,7 +28,13 @@ export const InteractionVoronoi = memo(function InteractionVoronoi({
...
     (d) => yScale(getY(d) ?? NaN)
   );
-  const voronoi = delaunay.voronoi([0, 0, chartWidth, chartHeight]);
+
+  const voronoi = delaunay.voronoi([
+    0,
+    0,
+    atLeastZero(chartWidth),
+    atLeastZero(chartHeight),
+  ]);
```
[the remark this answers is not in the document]
ptbrowne: I have seen a case yes, so this is defensive code. I do not remember when though.

app/package.json
ptbrowne: No, we should not, thanks for noticing. This is a test I made, because I've noticed that the fast refresh was not working in Storybook (we have fast refresh in the app, not in storybook :/). I've removed now the @storybook/react-vite package.

app/stores/interactive-filters.tsx
ptbrowne: Tables do not have it.

image

app/components/chart-published.tsx
ptbrowne: Nice catch.

app/components/chart-published.tsx
```diff
@@ -108,9 +111,9 @@ export const ChartPublished = (props: ChartPublishedProps) => {
           renderChart={renderChart}
         />
       )}
-    </>
+    </InteractiveFiltersProvider>
   ) : (
-    <>
+    <InteractiveFiltersProvider chartConfigs={state.chartConfigs}>
```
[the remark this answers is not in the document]
ptbrowne: Yep 🎉

ptbrowne:
Thanks for the thorough review, I've created an issue for the problem you saw 🙇

PR #1518, opened 2024-05-16
Preparation dashboard 2

ptbrowne:
- feat: Add react-grid
- refactor: Easier support for other layout types
- feat: Add icons & text for tiles layout
- fix: Wrap debug panel in div to keep number of rows stable
- fix: Too many import
- refactor: ChartPanelLayout hides underlying layout components


app/components/react-grid.tsx
```diff
@@ -0,0 +1,309 @@
...
+    "& .react-grid-item .add": {
+      cursor: "pointer",
+    },
+    "& .react-grid-dragHandleExample": {
+      cursor: "move",
+    },
+  },
+}));
+
+export const ChartGridLayout = (props: Props) => {
```
bprusinowski, line 208: Extra-nit: This is a personal preference, but I find it nice when props are defined just above the component, so we don't have to search for them 🕵️‍♂️

app/locales/de/messages.po
```diff
@@ -571,11 +571,15 @@ msgstr "Tab Layout"
...
+msgstr "Hoch"
+
+#: app/configurator/components/field-i18n.ts
+msgid "controls.layout.tiles"
+msgstr "Kacheln"
 
 #: app/configurator/components/field-i18n.ts
 msgid "controls.layout.vertical"
-msgstr ""
+msgstr "Hoch"
```
bprusinowski, line 582: Not 100% sure what we should put here, but for other languages translations for Tall and Vertical layout were different (here they are the same) 👀

app/components/chart-preview.tsx
```diff
@@ -484,7 +482,10 @@ export const ChartPreviewInner = (props: ChartPreviewInnerProps) => {
                   onToggleTableView={handleToggleTableView}
                   dimensions={dimensions}
                 />
-                <DebugPanel configurator interactiveFilters />
+                {/* Wrap in div for subgrid layout */}
+                <div>
+                  <DebugPanel configurator interactiveFilters />
+                </div>
```
bprusinowski, line 482: 👍

bprusinowski (approved):
LGTM 👌

app/locales/de/messages.po
ptbrowne: Fixed

PR #1519, opened 2024-05-17
Add dataset improvements

ptbrowne:
- Adding a toast on the added dataset row when dataset has been added (an event emitter is used to implement the functionality to facilitate connecting two components that are completely unaware of each other. used strong typing to facilitate use of event emitter)
- Try to find an observation with a lot of data to show the preview (to counter act the fact that years can not align between two cubes)
- Tweak add dataset dialog (typography, select menu position)
- Added snackbar since we might use it in the future


app/configurator/components/dataset-control-section.tsx
```diff
@@ -40,50 +75,93 @@ const DatasetRow = ({
...
+        setTimeout(() => {
+          setFlash(false);
+        }, 5000);
+      }
+    };
+    ee.on("dataset-added", handle);
+    return () => {
+      ee.removeListener("dataset-added", handle);
+    };
+  });
```
bprusinowski, line 101: Nice technique :) 👍 

I wonder if there's a need to explicitly `useEffect` and clean up event listeners like this, can't you use `useEventEmitter("dataset-added", () => { ... })` (it looks like useEffect and cleanup is already included in the hook)? On maybe I missed somthing 😱

app/configurator/components/add-dataset-dialog.tsx
```diff
@@ -483,6 +485,15 @@ export const PreviewDataTable = ({
...
   };
 
+  const previewObservations = useMemo(() => {
+    const data = observations.data?.dataCubesObservations.data ?? [];
+    const bestObservation = maxBy(data, (obs) => {
+      return allColumns.reduce((acc, dim) => acc + (obs[dim.iri] ? 1 : 0), 0);
+    });
+    const index = bestObservation ? data.indexOf(bestObservation) : 0;
+    return data.slice(index, index + 8);
+  }, [allColumns, observations.data?.dataCubesObservations.data]);
```
bprusinowski, line 525: 👌

bprusinowski: That's probably not going to happen, but if the best observation is the last one, would we only see one row?

bprusinowski (approved):
LGTM, cool changes 😎 

When testing the deployment preview, the snackbar didn't appear when I merged the datasets for the first time; only after I already merged, unmerged and merged again it was shown.

This was reproducible on every "clean" entry into the editing mode after the page refresh. Not sure if I missed something or this is a bug? 🐛 

https://github.com/visualize-admin/visualization-tool/assets/52032047/08506568-034b-49a0-b127-7cbec789ec56

app/configurator/components/dataset-control-section.tsx
ptbrowne: Ahah I had forgotten I had done that, since I copy pasted this file from another project. Thanks 😅

app/configurator/components/add-dataset-dialog.tsx
ptbrowne: Thanks, I handle the edge case now 🙇 https://github.com/visualize-admin/visualization-tool/pull/1519/commits/4cfbeb87822a60eda69d9973a468dc5cc42a2a02

ptbrowne:
Thanks for the bug for the first time, I fixed it 🙇

PR #1524, opened 2024-05-23
Fix preview

ptbrowne:
Multiple problems were fixed.

- Needed to wrap the ChartPublished inside a ChartConfiguratorState provider
- Change config that was triggering a Stardog error (:shrug:)
- Migrate config to latest version

### DIFF
--- a/app/pages/_preview.tsx
+++ b/app/pages/_preview.tsx
@@ -1,5 +1,7 @@
 import { Box } from "@mui/material";
-import React, { useEffect } from "react";
+import { useEffect } from "react";
+
+import { migrateConfiguratorState } from "@/utils/chart-config/versioning";
 
 const Page = () => {
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


app/pages/_preview.tsx
```diff
@@ -30,7 +32,7 @@ const Page = () => {
 
 export default Page;
 
-const chartState = {
+const chartState = migrateConfiguratorState({
```
bprusinowski, line 35: 👍 so it was not an immer error after all 😅

bprusinowski (approved):
👍

app/pages/_preview.tsx
ptbrowne: I added an issue for in the future check this problem #1525

Rdataflow:
@ptbrowne big thanks 🎉

PR #1545, opened 2024-05-30
Dashboard filters

ptbrowne:
Work in progress.

### Functionality

It is now possible to activate a single shared temporal filter for all charts depending on a particular temporal dimensions.

- [x] Shared time slider, takes unit from dimension
- [x] UI to activate/deactivate year filter
- [x] Ability to change default from/to  of the time rane
- [x] Dashboard filter is not shown inside the "configurating" view
- [x] Chart config gains a new attribute dashboardFilters where sharedFilters are stored

### How to test

1. Start from Photovoltaikanlagen
2. Add a line chart
3. Add a new dataset like Median electricity tariff for Switzerland
4. Proceed to layout options
5. Choose Dashboard as a layout
6. See that a section "Shared filters" appears in the left panel
7. Activate the shared filter on Year
8. You can now interactively control both charts with the same filter
9. Control the default extent through the date pickers in the left panel
10. Publish the chart

### Next steps in following PRs

- Implement the logic for data filters
- Implement the logic for "point in time" filter

### Refactors

Take the opportunity to explode the big configurator-state file into multiple ones. With this, it should be easier to add changes. I'd like your feedback on this Bartosz also.


app/configurator/components/ui-helpers.ts
```diff
@@ -27,6 +27,12 @@ const parseDay = timeParse("%Y-%m-%d");
 const parseMonth = timeParse("%Y-%m");
 const parseYear = timeParse("%Y");
 
+const formatSecond = timeFormat("%Y-%m-%dT%H:%M:%S");
+const formatMinute = timeFormat("%Y-%m-%dT%H:%M");
+const formatDay = timeFormat("%Y-%m-%d");
+const formatMonth = timeFormat("%Y-%m");
+const formatYear = timeFormat("%Y");
```
bprusinowski, line 35: Thinking out loud if these shouldn't be connected to RDF definitions? See that we already defined some of them here: https://github.com/visualize-admin/visualization-tool/blob/92532b7e066e25b01ed719e841a79e70347ef8d9/app/rdf/parse.ts#L301-L305

app/stores/interactive-filters.tsx
```diff
@@ -158,8 +161,24 @@ const InteractiveFiltersContext = createContext<
...
 export const getSharedFilters = (
-  chartConfigs: ChartConfig[]
+  state: ConfiguratorStateWithChartConfigs
 ): SharedFilters => {
+  const chartConfigs = state.chartConfigs;
+
+  // We show shared filters only when the chart is not being configured
+  if (state.state === "CONFIGURING_CHART") {
+    return [];
+  }
```
bprusinowski: Nit: `isConfiguring(state)`

app/utils/chart-config/versioning.ts
```diff
@@ -1137,6 +1137,26 @@ export const configuratorStateMigrations: Migration[] = [
...
+        },
+      };
+      return newConfig;
+    },
+    down: (config) => {
+      const newConfig = { ...config, version: "3.2.1" };
+      delete newConfig.dashboardFilters;
+      return newConfig;
+    },
+  },
```
bprusinowski, line 1159: I think you also need to add migration for configurator state, otherwise this wouldn't be triggered I think 🤔

app/configurator/configurator-state/init.tsx
```diff
@@ -0,0 +1,176 @@
...
+    getStateWithCurrentDataSource(SELECTING_DATASET_STATE),
+    { chartConfig }
+  );
+};
+/**
+ * Tries to parse state from localStorage.
+ * If state is invalid, it is removed from localStorage.
+ */
+
+export const initChartStateFromLocalStorage = async (
```
bprusinowski, line 107: Nit: Not sure if it works if there's an empty line between the docstring and function definition? 👀

app/configurator/configurator-state/reducer.tsx
```diff
@@ -0,0 +1,1066 @@
...
+const reducerLogging = false;
+const withLogging = <TState, TAction extends { type: unknown }>(
+  reducer: Reducer<TState, TAction>
+) => {
+  return (state: Draft<TState>, action: TAction) => {
+    const res = reducer(state, action);
+    return res;
+  };
+};
+export const reducer = reducerLogging ? withLogging(reducer_) : reducer_;
```
bprusinowski: 👍

app/configurator/configurator-state/reducer.tsx
```diff
@@ -1048,13 +1048,34 @@ export const reducer_: Reducer<ConfiguratorState, ConfiguratorStateAction> = (
...
+      return draft;
+
     default:
       throw unreachableError(action);
   }
 };
 
 /** Turn this on for the reducer to log state, action and result */
-const reducerLogging = false;
+const reducerLogging = true;
```
bprusinowski, line 1096: 🔍

app/configurator/components/layout-configurator.tsx
```diff
@@ -150,17 +184,28 @@ const LayoutSharedFiltersConfigurator = () => {
...
-            {sharedFilters.map((filter) => {
+            {potentialSharedFilters.map((filter) => {
+              const dimension = dimensionsByIri[filter.componentIri];
+              const sharedFilter = sharedFiltersByIri[filter.componentIri];
+              console.log(
+                dimension,
+                sharedFilter,
+                filter.componentIri,
+                sharedFiltersByIri
+              );
```
bprusinowski: 👓

app/stores/interactive-filters.tsx
```diff
@@ -141,80 +142,62 @@ type InteractiveFiltersContextValue = [
...
-    const { timeRange } = interactiveFiltersConfig;
-    return [{ type: "timeRange" as const, value: timeRange }].filter(
-      (x) => x.value.active
+export const getPotentialSharedFilters = (
+  chartConfigs: ChartConfig[]
+): PotentialSharedFilter[] => {
+  const temporalDimensions = chartConfigs.flatMap((c) => {
+    const chartSpec = getChartSpec(c);
+    const temporalEncodings = chartSpec.encodings.filter((x) =>
+      x.componentTypes.some((x) => x === "TemporalDimension")
```
bprusinowski, line 169: Wondering if `TemporalEntityDimension` wouldn't also work "out of the box", as it looks like it has a `timeUnit` just like `TemporalDimension`?

app/stores/interactive-filters.tsx
```diff
@@ -141,80 +142,62 @@ type InteractiveFiltersContextValue = [
...
-    type: filters[0].type,
-    value: (() => {
-      const type = filters[0].type;
-      switch (type) {
-        case "timeRange":
+    const chartTemporalDimensions = temporalEncodings
+      .map((x) => {
+        // @ts-ignore
+        const field = x.field in c.fields ? c.fields[x.field] : undefined;
+        if (field && "componentIri" in field) {
```
bprusinowski, line 175: Nit: Might be a bit more clear with e.g. `config` instead of `c` and `encoding` instead of `x`. But might also be clear in the code, just something I wasn't sure in GitHub interface when reviewing :)

app/components/chart-panel.tsx
```diff
@@ -96,15 +98,14 @@ export const ChartPanelLayout = (props: ChartPanelLayoutProps) => {
...
-       * TODO: Should be shown only in Dashboard layout, not in separate URLs
-       *  We want to completely remount this component if chartConfigs change,
-       */}
-      <DashboardInteractiveFilters
-        key={chartConfigs.map((x) => x.key).join(",")}
-      />
+      {state.layout.type === "dashboard" ? (
+        <DashboardInteractiveFilters
+          key={chartConfigs.map((x) => x.key).join(",")}
+        />
```
bprusinowski, line 107: Nit: Might be worth to still keep the comment?

app/configurator/components/layout-configurator.tsx
```diff
@@ -187,49 +204,50 @@ const LayoutSharedFiltersConfigurator = () => {
...
+                      label={
+                        <Typography variant="body2">
+                          <Trans id="controls.section.shared-filters.shared-switch">
+                            Shared
+                          </Trans>
+                        </Typography>
+                      }
+                      control={
+                        <Switch
+                          checked={!!sharedFilter ?? false}
```
bprusinowski, line 236: Nit: I think you don't need `?? false` since the value is already casted to boolean?

app/configurator/components/layout-configurator.tsx
```diff
@@ -240,6 +258,167 @@ const LayoutSharedFiltersConfigurator = () => {
...
+          <DatePickerField
+            name={`interactive-date-picker-${dimension.iri}`}
+            value={parseDate(sharedFilter.presets.from) as Date}
+            onChange={handleChangeFromDate}
+            isDateDisabled={(d) => !optionValues.includes(formatDate(d))}
+            timeUnit={timeUnit}
+            dateFormat={formatDate}
+            minDate={minDate}
+            maxDate={maxDate}
+            disabled={false}
```
bprusinowski, line 384: Nit: No need to pass `false` (I guess it's enabled by default)?

bprusinowski (approved):
Code LGTM, with some questions and small nits 💯 

When it comes to testing the deployment preview, I think it breaks when you merge cubes:
1. Go to [this link](https://visualization-tool-git-feat-dashboard-filters-ixt1.vercel.app/en/create/new?cube=https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/9&dataSource=Prod).
2. Add a line chart.
3. Merge **Heavy metal soil contamination dataset** to the line chart.
4. Proceed to layout options.
5. Switch to Free canvas Dashboard layout.
6. Enable shared filters.
7. See that the filters do not work and the slider is not shown on top of the charts.

bprusinowski:
For the breaking of giant configurator state into smaller files, I think it's a very good step into making this more manageable, which also shouldn't really change any behavior if I am not mistaken, so I am all for it :)

app/utils/chart-config/versioning.ts
ptbrowne: Seen together, dashboardFilters is directly at configuratorState level, so we are good.

app/configurator/components/ui-helpers.ts
bprusinowski: Yes, I was wondering if this would make sense and is actually needed. It looks that the formatters I shared in the link are only used in `interpolateTimeValues` function from the same file which is not used anywhere; so I guess we could just clean up and remove these server-side values 👀

app/utils/chart-config/versioning.ts
bprusinowski: Discussed offline :)

app/stores/interactive-filters.tsx
ptbrowne: I created an issue for that : https://github.com/visualize-admin/visualization-tool/issues/1555

bprusinowski: Thanks 🙇‍♂️

PR #1551, opened 2024-06-03
Dashboard canvas works in published mode

ptbrowne:
Dashboard canvas works in published mode

I had not repercuted some of the improvements from chart-preview to chart-published.
Passing ref, className, and children is necessary for react-grid-layout to work
correctly.


bprusinowski (approved):
👍

PR #1561, opened 2024-06-05
fix: Whiskers not updating correctly

### LINKED ISSUE #1558: "Preview via API" renders error bars wrong
**Describe the bug**
When rendering a chart in an iframe via the preview api (https://visualize.admin.ch/docs/charts/preview-via-api), the error bars are all rendered at the top left and the tooltips are positioned weirdly (see screenshot below)

image

There are also 10'000+ error messages like these:
```
attr.js:31 Error: <rect> attribute height: A negative value is not valid. ("-8.459137207031262")
attr.js:31 Error: <rect> attribute height: A negative value is not valid. ("-12.37485956054688")
attr.js:31 Error: <rect> attribute height: A negative value is not valid. ("-5.6884632812499945")
attr.js:31 Error: <rect> attribute height: A negative value is not valid. ("-2.934952822265643")
attr.js:31 Error: <rect> attribute height: A negative value is not valid. ("-3.162981738281246")
attr.js:31 Error: <rect> attribute height: A negative value is not valid. ("-2.1478852734374954")
attr.js:31 Error: <rect> attribute height: A negative value is not valid. ("-3.4572125976562518")
attr.js:31 Error: <rect> attribute height: A negative value is not valid. ("-3.4682462548828")
attr.js:31 Error: <rect> attribute height: A negative value is not valid. ("-8.473848749999988")
```

**To Reproduce**
 I called the iframe like described in the docs:
```html
<iframe id="chart" src="https://test.visualize.admin.ch/preview"></iframe>
```

```js
const iframe = document.getElementById("chart");
iframe.onload = () => {
  const iframeWindow = iframe?.contentWindow;
  if (iframeWindow) {
    iframeWindow.postMessage(chartState, "*");
  }
};
```

With chartState e.g. 
```js
{
    "version": "3.2.1",
    "state": "CONFIGURING_CHART",
    "dataSource": {
        "type": "sparql",
        "url": "https://lindas-cached.cluster.ldbar.ch/query"
    },
    "layout": {
        "type": "tab",
        "meta": {
            "title": {
                "de": "",
                "en": "",
                "fr": "",
                "it": ""
            },
            "description": {
                "de": "",
                "en": "",
                "fr": "",
                "it": ""
            }
        }
    },
    "chartConfigs": [
        {
            "key": "zTuhSfYsiB7W",
            "version": "3.2.0",
            "meta": {
                "title": {
                    "en": "",
                    "de": "",
                    "fr": "",
                    "it": ""
                },
                "description": {
                    "en": "",
                    "de": "",
                    "fr": "",
                    "it": ""
                }
            },
            "cubes": [
                {
                    "iri": "https://environment.ld.admin.ch/foen/nfi/nfi_C-20/cube/2023-3",
                    "publishIri": "https://environment.ld.admin.ch/foen/nfi/nfi_C-20/cube/2023-3",
                    "filters": {}
                }
            ],
            "chartType": "column",
            "interactiveFiltersConfig": {
                "legend": {
                    "active": false,
                    "componentIri": ""
                },
                "timeRange": {
                    "active": false,
                    "componentIri": "https://environment.ld.admin.ch/foen/nfi/inventory",
                    "presets": {
                        "type": "range",
                        "from": "",
                        "to": ""
                    }
                },
                "dataFilters": {
                    "active": false,
                    "componentIris": []
                },
                "calculation": {
                    "active": false,
                    "type": "identity"
                }
            },
            "fields": {
                "x": {
                    "componentIri": "https://environment.ld.admin.ch/foen/nfi/inventory",
                    "sorting": {
                        "sortingType": "byAuto",
                        "sortingOrder": "asc"
                    }
                },
                "y": {
                    "componentIri": "https://environment.ld.admin.ch/foen/nfi/Topic/73r"
                }
            }
        }
    ],
    "activeChartKey": "zTuhSfYsiB7W"
}
```

See [this jsfiddle ](https://jsfiddle.net/b3ng8Lhp/) for a minimal example.

**Expected behavior**
I expect to render the chart the same as when I do it in visualize directly, like this:

image

bprusinowski:
Fixes #1558


ptbrowne (approved):
💯

app/charts/column/columns-grouped.tsx
```diff
@@ -28,6 +28,9 @@ export const ErrorWhiskers = () => {
   const ref = useRef<SVGGElement>(null);
   const enableTransition = useTransitionStore((state) => state.enable);
   const transitionDuration = useTransitionStore((state) => state.duration);
+  // As xScale object is not re-created when its domain or range changes, we use this
+  // trick to force the re-rendering of the whiskers when the xScale changes.
+  const bandwidth = xScaleIn.bandwidth();
```
ptbrowne, line 33: Super nit: So if I understand correctly, if the domain and range change but the bandwidth happen to stay the same, we would not re-render right ? Will never happen I guess, but semantically we could also use directly xScaleIn.range() and xScaleIn.domain() in the useMemo deps right ?

bprusinowski: Here the problem is that we had `xScale` as a dependency, and this object was not updated when chart dimensions changed, because its range and domain were updated outside of `useMemo` in which it's created. I think it might also work to add xScaleIn.bandwidth() as dependency, but wondering if it would make it more clear? I think we still would need to have this comment (or re-create the whole `xScale` object when dimensions change) 🤔

PR #1565, opened 2024-06-05
feat: Make the y scale's min value dynamic in some cases

### LINKED ISSUE #1500: For scaleType `Interval` - start y-Axis at lowest value (instead of 0)
**DoD**
 The y-Axis of chart dimensions of scaleType Interval starts at the lowest value (according to the available data). Dimensions of Ratio scale keep their current behaviour and start at 0.   
   
If there is a visualization, where the relevant y-axis is of scaleType Interval values are typically in a limited range of values that is unrelated to zero, letting the y-axis start at zero is not helpful for the visualization. See screenshot:

Image

For IntervalScale Dimensions there could be a "auto-tune" function to detect the right value range for the y-axis depending on the data that would be displayed. But this function should be only for IntervalScale, _while for RatioScale, it is very important that the y-axis starts exactly at zero no matter what the data is._ An example how it should loke with "autotune" is given below:

Image

**Usecases and impact**
I think there must be tons of visualizations that would benefit from such a behaviour.

**updated description**
this updated description is provided and coordinated with Sabine and Tobi

related but outdated_ https://github.com/visualize-admin/visualization-tool/issues/1255Usecases and impact

bprusinowski:
Closes #1500


app/charts/line/lines-state.tsx
```diff
@@ -132,12 +133,18 @@ const useLinesState = (
...
+  );
   const maxValue = max(scalesData, getY) ?? 0;
   const yDomain = [minValue, maxValue];
   const yScale = scaleLinear().domain(yDomain).nice();
 
-  const paddingMinValue = Math.min(min(paddingData, getY) ?? 0, 0);
+  const paddingMinValue = getMaybeDynamicMinYScaleValue(
+    yMeasure.scaleType,
+    min(paddingData, getY)
+  );
```
ptbrowne, line 147: Would it make sense to use a hook like useYScale similar to the useYScales used in he combo charts ?

PR #1566, opened 2024-06-05
Sync dashboard layouts with chart configs

ptbrowne:
This makes sure that while we are adding and removing chart configs, when we go back to the canvas, the the charts have a correct size (previously they would be very tiny). 
Also it makes it easier to resize the last charts vertically by adding a margin bottom to the canvas layout in preview mode.


app/configurator/configurator-state/reducer.tsx
```diff
@@ -490,6 +493,7 @@ export const updateColorMapping = (
       }
     }
 
+    console.log({ colorMapping, path: path.concat("colorMapping") });
```
bprusinowski, line 496: 🔍

app/configurator/configurator-state/reducer.tsx
```diff
@@ -1094,13 +1102,68 @@ export const reducer_: Reducer<ConfiguratorState, ConfiguratorStateAction> = (
 };
 
 /** Turn this on for the reducer to log state, action and result */
-const reducerLogging = false;
+const reducerLogging = true;
```
bprusinowski, line 1105: 🔎

app/configurator/configurator-state/reducer.tsx
```diff
@@ -1094,13 +1102,68 @@ export const reducer_: Reducer<ConfiguratorState, ConfiguratorStateAction> = (
...
 const withLogging = <TState, TAction extends { type: unknown }>(
   reducer: Reducer<TState, TAction>
 ) => {
   return (state: Draft<TState>, action: TAction) => {
     const res = reducer(state, action);
+    console.log(`Action: ${action.type}`, action, res);
```
bprusinowski, line 1111: 👁️‍🗨️

app/configurator/configurator-state/reducer.tsx
```diff
@@ -1094,13 +1102,68 @@ export const reducer_: Reducer<ConfiguratorState, ConfiguratorStateAction> = (
...
+  draft: WritableDraft<ConfiguratorState>
+) {
+  if (
+    hasChartConfigs(draft) &&
+    draft.layout.type === "dashboard" &&
+    draft.layout.layout === "canvas"
+  ) {
+    const chartConfigKeys = draft.chartConfigs.map((c) => c.key).sort();
+
+    const canvasLayouts = draft.layout.layouts["lg"];
```
bprusinowski, line 1127: Is this property always there? What if the size of the canvas was initialized as "md"?

bprusinowski (approved):
LGTM 👍

app/configurator/configurator-state/reducer.tsx
ptbrowne: This one is OK, it's when we have activated the loggin.

bprusinowski: Ah, you're right 👍

app/configurator/configurator-state/reducer.tsx
ptbrowne: Right now this property is always there yes. The chart is always initialized with md.

ptbrowne: The responsive behavior could be checked later, but I think right now it works well, I've added https://github.com/visualize-admin/visualization-tool/issues/1567

bprusinowski: Thanks 💯

PR #1571, opened 2024-06-07
fix: Full data download (merged cubes)

### LINKED ISSUE #1536: Full data set cannot be downloaded when cubes are merged
**Description**
I am not sure if this is the intended behaviour, but when a chart is created out of 2 merged datasets the full data set cannot be downloaded, only the chart data set. If it is not possible to download the full dataset in a combined chart, we should consider disabling the option. 

**To Reproduce**
1. Go to https://int.visualize.admin.ch/en/v/FgIcET7GlAgw?dataSource=Prod
2. Click on "download dataset" click on "full data set" (either CSV or XLS) and see that nothing happens 

**Additional Context**
- The chart data set is downloadble. 

Visualize Version: INT v4.5.1 ([5ed675c](https://github.com/visualize-admin/visualization-tool/commit/5ed675c47b48be9d2a3de303ccaec368d31ea0e5))
Browser: Chrome

bprusinowski:
Fixes #1536

### DIFF
--- a/app/browser/dataset-preview.tsx
+++ b/app/browser/dataset-preview.tsx
@@ -202,7 +202,7 @@ export const DataSetPreview = ({
               <DataDownloadMenu
                 dataSource={dataSource}
                 title={dataCubeMetadata.title}
-                filters={[variables.cubeFilter]}
+                filters={variables.cubeFilter}
               />
             </Flex>
             <FirstTenRowsCaption />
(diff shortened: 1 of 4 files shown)


ptbrowne:
I do not understand what's happening here ? Do we a joinBy iri inside `dataCubeMetadata.iri` in case of merged cubes ?

bprusinowski:
Currently, we pass `queryFilters` which is an array in case of having merged cubes. As full data download is scoped per given dataset, we need to pass correct query filters (per given cube) 👍

ptbrowne (approved):
Thanks for the explanation !

bprusinowski:
I refactored to make it clear from the code 👍

PR #1573, opened 2024-06-07
chore: eIAM cleanup

adintegra:
As we are using a public OIDC flow with PKCE we should be able to remove the client secret.


bprusinowski (approved):
👌

ptbrowne (approved):

PR #1575, opened 2024-06-07
feat: Exploration zoom WIP

ptbrowne:
Short exploration to see if a zoom feature would be possible/useful
for dashboards.

### DIFF
--- a/app/components/chart-panel.tsx
+++ b/app/components/chart-panel.tsx
@@ -1,7 +1,7 @@
 import { Box, BoxProps, Theme } from "@mui/material";
 import { makeStyles } from "@mui/styles";
 import clsx from "clsx";
-import React, { forwardRef, HTMLProps } from "react";
+import React, { HTMLProps, forwardRef } from "react";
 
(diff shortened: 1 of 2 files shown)


bprusinowski (approved):
Nice :)

A few thoughts:
- I initially expected to be able to interact with charts when zooming out, especially that there is a blue outline when hovering over them,
- also at first I thought that the whole canvas area extends, so I could have more charts on the sides 😅 would it make sense to add canvas outline when zooming out, so it's more clear that the canvas area stays "fixed"?
- also a thought it we actually need 0.25x and 0.5x; maybe it also would make sense to have an "overview" button that would always show all charts and dynamically adjust the scale?

This is just to write things down, the implementation looks good 💯

ptbrowne:
Sorry should have been a draft PR. It was mostly to get feedback and allow @KerstinFaye to see it.

* Yes you should be able to click back on the chart to zoom in again, this is why I added the outline on the charts
* Good idea to have the canvas outline 👍 
* Yes I had a similar idea, seems good 👍

PR #1576, opened 2024-06-07
feat: Initialize layout step when editing saved dashboard

### LINKED ISSUE #1550: Edit saved dashboard - jump to layout options
After clicking edit on a saved dashboard in my profile arrive at the edit mode for all charts instead of the layout options step.

1. Create dashboard
2. Save dashboard
3. Click edit on the saved dashboard
4. I end up in the edit more of all charts instead of the layout options step. 

IArriving at the layout options step when revising a dashboard could enhance the comparison and decision-making process for chart revisions.

bprusinowski:
Closes #1550

### DIFF
--- a/app/configurator/configurator-state/context.tsx
+++ b/app/configurator/configurator-state/context.tsx
@@ -1,6 +1,6 @@
 import { PUBLISHED_STATE } from "@prisma/client";
 import { NextRouter, useRouter } from "next/router";
-import { createContext, Dispatch, useContext, useEffect, useMemo } from "react";
+import { Dispatch, createContext, useContext, useEffect, useMemo } from "react";
 import { useClient } from "urql";
 import { useImmerReducer } from "use-immer";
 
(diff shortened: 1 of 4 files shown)


app/configurator/configurator-state/context.tsx
```diff
@@ -271,7 +271,10 @@ export const ConfiguratorStateProviderInternal = (
           } else if (query.edit && typeof query.edit === "string") {
             newChartState = await initChartStateFromChartEdit(
               client,
-              query.edit
+              query.edit,
+              typeof query.overrideState === "string"
+                ? query.overrideState
```
ptbrowne, line 276: Nit: I think there is no need for "override", maybe simply "state" or "configuratorState"

ptbrowne (approved):
Nice 👍 I like that this approach gives us the flexibility also go to the CONFIGURATOR_CONFIGURING_CHART state  in the future if we need it. Or any other step :)

PR #1577, opened 2024-06-07
Chart data wrapper key

ptbrowne:
fix: No JoinBy dimension found in cube

When merging some datasets together, we ended up with a "no joinBy dimension
found in cube" error in chart hooks. It seems it is due to stale data
inside useQuery hooks. To prevent those, we unmount the whole ChartDataWrapper
and remount it when cubeIris and or joinBy change.

Fix https://github.com/visualize-admin/visualization-tool/issues/1475


bprusinowski (approved):
💯

PR #1578, opened 2024-06-07
chore: Use npx knip to clean up unused things

bprusinowski:


ptbrowne (approved):
Awesome I did not know this tool. If the tests pass and some manual testing is done around the app, with merging cubes, and publishing etc... then I think we are good to go.

bprusinowski:
Thanks 👍 It was not automated, I manually removed and checked every removed piece, but will still test if nothing is broken 🧪

bprusinowski:
Everything seems to work the same 😄

PR #1586, opened 2024-06-10
feat: Activate shared filters for "tab" layout

ptbrowne:
Activate dashboard filters for tab layout

- [x] In configurator mode
- [x] In published mode

### DIFF
--- a/app/components/chart-panel.tsx
+++ b/app/components/chart-panel.tsx
@@ -1,7 +1,7 @@
 import { Box, BoxProps, Theme } from "@mui/material";
 import { makeStyles } from "@mui/styles";
 import clsx from "clsx";
-import React, { forwardRef, HTMLProps } from "react";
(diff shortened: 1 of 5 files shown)


bprusinowski (commented):
Code looks good, but I think the state of the filters should be persisted when switching through the charts in the tab layout, otherwise the "shared" part doesn't really apply 😅 I can also take a look in the coming days, let me know 👍

bprusinowski:
Ah, I see that it's in fact persisted, it's just the slider that's initialized with the full range on switch 🤔

bprusinowski:
On a side note: I wonder if we shouldn't disable enter animations for charts, currently it's quite bad for a column chart (falls down from the top). This would also make it easier to compare the changes when switching thought the tabs, what do you think?

ptbrowne:
I'll take a look, thanks.

bprusinowski (approved):
🎉

PR #1593, opened 2024-06-11
Re-activate some tests

ptbrowne:
It seems they only needed a bit more time on some selectors.

I also removed tags from the network fixtures, this way we can add/remove tags without having to rename fixtures.


bprusinowski (approved):
Nice 🎉

PR #1614, opened 2024-06-13
Fix: cube query filters not found

ptbrowne:
In https://github.com/visualize-admin/visualization-tool/pull/1571, we  introduce the use of cube query filters, but this does not play well with tab layout when components are reused.

Some of the data hooks were reused, introducing a mismatch between what was fetched inside the hooks and the new chartConfig.

To force remounting, I used a key on ChartPublishedInner and made it so that we cannot forget it.

This fixes going to the second chart of https://test.visualize.admin.ch/en/v/FeByTTGfNhxL


bprusinowski (approved):
Thanks 🙇‍♂️

PR #1630, opened 2024-06-21
use `POST` to avoid `414 URI too long` for longer queries

Rdataflow:
@bprusinowski just caught by accident...
`GET` can in some cases result in `414 URI too long`

proposition:
apply `POST` as for the other queries :+1: - THX for your review


bprusinowski (approved):
@Rdataflow nice catch, thanks!

PR #1640, opened 2024-06-25
fix cube search performance

Rdataflow:
currently cube search queries for subthemes regardless of the organization filter 
- effectively there is no need for this information unless foen org is chosen

currently the subtheme graph isn't constrained (unless org is foen)
- this leads to degraded performance by needlessly crawling among graphs


Rdataflow:
@bprusinowski I hope you can review soon and hence we can retest...

PR #1653, opened 2024-07-08
fix: Caching per cube iri in geo queries

bprusinowski:


Rdataflow:
@bprusinowski thank you for this fix - I'll be able to check tonight...

Rdataflow:
@bprusinowski Caching endpoint is good for all queries now👍

PR #1662, opened 2024-07-17
fix: improve preview speed on Stardog >= 10

Rdataflow:
wrt Stardog issue 2094
see also SBAR-1122

@bprusinowski can you please push this urgent workaround to the upcoming release? THX
Context: The goal is to mitigate unusual long preview query times i.e. preview on Strompreise would take ~60s otherwise :/

cc @sosiology @adintegra

### DIFF
--- a/app/rdf/query-cube-preview.ts
+++ b/app/rdf/query-cube-preview.ts
@@ -110,7 +110,9 @@ CONSTRUCT {
     VALUES ?cube { <${iri}> }
     ?cube cube:observationConstraint/sh:property/sh:path ?observationPredicate .
     { SELECT ?observation ?observationPredicate ?observationValue ?observationLabel ?observationPosition WHERE {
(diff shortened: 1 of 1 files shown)


adintegra:
@Rdataflow thanks for reaching out. As you're aware, we currently don't have active resources on the project. Nonetheless, we'll try and get this workaround integrated before the upcoming PROD release.
/cc @sosiology

bprusinowski (approved):

Rdataflow:
@adintegra thanks we are aware - that's why I only ask for this minimal change 👍 
@bprusinowski THX 🎉

ortnever:
Thank you all (@Rdataflow @adintegra @bprusinowski ) for reacting so quickly to fix this performance problem. Deployment on PROD is scheduled for the first week of August, so this will be part of it. Great !

PR #1666, opened 2024-07-19
fix: Container size

### LINKED ISSUE #1665: Display the bar chart when a label is very long
When one of the labels is too long on the x axis, the height of the bar chart is shortened to the point where the graph is no longer visible.
@bprusinowski  : Do you see a solution to avoid this?
- Force a line break in the label text
and/or
- guarantee a minimum height for the chart
We could of course adapt the label text in the data, but this is sometimes difficult.
https://int.visualize.admin.ch/fr/v/YzDlIzcQKrT6

bprusinowski:
Fixes #1665

## How to reproduce
1. Go to [this link](https://int.visualize.admin.ch/en/v/YzDlIzcQKrT6?dataSource=Int).
2. ❌ See that the chart is really short.

## How to test
1. Go to [this link](https://visualization-tool-git-fix-container-size-ixt1.vercel.app/en/v/_sef-ZIvhwoc?dataSource=Prod).
2. ✅ See that the chart is displayed as expected.
3. ✅ Compare with the [same chart from Visualize PROD](https://visualize.admin.ch/en/v/IqjL1uuEeQPw?dataSource=Prod).

### DIFF
--- a/app/charts/area/areas-state.tsx
+++ b/app/charts/area/areas-state.tsx
@@ -23,7 +23,7 @@ import {
   useAreasStateVariables,
 } from "@/charts/area/areas-state-props";
 import {
-  getChartBounds,
+  useChartBounds,
   useChartPadding,
 } from "@/charts/shared/chart-dimensions";
 import {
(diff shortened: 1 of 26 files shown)


sosiology:
Tested on the PR and LGTM @bprusinowski thanks! 👏

PR #1677, opened 2024-07-25
feat: Categorical dashboard filters

### LINKED ISSUE #1669: Categorical Shared Filters
Add the possibility to include a global categorical filter in the dashboards similar to the temporal global filter.

bprusinowski:
Closes #1669

This PR introduces categorical dashboard filters.
- Each categorical filter displays all available dimension values; there's no cascading filters behavior, as every chart can have a different combination of other filters, which makes it impossible to have a "global cascading filter logic"
- Global categorical filters take precedence over local, chart-defined interactive filters, whose UI is disabled when global filters are active
- Cascading interactive filters behavior for chart is disabled when global categorical filters are enabled – in this mode we want to see no data, when a given combination actually yields no data, instead of reloading the chart filters and being out-of-sync with global filters.

## How to test
1. Go to [this link](https://visualization-tool-git-feat-categorical-dashboard-filters-ixt1.vercel.app/en/create/new?cube=https://environment.ld.admin.ch/foen/ubd003701/5&dataSource=Prod).
2. Add two more charts, so that there are three in total.
3. Modify the filters of every chart in a way that there's a different set of filters for every chart.
4. Proceed to layout options.
5. Switch to dashboard / free canvas layout.
6. See that there are four shared filters to be selected.
7. Toggle the filters and see that the select elements appeared on top of the canvas.
8. Use the filters to see that every chart adheres to global filters.
9. Toggle the filters off to see that the chart reverts back to its original filters.
10. You can also experiment with adding interactive filters to charts and seeing that they would be overwritten by global filters and that the state is reverted back to individual interactive filter after disabling global filter.

## To do in the future
- Refactor interactive filters so that they are scoped per cube. Currently if there are dimensions with the same iris across different cubes, we'd combine them and filter as one.

### DIFF
--- a/app/components/dashboard-interactive-filters.tsx
+++ b/app/components/dashboard-interactive-filters.tsx
@@ -1,8 +1,21 @@
-import { Slider, sliderClasses, useEventCallback } from "@mui/material";
+import {
+  Box,
+  BoxProps,
+  SelectChangeEvent,
+  Slider,
+  sliderClasses,
+  useEventCallback,
+} from "@mui/material";
 import { Theme } from "@mui/material/styles";
 import { makeStyles } from "@mui/styles";
-import { useEffect, useMemo, useState } from "react";
+import uniq from "lodash/uniq";
+import { ChangeEvent, useEffect, useMemo, useRef, useState } from "react";
 
+import {
+  DataFilterGenericDimension,
+  DataFilterHierarchyDimension,
+  DataFilterTemporalDimension,
+} from "@/charts/shared/chart-data-filters";
 import {
   DashboardTimeRangeFilter,
   hasChartConfigs,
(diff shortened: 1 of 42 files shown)


app/stores/interactive-filters.tsx
```diff
@@ -239,24 +232,13 @@ export const InteractiveFiltersProvider = ({
...
       stores,
     }),
-    [
-      potentialTimeRangeFilterIris,
-      timeRange,
-      potentialDataFilterIris,
-      dataFilters,
-      stores,
-    ]
+    [potentialTimeRangeFilterIris, potentialDataFilterIris, stores]
```
ptbrowne, line 241: question: Why this change ? Was it causing a problem ? I feel like from the consumer side, it makes it more complex and less flexible. I think I had "hidden" it behind the context so that we could be flexible in the context and every consumer do not have to worry where or how the filters are stored/computed.

bprusinowski: Discussed offline: it was to keep the separation between "static",  config filters and "dynamic", interactive filters more clear. We could still improve this by introducing e.g. `useConfigDashboardInteractiveFilters` and `useConfigInteractiveFilters` to hide the details of accessing them.

app/components/dashboard-interactive-filters.tsx
```diff
@@ -213,3 +238,180 @@ function stepFromTimeUnit(timeUnit: TimeUnit | undefined) {
...
+            } else {
+              delete dataFilters[componentIri];
+              store.setState({ dataFilters });
+            }
+          }
+        }
+      }
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [dimension]);
```
ptbrowne: Not sure about this.

- I do not think we should so tightly bind the component display and the store save/restore functionality
- I am not sure the component should be this knowledgeable about the store.
- Could the store actually provide functions that save/store this ?

I think I would be more happy with something like this

```

const { saveDataFilters, restoreDataFilters } = useDashboardInteractiveFilters()
useEffect(() => {
  const snapshot = saveDataFiltersSnapshot(componentIri)
  return () => {
      restoreDataFiltersSnapshot(componentIri)
  })

}, [dimensionIri])
```

Then I am still thinking that this should not be tied to UI but more to chartConfigs state. Actually it seems to me that useDashboardInteractiveFilters could maybe completely hide this (react on changes to the config to see if a data filters is toggled off/on and save/restore accordingly) ?

What do you think ?

app/components/dashboard-interactive-filters.tsx
```diff
@@ -213,3 +238,180 @@ function stepFromTimeUnit(timeUnit: TimeUnit | undefined) {
       return 1;
   }
 }
+
+const useDataFilterStyles = makeStyles((theme: Theme) => ({
+  wrapper: {
+    display: "grid",
+    gridTemplateColumns: "repeat(auto-fill, minmax(200px, 1fr))",
+    gap: theme.spacing(2),
```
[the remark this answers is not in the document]
bprusinowski: I think this was not changed in later commits, there are X columns that move elements to further rows if there's more of them than the number of columns – or did you mean something different? 🤔

app/components/dashboard-interactive-filters.tsx
```diff
@@ -213,3 +238,180 @@ function stepFromTimeUnit(timeUnit: TimeUnit | undefined) {
...
+    )
+  );
+
+  if (cubeIris.length > 1) {
+    console.error(
+      `Data filter ${componentIri} is used in multiple cubes: ${cubeIris.join(", ")}`
+    );
+  }
+
+  const cubeIri = cubeIris[0];
```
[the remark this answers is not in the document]
bprusinowski: I've added a TODO comment 👍 We already have a few places across the app that suffer from this issue, we should have it holistically solved as a next step for merging of cubes feature. As it's still in beta, I wouldn't put another error banner 🫣

app/components/dashboard-interactive-filters.tsx
bprusinowski: For a bit of context, here I also wanted to be consistent about how we handle this with dashboard time slider, where we also keep the "dashboard time filter state" inside the time slider, and just update the individual chart stores. Maybe the logic here got a bit too complex with saving / restoring things 😅 

I agree it would make it a bit easier to handle things by moving the state up for both areas, time filter and categorical filters to the dashboard state 👍

ptbrowne (approved):
LGTM 👍

app/components/dashboard-interactive-filters.tsx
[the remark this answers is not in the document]
ptbrowne: It wraps automatically due to auto-fill, my bad :)

Issue #1683, opened 2024-08-12
nightly job fails to preload the cache


Rdataflow:
**Describe the bug**
@bprusinowski I just noticed the nightly job fails to preload the charts to cache
https://github.com/visualize-admin/visualization-tool/actions/workflows/preload-varnish-cache.yml

thank you for fixing this issue

**To Reproduce** Steps to reproduce the behavior:

1. Go to https://github.com/visualize-admin/visualization-tool/actions/workflows/preload-varnish-cache.yml

**Expected behavior**
jobs preloading the cache successfully

**Environment (please complete the following information):**

- Visualize environment and version: GH repo at main branch

Rdataflow:
@ptbrowne not sure, thus asking: did the commit get this to work as expected?

Rdataflow:
@ptbrowne :tada:

PR #1684, opened 2024-08-12
fix: Do not completely fail if a chart cannot be loaded

ptbrowne:
We do not want to fail if a chart takes a long time to load. It could be
due to factors that we do not control like the availability of the database.
Here, we put a limit of 40s for a chart to load, and if it does not we 
issue an error in the console, but the script does not crash.


ptbrowne:
Should fix #1683

noahonyejese (approved):
Looks all fine to me!

PR #1694, opened 2024-08-27
fix: [ No title ] single charts

### LINKED ISSUE #1634: Name is not saved in Draft of a Visualization, even though the chart has a title
https://test.visualize.admin.ch/de/create/new?edit=MR9kR2D9K1Q5

1. I made a chart and gave a chart title
2. I saved the chart
3. Went to my visualizations - Draft does not show chart title

[Screen recording](https://www.dropbox.com/scl/fi/qln7yacahx2rg9ugb5qco/draftname-is-dashboard-name.mov?rlkey=4s0bse9v1t7oziv0bwl46jn5v&dl=0)

bprusinowski:
Fixes #1634

This PR fixes a problem with chart titles not being retrieved in the user profile for single charts. The problem was that we used a nullish coalescing operator operator (`??`) instead of a regular or (`||`), when layout title was an empty string by default. This led to actual chart title not being even considered to be displayed.

I've also added new transition strings for "single" and "dashboard" texts in user profile, and fixed rename dialog in a way that now it:
- updates chart title in case of single charts,
- updates layout title in case of multi-charts.

## How to test
This PR can be tested once merged and deployed to TEST by renaming a single chart.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ You can also check the
     mode
   - Table single filters now correctly display selected state
   - Drag handles are now centered in table configurator
+  - Single chart titles are now correctly displayed in user profile
+  - Renaming of charts through user profile now works correctly
 
 # [4.7.4] - 2024-07-23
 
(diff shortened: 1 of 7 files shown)


noahonyejese (approved):
Makes sense using ?? Should be used w/ caution anyway!

**Good** that you changed "d" to chartView!

PR #1698, opened 2024-08-29
fix: Delayed Tooltip for Save my Draft

noahonyejese:
Just increased the Tooltip delay by a few seconds, after clicking "Save Draft"

### DIFF
--- a/app/charts/combo/axis-height-linear-dual.tsx
+++ b/app/charts/combo/axis-height-linear-dual.tsx
@@ -35,8 +35,8 @@ export const AxisHeightLinearDual = (props: AxisHeightLinearDualProps) => {
   const otherAxisTitle = y[leftAligned ? "right" : "left"].label;
   const otherAxisTitleWidth = getTextWidth(otherAxisTitle, { fontSize: axisLabelFontSize }) + TICK_PADDING;
   const overLappingTitles = axisTitleWidth + otherAxisTitleWidth > bounds.chartWidth;
+  const overLappingAmount = (axisTitleWidth + otherAxisTitleWidth) / bounds.chartWidth
 
-  const largerTitleWidth = Math.max(axisTitleWidth, otherAxisTitleWidth);
 
   useRenderAxisHeightLinear(ref, {
     id: `axis-height-linear-${orientation}`,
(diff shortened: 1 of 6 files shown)


bprusinowski (approved):
👍

noahonyejese:
Legend now stacks on top of each other when the space gets titght and they start overlapping. The larger Legend Label gets stacked over the smaller Legend Label.

bprusinowski:
Or maybe just removing the behavior of "largest on top" and rather having "left on top" would make it a bit more clear. Maybe @KerstinFaye you also have some view on this?

noahonyejese:
I'm assuming you mean line breaks in the sense of the text? Also what approach is better just left on top? Or removing everything and using line breaks for the text?

bprusinowski:
I thought about something like this, so you could still keep the connection between the title and axis 👀 

Screenshot 2024-08-29 at 15 48 04

Then possibly it could also be applicable to other types of axes, to avoid situations like below in regular charts. What do you think? We can also wait for Kerstin's opinion on that before changing the approach 👍 

Screenshot 2024-08-29 at 15 46 47

KerstinFaye:
Thanks @noahonyejese for the approach. Adn thank you @bprusinowski for your comment. I didn't thought having them on top of each other is too big of an issue since also the colors are for dual axis charts an indicator of what belongs to which graph. But of course having a line-break in the label could also be solution since we could sort it to the according axis. My only concern with a line-break are long label names that can not really be divided. And also very narrow screen sizes where we would have to break the label over three lines.

bprusinowski:
@KerstinFaye yes, this could introduce another issue with too many line breaks... But the problem I mentioned was not about connecting the colors to lines, but rather about connecting the axis title to axis ticks; in the picture below, you can't tell if Leistung in kW has range of 0-4000 (left axis) or 0-100000 (right axis) – we'd need to also apply color to ticks in order to have this connection 🤔 In desktop mode, this is solved by always having "fixed" positions of the titles, just above the axis ticks.

362701056-fe9de4d3-55f3-4147-9223-9e75131106cf

KerstinFaye:
@bprusinowski ah, yes, thats absolutely true. Sorry. In this case line-break would definitely be the more clear solution. And I also realized that the labels have a rather big font size. We could go down to 12pt or for small screen even be 10pt. Linke this we might need less line-breaks as well.

noahonyejese:
> @bprusinowski ah, yes, thats absolutely true. Sorry. In this case line-break would definitely be the more clear solution. And I also realized that the labels have a rather big font size. We could go down to 12pt or for small screen even be 10pt. Linke this we might need less line-breaks as well.

Alright I will implement this change, asap

bprusinowski (approved):
Thanks @noahonyejese, much better now 🎉 

I'd increase the font size of the legend to 12px however, to align with the size of footer elements, I think it's too small now 👀 
Screenshot 2024-09-02 at 14 48 34

And also it could be nice to apply the same behavior to other legend types, not only combo-dual (but this could be a future PR) 🥹

noahonyejese:
> Thanks @noahonyejese, much better now 🎉
> 
> I'd increase the font size of the legend to 12px however, to align with the size of footer elements, I think it's too small now 👀 Screenshot 2024-09-02 at 14 48 34
> 
> And also it could be nice to apply the same behavior to other legend types, not only combo-dual (but this could be a future PR) 🥹

Alright I will increase the font to 12 px and merge 👍 Yes we could implement it across all the charts in a future PR

bprusinowski:
Hi @noahonyejese, it looks like the PR was closed and not merged? Was this intended? 👀

noahonyejese:
> Hi @noahonyejese, it looks like the PR was closed and not merged? Was this intended? 👀

No it was not my bad.. I thought I merged 😅

noahonyejese:
@bprusinowski Do you know why so many imports are missing suddenly?

bprusinowski:
@noahonyejese it looks like you have an outdated state now in the PR, as e.g. the `SaveDraftButton` is still in `chart-selection-tabs.tsx` in your branch, but on main it was moved to `configurator.tsx` around two weeks ago 👀

bprusinowski:
The same with NextStepButton, I think you can remove them from your branch, as it was also already moved to `configurator.tsx`. Not 100% sure how it got to this state, maybe `main` should have been rebased, and not merged?

noahonyejese:
> The same with `NextStepButton`, I think you can remove them from your branch, as it was also already moved to `configurator.tsx`. Not 100% sure how it got to this state, maybe `main` should have been rebased, and not merged?

I'm not extremely familiar with github rebase and I certainly don't want to break anything. Should I close this PR again?

bprusinowski:
@noahonyejese I pulled the branch to my machine, un-did the merge commit and rebased the branch onto `main` – for some reason the chart tabs file was showing up as a conflict, but should be safe to merge if CI passes, as the only changes visible in the PR now are scoped to the axis titles 👍

noahonyejese:
> @noahonyejese I pulled the branch to my machine, un-did the merge commit and rebased the branch onto `main` – for some reason the chart tabs file was showing up as a conflict, but should be safe to merge if CI passes, as the only changes visible in the PR now are scoped to the axis titles 👍 

Thanks @bprusinowski

PR #1708, opened 2024-09-04
feat: Improve layouting step

bprusinowski:
This PR:
- makes the page scroll to a shared filter when adding or removing the filters in the layouting step, so that users won't miss the fact that they were added at the top of the viewport,
- makes the application remember past layouts when switching between them in the layouting step.

The latter could technically still be improved in a way that a previous layout is also kept when switching to editing step and back to layouting, but as I was unsure if this scenario of:
1. Spending a lot of time doing free canvas layout.
2. Switching to a tab layout.
3. Going back to editing mode.
4. Switching the layout back to free canvas layout.
would be very common, so I went with less complex implementation for now.

## How to test
1. Go to this link.
2. Add a new chart.
3. Proceed to layout options.
4. Make the window small, and scroll to the bottom of the page.
5. Activate a shared filter.
6. ✅ See that the page was scrolled to the top.
7. Switch to a dashboard / free canvas layout.
8. Change the positions of charts.
9. Toggle through dashboard sub-layouts (vertical, tall and free canvas).
10. ✅ See that the layout was persisted.
11. Switch to a tab layout, single URLs and dashboard layout again.
12. ✅ See that the layout was persisted.

### DIFF
--- a/app/components/dashboard-interactive-filters.tsx
+++ b/app/components/dashboard-interactive-filters.tsx
@@ -9,7 +9,7 @@ import {
 import { Theme } from "@mui/material/styles";
 import { makeStyles } from "@mui/styles";
 import uniq from "lodash/uniq";
-import { ChangeEvent, useEffect, useMemo, useState } from "react";
+import { ChangeEvent, useEffect, useMemo, useRef, useState } from "react";
 
 import {
   DataFilterGenericDimension,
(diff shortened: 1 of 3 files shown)


KerstinFaye:
That's great. Thanks @bprusinowski! I think for now we can really leave it like it is. When the user would go back to editing and then went again to layout step and switches to anther option, I think this could be kept in mind for a later point. But for now what you implemented is already fantastic.

noahonyejese (approved):
LGTM, worked as described. 🎉

PR #1724, opened 2024-09-05
feat: Preview of charts in different screen sizes

### LINKED ISSUE #1710: Free Canvas Layout: Quickly identify the minimum size of the canvas
**Problem:** 
The minimum width of the diagrams in the canvas is difficult to understand. It varies depending on the screen size. Usually only two diagrams fit next to each other. Users are confused that the layout looks different in edit mode than in the published status (more diagrams next to each other). 

**Improvement:**
- Option to ‘Preview the published view’.
- Additional breakpoint so that 3 charts can also be placed next to each other.

Design:
[Interactive prototype](https://www.figma.com/proto/Hn9wvuEYLUmdJtg4KwZYUx/visualize.admin.ch---design?page-id=8%3A145669&node-id=5235-123489&node-type=FRAME&viewport=1199%2C369%2C0.13&t=V5EtBLKMUERyDo6g-1&scaling=scale-down&content-scaling=fixed&starting-point-node-id=5235%3A123489&show-proto-sidebar=1)

Image

Image

bprusinowski:
Closes #1710.

This PR adds an option to preview every layout (tab, dashboard and single URLs) in three modes:
- full available window width,
- tablet width (`md` breakpoint),
- mobile width (`sm` breakpoint).

## How to test
1. Go to this link.
2. Add a few more charts.
3. Proceed to layout options.
4. ✅ See that there's a new toggle menu in the right top corner with device icons.
5. ✅ Click through them to see that the chart area adjusts accordingly – you can toggle off the preview by clicking on a currently select preview mode, or by `>>` button in the left top corner.
6. Switch to dashboard / free canvas layout and modify it.
7. Preview different viewport sizes and exit the preview mode.
8. ✅  See that the original layout was persisted.

## Next steps
We should probably persist the changes made in the preview mode, so that users can not only modify largest viewport layout (`lg`), but rather also smaller ones. Currently they are just preview, without an option to be saved.

### DIFF
--- a/app/components/chart-preview.tsx
+++ b/app/components/chart-preview.tsx
@@ -120,11 +120,6 @@ const snapCornerToCursor = createSnapCornerToCursor({
   xOffset: -DRAG_OVERLAY_WIDTH,
 });
 
-type DashboardPreviewProps = ChartPreviewProps & {
-  layoutType: Extract<Layout, { type: "dashboard" }>["layout"];
-  editing?: boolean;
-};
-
 const useStyles = makeStyles(() => ({
   canvasChartPanelLayout: {
     // Provide some space at the bottom of the canvas layout to make it possible
@@ -133,7 +128,12 @@ const useStyles = makeStyles(() => ({
   },
 }));
 
-const DashboardPreview = (props: DashboardPreviewProps) => {
+const DashboardPreview = (
+  props: ChartPreviewProps & {
+    layoutType: Extract<Layout, { type: "dashboard" }>["layout"];
+    editing?: boolean;
+  }
+) => {
   const { dataSource, layoutType, editing } = props;
   const [state, dispatch] = useConfiguratorState(hasChartConfigs);
   const theme = useTheme();
--- a/app/components/react-grid.tsx
+++ b/app/components/react-grid.tsx
@@ -237,7 +237,7 @@ export const ChartGridLayout = ({
     setMounted(true);
   }, []);
 
-  const hasBeenConstrained = useRef(false);
+  const hasBeenConstrained = useRef(!initialize);
   useEffect(() => {
     if (!mountedForSomeTime || hasBeenConstrained.current) {
       return;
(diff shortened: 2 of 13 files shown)


noahonyejese (approved):
LGTM, nice refactoring, I felt like that was required.

noahonyejese (commented):
You could add a function like this to maybe consolidate the dispatch logic, since you are calling it quite a few times: 

```
const updateLayout = (layoutValue: Breakpoint | null) => {
  dispatch({
    type: "LAYOUT_CHANGED",
    value: {
      ...state.layout,
      activeField: undefined,
      layout: layoutValue,
    },
  });
};

```

noahonyejese (commented):
Maybe using a useMemo to avoid recalc. on each render for calculating the maxWidth 

`const computedMaxWidth = useMemo(() => {
  if (isSingleURLs) {
    return { xs: "100%", lg: 1280 };
  }
  return maxWidth && maxWidth !== "lg"
    ? (t) => t.breakpoints.values[maxWidth]
    : "100%";
}, [isSingleURLs, maxWidth]);`

bprusinowski:
Thanks for reviewing @noahonyejese! I've share the common part of the layout update logic in https://github.com/visualize-admin/visualization-tool/pull/1724/commits/7d67053496bee6b195b1b3016a0856557aa17580; for the `useMemo` in maxWidth, this was done in later commits 👍

noahonyejese:
> Thanks for reviewing @noahonyejese! I've shared the common part of the layout update logic in [7d67053](https://github.com/visualize-admin/visualization-tool/commit/7d67053496bee6b195b1b3016a0856557aa17580); for the `useMemo` in maxWidth, this was done in later commits 👍

Ah nice, my bad!

KerstinFaye:
@bprusinowski this is absolutely fantastic! 🤩 so cool we have this now.

sosiology:
This is very awesome. Thanks @bprusinowski 

@KerstinFaye i think we need to test it carefully. At the moment if i check the mobile preview it shows me this layout 
image

But when i check the chart on my phone i see this which is slightly different from the preview. 

https://github.com/user-attachments/assets/9488140e-1b45-42dd-bb29-5b7488a13dae


Tested with this dashboard: https://test.visualize.admin.ch/de/v/E4CMla1F2Suz

PR #1725, opened 2024-09-06
fix: DashboardTimeRangeSlider

### LINKED ISSUE #1617: Dashboard time filter's last option is not displayed correctly
See that the last option in dashboard shared time filter has less "padding" than it should have e.g. [here](https://test.visualize.admin.ch/en/v/6YW3HU5Tq2Q9?dataSource=Int). This is related to the incorrect `step` calculation, which could be fixed to make sure the spacing is consistent.

bprusinowski:
Fixes #1617

This PR fixes a problem of wrong positioning of marks in `DashboardTimeRangeSlider`. Due to the conversion of dates to milliseconds, and interpolation of values between min and max dates, there were some non-obvious issues with marks positioning. To see one such issue, see [this link](https://test.visualize.admin.ch/en/v/ZJSPzC4KHBvM?dataSource=Prod) from TEST and slide through the slider values – see that 2022 and 2023 and extra-close to each other, while they should have the same spacing as previous years.

To fix this issue, I changed the Slider's behavior from continuous to discrete, passing the values that actually appear in the data. It also makes sure that we use combined values from merged date dimension.

## How to test
1. Go to [this link](https://visualization-tool-git-fix-time-range-dashboard-slider-ixt1.vercel.app/en/create/new?cube=https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/10&dataSource=Prod).
2. Add a line chart.
3. Proceed to layout options.
4. Enable shared date filter.
5. ✅ Scroll through the values and see that 2022 and 2023 have equal spacing, as between other years.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@ You can also check the
   - Renaming of charts through user profile now works correctly
   - Manually entering dates in date pickers works correctly again
   - Improved scrolling behavior in the chart tabs
+  - Shared time range filter now correctly positions the marks in the time
+    slider
 - Style
   - Aligned editor and layouting page layouts
   - Removed dimension selection from modal when merging cubes
--- a/app/charts/shared/use-combined-temporal-dimension.ts
+++ b/app/charts/shared/use-combined-temporal-dimension.ts
@@ -1,0 +1,87 @@
+import { t } from "@lingui/macro";
+import { ascending, descending } from "d3-array";
+import uniqBy from "lodash/uniqBy";
+import { useMemo } from "react";
+
+import {
+  hasChartConfigs,
+  useConfiguratorState,
+} from "@/configurator/configurator-state";
+import {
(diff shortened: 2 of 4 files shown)


noahonyejese (commented):
Isn't the dimensions array already memoized via the first useMemo hook? i might be wrong on this one...

noahonyejese (approved):
LGTM ✔️

bprusinowski:
@noahonyejese yep, technically the dimensions should come memoized from the first hook, except when it's still fetching the data; in this case the result could be null, and using `?? []` would force re-calculations of second useMemo on every render 👀 I've moved this part of the logic to second `useMemo`, I think it's cleaner now, thanks!

noahonyejese (approved):
👍

KerstinFaye:
@bprusinowski, looks good. Thanks

PR #1726, opened 2024-09-06
fix: Search bots behavior

### LINKED ISSUE #1676: Enable site indexing in Production
**Describe the bug**
Currently, the Visualize production site includes meta directives which discourage it from being indexed by search engines.

**To Reproduce**

1. Go to https://visualize.admin.ch
2. Open Dev Tools and check the `X-Robots-Tag` HTTP header

visualize_seo

**Expected behavior**
In the production environment the [X-Robots-Tag](https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag#xrobotstag) header should not be set.

**Environment**

- Visualize environment and version: PROD 4.0.0

**Additional context**
The functionality for the correct behaviour is already implemented and is controlled by the presence of the environment variable `ALLOW_SEARCH_BOTS`. However, this doesn't appear to be correctly evaluated at runtime.

bprusinowski:
Fixes #1676

It looks like we had a legacy variable – `ALLOW_SEARCH_BOTS` – that was not passed via the Dockerfile and couldn't be picked up during build. The exact rule we had in the code was "if ALLOW_SEARCH_BOTS is not 'true'", then disable the indexing, which resulted in the problem described in #1676.

I think it makes more sense to enable indexing by default, and disable it conditionally; that's why there's a new, `PREVENT_SEARCH_BOTS` variable to disable this behavior where needed.

## Todo
- [ ] Add `PREVENT_SEARCH_BOTS=true` env variable to Vercel
- [ ] Add `PREVENT_SEARCH_BOTS=true` env variable to GitLab TEST pipeline
- [ ] Reach out to VSHN to add `PREVENT_SEARCH_BOTS=true` variable to INT pipeline
- [ ] PROD doesn't need a variable at all, once we deploy the behavior should be right due to reversed condition

cc @adintegra

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,7 @@ WORKDIR /usr/src/app
 
 # build with
 # docker build \
+#   --build-arg PREVENT_SEARCH_BOTS=<true/false> \
 #   --build-arg COMMIT=$(git rev-parse HEAD) \
 #   --build-arg VECTOR_TILE_URL=<url of the vector service> \
 #   --build-arg MAPTILER_STYLE_KEY=<maptiler style key> \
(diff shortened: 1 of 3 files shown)


noahonyejese (approved):
LGTM 👍

PR #1727, opened 2024-09-06
fix: Server-side eIAM session termination

### LINKED ISSUE #1691: Cleanly terminate eIAM session
While logging out, the application currently makes a POST request to https://visualize.admin.ch/api/auth/signout. While this disconnects the session application-side, it doesn't initiate clean termination on the eIAM side. In order to ensure this, Visualize should call the SLO endpoint – this will terminate the eIAM session cleanly server-side.

Please use the official `end_session_endpoint` from the metadata:

- TEST: feds-r.eiam.admin.ch/realms/uvek_bafu-visualise/.well-known/openid-configuration
- INT: feds-a.eiam.admin.ch/realms/uvek_bafu-visualise/.well-known/openid-configuration
- PROD: feds.eiam.admin.ch/realms/uvek_bafu-visualise/.well-known/openid-configuration

For PROD then for example: https://identity-eiam.eiam.admin.ch/realms/uvek_bafu-visualize/protocol/openid-connect/logout

**Expected behavior**
The eIAM session is cleanly terminated.

bprusinowski:
Fixes #1691

This PR fires a request to eIAM servers when user signs out, to also terminate the session on their side.

As we can't connect to eIAM infrastructure from localhost, I wasn't able to test out if this works – we would be able to do so when the change is deployed to TEST.

It would be extremely helpful to be able to connect to e.g. TEST eIAM auth server to work on such issues. @adintegra @ptbrowne do you know if we have such access, maybe I missed this? If not, could we potentially request it?

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,8 @@ yarn dev:ssl # If you are working with the login process
 > experience an SSL error when the authentication provider redirects you back to
 > the app after login. You can either remove the trailing s in the URL after the
 > redirection, or use the `yarn dev:ssl` command to use HTTPs for the
-> development server.
+> development server. Also, make sure to set the `NEXTAUTH_URL` environment
+> variable to `https://localhost:3000` in your `.env.local` file.
 
 > 👉 In [Visual Studio Code](https://code.visualstudio.com/), you also can run
 > the **default build task** (CMD-SHIFT-B) to start the dev server, database
@@ -350,6 +351,9 @@ We use Next-auth to integrate our application with it, through a
 You can use the ref eIAM. ADFS environment variables should be configured in
 your `.env.local` file. You'll find those secret variables in our shared
 1Password in the "Visualize.admin .env.local" entry.
+
+Make sure to set the `NEXTAUTH_URL` environment variable to
+`http://localhost:3000` and run the dev server with `yarn dev:ssl`.
 
 ## 10. <a name='Translations'></a>Translations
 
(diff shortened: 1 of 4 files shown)


adintegra:
> It would be extremely helpful to be able to connect to e.g. TEST eIAM auth server to work on such issues

@bprusinowski localhost should be whitelisted in eIAM TEST so should be testable. From your description, I'm not 100% sure if this is the configuration you tested against (https://feds-r.eiam.admin.ch/realms/uvek_bafu-visualize/.well-known/openid-configuration)?

noahonyejese (approved):
Generally the authentication needs to be more easily accessable during dev.

bprusinowski:
@adintegra I believe so, I used the same env variables that are defined in GitLab that deploys to TEST. Are you able to authenticate successfully from your machine?

adintegra:
> Are you able to authenticate successfully from your machine?

@bprusinowski yes, it's testable. What's critical is this var (which is different from the deployment environment):

`NEXTAUTH_URL=https://localhost:3000`

Also, ensure you run `npm run dev:ssl` to get SSL support :)

bprusinowski:
Thanks @adintegra 💯 I'll add this to README, I missed this part 🥲

adintegra:
Thanks @bprusinowski! Yes, the documentation on the specifics was a bit sparse – sorry for that.

bprusinowski:
No problem @adintegra! In fact, we already had the `yarn dev:ssl` in the docs, but missed `NEXTAUTH_URL` in `.env.development`. I've updated both to mention this 👍

PR #1729, opened 2024-09-09
feat: Login & UI improvements

bprusinowski:
This PR:
- updates last edit timestamp to match the design (removed seconds),
- added tooltip to chart titles in user profile,
- aligned chart selection popover styles to match the design,
- fixes hard-refresh when using action menu items,
- adds an on-click action to chart titles in user profile to open or edit the chart, depending on its publication status.


noahonyejese (approved):
Nice!

PR #1734, opened 2024-09-09
fix: Minimum dashboard chart sizes

### LINKED ISSUE #1732: Improve minimum size of chart
The minimum size of the charts should be improved again. 
https://test.visualize.admin.ch/create/c-iMHGERXZRr?dataSource=Prod

1. Started with the dataset: Marktzahlen Milch und Milchprodukte - Monatliche Konsumentenpreise
2. Duplicated the chart and added two charts based on the same dataset
3. Changed to Dashboard - Free canvas

Got this result: 
Image

Another dashboard looks also off in the published mode: 
https://test.visualize.admin.ch/create/_JQcxuZQPg4f?dataSource=Prod

Image

bprusinowski:
Fixes #1732

This PR:
- fixes some issues with tab labels in layouting step,
- fixes computations of minimal height of free canvas when adding a new chart to an existing layout,
- increases minimal height of free canvas chart from 100 to 150 pixels, which should help with wrong initializations in some cases,
- improves Table appearance in free canvas layout (correct initial sizing and non-overflowing content).

There's still a potential to improve the table chart in free canvas further, as it looks that compact view doesn't work as expected in small sizes and scrollbar position doesn't seem right, but this can be done in another PR, as it seems like a table issue, and not free canvas-related problem.

## How to test
1. Go to [this link](https://visualization-tool-git-fix-minimum-dashboard-chart-sizes-ixt1.vercel.app/en/create/new?cube=https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/10&dataSource=Prod).
2. Add a new table chart.
3. ✅ See that the active tab's chart icon is blue, and inactive icon is black.
4. Proceed to layout options.
5. ✅ See that the Label tab is not available under the Title & Description section in the sidebar.
6. ✅ See that the chart icons are visible in the tabs.
7. Switch to a dashboard / free canvas layout.
8. ✅ See that the charts were resized correctly.
9. Go back to the editing mode.
10. Add a new, pie chart.
11. Proceed to layout options.
12. ✅ See that the pie chart was correctly initialized.

## How to reproduce
1. Go to [this link](https://test.visualize.admin.ch/en/create/new?cube=https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/10&dataSource=Prod&flag__layouter.dashboard.free-canvas=true).
2. Add a new table chart.
3. ❌ See that the tabs are both gray.
4. Proceed to layout options.
5. ❌ See that the Label tab is available under the Title & Description section in the sidebar.
6. ❌  See that the chart icons are missing from the tabs.
7. Switch to a dashboard / free canvas layout.
8. See that the charts were not resized correctly (e.g. table is not visible at all).
9. Go back to the editing mode.
10. Add a new, pie chart.
11. Proceed to layout options.
12. ❌ See that the pie chart was not correctly initialized.

### DIFF
--- a/app/charts/shared/containers.tsx
+++ b/app/charts/shared/containers.tsx
@@ -15,6 +15,7 @@ import { useTransitionStore } from "@/stores/transition";
 
 export const useStyles = makeStyles<{}, {}, "chartContainer">(() => ({
   chartContainer: {
+    overflow: "hidden",
     position: "relative",
     width: "100%",
     flexGrow: 1,
--- a/app/charts/table/chart-table.tsx
+++ b/app/charts/table/chart-table.tsx
@@ -1,6 +1,7 @@
 import { memo } from "react";
 
 import { ChartDataWrapper } from "@/charts/chart-data-wrapper";
+import { ChartContainer } from "@/charts/shared/containers";
 import { Table } from "@/charts/table/table";
 import { TableChart } from "@/charts/table/table-state";
 import { TableConfig } from "@/configurator";
@@ -28,7 +29,9 @@ export const ChartTableVisualization = (
 const ChartTable = memo(function ChartTable(props: ChartProps<TableConfig>) {
   return (
     <TableChart {...props}>
-      <Table />
+      <ChartContainer>
+        <Table />
+      </ChartContainer>
     </TableChart>
   );
 });
--- a/app/charts/table/table.tsx
+++ b/app/charts/table/table.tsx
@@ -10,18 +10,17 @@ import {
   useSortBy,
   useTable,
 } from "react-table";
+import AutoSizer from "react-virtualized-auto-sizer";
 import { FixedSizeList, VariableSizeList } from "react-window";
 
 import { useChartState } from "@/charts/shared/chart-state";
 import { CellDesktop } from "@/charts/table/cell-desktop";
 import { DDContent } from "@/charts/table/cell-mobile";
-import { TABLE_HEIGHT } from "@/charts/table/constants";
 import { GroupHeader } from "@/charts/table/group-header";
 import {
   TableContent,
   TableContentProvider,
 } from "@/charts/table/table-content";
-import { scrollbarWidth } from "@/charts/table/table-helpers";
 import { TableChartState } from "@/charts/table/table-state";
 import Flex from "@/components/flex";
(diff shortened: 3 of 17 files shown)


noahonyejese (commented):
You are right about the free Canvas view there is still a bit of room for improvement.

noahonyejese:
Nit: Quite a lot commits, you could potentially split them up a bit? Other than that it looks fine!

noahonyejese (approved):

noahonyejese:
👍🏽 not too bad

KerstinFaye:
thanks a lot @bprusinowski. It so much improved already. that's fantastic. 
I saw something very small. Let's call the section under title & description "Tab label", instead "Label". I think this would be a bit more clear. (I could not identify the correct position to change this in Accent).

Another thing that I discovered in the free canvas step is, when increasing one of the charts (tested it with the table chart), and then clicking through the preview buttons, and after getting the sidebar to edit again, the chart is always jumping back to the minimum height:

https://github.com/user-attachments/assets/67a66dcb-64bd-43a8-b560-87cc3d1945ee

bprusinowski:
Hi @KerstinFaye, thanks for testing and feedback, I'll change the label 👍 

For the issue with going back to minimum height, in fact it goes back to original height before entering the preview. I think we could either disable an ability to move the charts around and resize them while in preview mode (quick thing to do, maybe good anyway to not confuse the users?) or make it possible to edit the layouts when in preview mode, currently it's not yet implemented :/

Let me know what would make more sense from your side, I believe it could be nice to be able to edit while in preview mode to be able to e.g. optimize the mobile layout. But I'd do this in another PR 👀

KerstinFaye:
@bprusinowski yes, I really like the ability to change the layout while in preview mode. But would it be possible to save the height and width that was set for the charts when changing between the preview buttons? So when going back to the maximum size the user sees the height that they set before?

bprusinowski:
@KerstinFaye I think we could keep different layout configurations at all times, currently we only keep `lg` when on a big screen, and lower ones are generated and removed "on the fly", as needed, that's why we lose the settings.

In this case I'll focus on improving that in a next PR :)

PR #1744, opened 2024-09-12
feat: Allow editing free canvas in preview mode

### LINKED ISSUE #1739: Allow editing free canvas dashboard layout in preview mode
Currently, we are able to preview free canvas dashboard layout in different screen sizes, but the changes applied to it in a preview mode are not persisted on exiting it – we should change this behavior to save such changes.

bprusinowski:
Closes #1739

This PR:
- saves the changes made while previewing free canvas layout in different breakpoints, and keeps the layouts separate for each breakpoint (`xl`, `lg`, `md`, `sm`),
- makes the preview mode size reflect the true size of the breakpoint, making the area horizontally scrollable when the screen size if lower than the breakpoint that's being previewed,
- adds a new preview breakpoint (`xl`) and aligns all of them with free canvas layout breakpoints.

## How to test
1. Go to [this link](https://visualization-tool-git-feat-allow-editing-free-canv-12397f-ixt1.vercel.app/en/create/new?cube=https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/10&dataSource=Prod).
2. Add two new charts.
3. Proceed to layout options.
4. ✅ See that there are four options in the preview menu (up from three).
5. Switch to a dashboard / free canvas layout.
6. ✅ Enable preview mode and see that the first one shows 4 columns, the next one 3, then 2 and 1.
7. ✅ Edit the dashboard in different preview modes and see that the changes are persisted between switches.
8. Make your browser window smaller.
9. Enable first (`xl`) preview mode.
10. ✅ See that the area still shows 4 columns and is horizontally scrollable, which allows users with smaller screens prepare their dashboards for larger ones when they're embedded.

### DIFF
--- a/app/configurator/components/preview-breakpoint.tsx
+++ b/app/configurator/components/preview-breakpoint.tsx
@@ -9,15 +9,13 @@ import {
 } from "@mui/material";
 import { makeStyles } from "@mui/styles";
 import clsx from "clsx";
-import { useEffect, useMemo, useRef, useState } from "react";
+import { ReactNode, useMemo, useState } from "react";
 
-import {
-  isLayouting,
-  useConfiguratorState,
-} from "@/configurator/configurator-state";
+import { FREE_CANVAS_BREAKPOINTS } from "@/components/react-grid";
 import { Icon, IconName } from "@/icons";
+import { theme } from "@/themes/federal";
 
-type PreviewBreakpoint = "lg" | "md" | "sm";
+type PreviewBreakpoint = keyof typeof FREE_CANVAS_BREAKPOINTS;
 
 export const usePreviewBreakpoint = () => {
   const [breakpoint, setBreakpoint] = useState<PreviewBreakpoint | null>(null);
@@ -32,7 +30,7 @@ export const PreviewContainer = ({
   breakpoint,
   singleColumn,
 }: {
-  children: React.ReactNode;
+  children: ReactNode;
   breakpoint: PreviewBreakpoint | null;
   singleColumn?: boolean;
 }) => {
@@ -41,18 +39,20 @@ export const PreviewContainer = ({
   const maxWidth = useMemo(() => {
     if (breakpoint) {
       switch (breakpoint) {
+        case "xl":
+          return breakpoints.values.lg - 1;
         case "lg":
-          return `calc(100% - ${PREVIEW_CONTAINER_PADDING}px)`;
+          return breakpoints.values.md - 1;
         case "md":
-          return `min(calc(100% - ${PREVIEW_CONTAINER_PADDING}px), ${breakpoints.values.md}px)`;
+          return breakpoints.values.sm - 1;
         case "sm":
-          return 480;
+          return FREE_CANVAS_BREAKPOINTS.md;
         default:
           const _exhaustiveCheck: never = breakpoint;
           return _exhaustiveCheck;
       }
     }
-    return singleColumn ? SINGLE_COLUMN_MAX_WIDTH : "100%";
+    return singleColumn ? theme.breakpoints.values.lg - 1 : "100%";
   }, [breakpoint, breakpoints, singleColumn]);
   return (
     <div className={classes.container} style={{ maxWidth }}>
@@ -75,11 +75,19 @@ export const PreviewBreakpointToggleMenu = ({
     title: string;
   }[] = [
     {
-      breakpoint: "lg",
+      breakpoint: "xl",
       iconName: "desktop",
       title: t({
+        id: "controls.layout.preview-xl",
+        message: "Preview using extra large width",
+      }),
+    },
+    {
+      breakpoint: "lg",
+      iconName: "laptop",
+      title: t({
         id: "controls.layout.preview-lg",
-        message: "Preview using available width",
+        message: "Preview using large width",
       }),
     },
     {
--- a/app/configurator/components/preview-breakpoint.tsx
+++ b/app/configurator/components/preview-breakpoint.tsx
@@ -20,23 +20,12 @@ import { Icon, IconName } from "@/icons";
 type PreviewBreakpoint = "lg" | "md" | "sm";
 
 export const usePreviewBreakpoint = () => {
-  const [state] = useConfiguratorState(isLayouting);
   const [breakpoint, setBreakpoint] = useState<PreviewBreakpoint | null>(null);
-  const layoutRef = useRef(state.layout);
-  useEffect(() => {
-    if (!breakpoint) {
-      layoutRef.current = state.layout;
-    }
-  }, [breakpoint, state.layout]);
   return {
     previewBreakpoint: breakpoint,
     setPreviewBreakpoint: setBreakpoint,
-    previewBreakpointLayout: layoutRef.current,
   };
 };
-
-const SINGLE_COLUMN_MAX_WIDTH = 1280;
-const PREVIEW_CONTAINER_PADDING = 216;
 
 export const PreviewContainer = ({
   children,
--- a/app/configurator/components/preview-breakpoint.tsx
+++ b/app/configurator/components/preview-breakpoint.tsx
@@ -36,7 +36,7 @@ export const PreviewContainer = ({
 }) => {
   const classes = useStyles();
   const { breakpoints } = useTheme();
-  const maxWidth = useMemo(() => {
+  const width = useMemo(() => {
     if (breakpoint) {
       switch (breakpoint) {
         case "xl":
@@ -55,8 +55,16 @@ export const PreviewContainer = ({
     return singleColumn ? theme.breakpoints.values.lg - 1 : "100%";
   }, [breakpoint, breakpoints, singleColumn]);
   return (
-    <div className={classes.container} style={{ maxWidth }}>
-      {children}
+    <div
+      className={classes.container}
+      style={{ marginTop: breakpoint ? "3rem" : 0 }}
+    >
+      <div
+        className={classes.chartContainer}
+        style={{ width, paddingTop: breakpoint ? "0.5rem" : 0 }}
+      >
+        {children}
+      </div>
     </div>
   );
 };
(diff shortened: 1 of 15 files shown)


app/locales/en/messages.po
```diff
@@ -647,6 +647,10 @@ msgstr "Preview using medium width"
 msgid "controls.layout.preview-sm"
 msgstr "Preview using small width"
 
+#: app/configurator/components/preview-breakpoint.tsx
+msgid "controls.layout.preview-xl"
+msgstr "Preview using extra large width"
```
ptbrowne, line 652: I think it would be easier for the user to see 

Preview using mobile
Preview using tablet
Preview using small desktop screen
Preview using large desktop screen

app/configurator/components/preview-breakpoint.tsx
```diff
@@ -149,7 +157,14 @@ export const PreviewBreakpointToggleMenu = ({
...
   container: {
     width: "100%",
+    overflowX: "auto",
+    // Show the scrollbar on top
+    transform: "rotateX(180deg)",
+  },
+  chartContainer: {
     margin: "0 auto",
+    // Show the scrollbar on top
+    transform: "rotateX(180deg)",
```
ptbrowne, line 167: I do not understand this trick, what is it about ?

app/locales/en/messages.po
bprusinowski: cc @KerstinFaye as these can be now edited via Accent (I used the ones from the design) – but agree that it would be more obvious 👍

app/configurator/components/preview-breakpoint.tsx
bprusinowski: It makes the scrollbar appear on top, otherwise it could sometimes be hard to understand that the area is scrollable, as the scrollbar would be outside of viewport as the bottom 👀 It's only scoped to the preview mode though, as in published mode the chart adapts to the viewport width :)

top scrollbar

bprusinowski: So with double flip, everything looks the same, but the container scrollbar's position is reversed 😄

ptbrowne:
I find the behavior of the preview buttons a bit confusing. 

- When I click on the left arrow when the panel is hidden, to reshow the panel, I am out of the preview mode, and I find it confusing since the UI elements are so far away from each other on the screen. 
- Maybe the fact that we are "in preview mode" should be more explicit.
- I can open the left panel by clicking on the title, but then I cannot close it
- Would it be possible to show the extent of the screen + the width in pixel ?

On the coding size, nothing stood out as blocking. I think it would be good to add tests, particularly around the "initialized" behavior, and around the changing layouts responsively and checking what happens.


I wonder as a user if we should mark layouts differently in the UI when they have been changed from the auto layout ? Will the user have to change the 4 layouts ? I think having to think about 4 different screen size is a bit difficult for the user, I wonder why decision was made to increase to 4, instead of reducing to 2 layouts ? As far as I understood, dashboards would be used mainly in articles in a CMS where the width would be not so large ?

app/configurator/components/preview-breakpoint.tsx
ptbrowne: Not a blocker: OK thanks, I wonder if this is intuitive enough for the user, as they are not accustomed to this. First time I see this need, I wonder if this is not indicative of a layout problem.

bprusinowski: With a layout problem, you mean that we can't show the scrollbar at the bottom of the viewport, even if the content is taller? I could also do it, but then the the vertical scroll would only happen in that container, instead of the whole layouting area, as is the case now. But maybe it would be better anyway, to keep the open sidebar / preview buttons always visible 👍

bprusinowski:
Thanks for reviewing and feedback @ptbrowne 🙇‍♂️ 

- For the showing / closing of sidebar, I would ping @KerstinFaye – I think the initial idea was to give more screen estate to be able to preview the charts without distractions, as it's not intended to edit e.g. shared filters or titles while in this mode. Maybe changing the `>>` icon to `X` would already help?
- Here also ping @KerstinFaye, I agree it could be a bit more explicit, maybe again with the "close preview" button, rather than "show sidebar and exit the preview by proxy",
- I am not sure if I reproduce, I'm able to close the title / description annotator with "Back to main button". But it's another question if we should be able to edit the titles while in preview, I guess it should be disabled 👍 
- Do you mean some additional text, e.g. at the top, indicating whether the given layout "fits" on user's screen? Like `Preview width: 1280px / Screen width: 1086px`? I tried to make it obvious by having scrollbar on top, so you immediately see that it needs to be scrolled, but that would be another solution :)

For marking the layouts as "edited", we also discussed this briefly with @KerstinFaye yesterday; we could have e.g. a "reset" button to be able to get back to original layout, derived automatically from largest one, currently we need to switch the layout from free canvas to a different one, go back to editing mode, then to layouting and change layout to free canvas again. Here we didn't make a decision yet, but could be a nice improvement 😄 

I think we had even more breakpoints at first (5?), but without an option to modify the smaller ones. I think it's good to give an option to edit each layout separately, as e.g. by default the two-column layout moves charts to the right, and not left, as I'd expect. Having a way to fix it or tailor the layout for each breakpoint would help mitigate that, but also could make it more time-consuming if you have lots of charts and really custom layout with totally different order than a default one.

Here I think also a potential solution would be something like "propagate layout down", to auto-create layouts for breakpoints smaller than the current one based on current layout 🤔 

Screenshot 2024-09-12 at 14 33 51

ptbrowne:
> Here I think also a potential solution would be something like "propagate layout down", to auto-create layouts for breakpoints smaller than the current one based on current layout 🤔

This is the default behavior of the library if you do not specify a layout. So maybe if a responsive layout has not been changed by the user, it should not be saved ?

> I am not sure if I reproduce, I'm able to close the title / description annotator with "Back to main button".

My mistake, I thought "Back to main" would go to the previous page, sorry.

>  But it's another question if we should be able to edit the titles while in preview, I guess it should be disabled 👍

Mmh, not sure, maybe it's interesting to be able to see that your title on mobile is very long and therefore should be changed ?

> Do you mean some additional text, e.g. at the top, indicating whether the given layout "fits" on user's screen? Like Preview width: 1280px / Screen width: 1086px?

I think it could be a good information for the user to know what is the size in pixel of the layout. I imagine a chart editor wanting to make a dashboard for an article, maybe they know the size of their article layout ? 

> I think it's good to give an option to edit each layout separately, as e.g. by default the two-column layout moves charts to the right, and not left, as I'd expect

I wonder about the usecase here to be able to change 4 screens ? I personally would feel bad to write in a documentation: "For best results, you are required to modify the 4 responsive screens". I think people generally think in terms of "mobile", "desktop" and anything beyond that is too complicated.

bprusinowski:
> This is the default behavior of the library if you do not specify a layout. So maybe if a responsive layout has not been changed by the user, it should not be saved ?

Yup, I think it's a good approach to only save the layout if it has been actually manually modified by the user 👍 Still, they could go out-of-sync if larger layout is edited afterwards, but this maybe will eliminate at least some cases of reorganizing layouts in different previews (maybe with some indication of already being edited, this would be clear).

> Mmh, not sure, maybe it's interesting to be able to see that your title on mobile is very long and therefore should be changed ?

Agree :)

> I think it could be a good information for the user to know what is the size in pixel of the layout. I imagine a chart editor wanting to make a dashboard for an article, maybe they know the size of their article layout ?

Right, this also makes sense 👍 

> I wonder about the usecase here to be able to change 4 screens ? I personally would feel bad to write in a documentation: "For best results, you are required to modify the 4 responsive screens". I think people generally think in terms of "mobile", "desktop" and anything beyond that is too complicated.

To me it's still a trade-off between an effort to create a chart, and potentially fitting better e.g. a tablet or smaller laptop screen :) Either we give an option to tailor the layouts better for more screens, resulting in better end-user experience, or we limit this to two, to make editor's life easier. I think I'd personally prioritize end-users, but there's also a point in keeping it simple.

I would like to also discuss this with Kerstin before making bigger changes, but thanks again for the comments, I think they're really valuable 💯

ptbrowne:
> I would like to also discuss this with Kerstin before making bigger changes, but thanks again for the comments, I think they're really valuable 💯

For sure, as I said, nothing is blocking for me :)

> Either we give an option to tailor the layouts better for more screens, resulting in better end-user experience, or we limit this to two, to make editor's life easier. I think I'd personally prioritize end-users, but there's also a point in keeping it simple.

I think the risk I see is that by making the experience of responsiveness too complicated, people do not do it, and then end users suffer ;) Also I think there is a point of facilitating chart creators life because if they spend less time on tweaking responsiveness issue, maybe they can spend more time finding a good title / description etc.. which in the end would help the end user ;)

app/locales/en/messages.po
KerstinFaye: @ptbrowne We had this kind of naming before. I changed it when I thought about the embed mode which is not always full width, even when we look at it on a large screen. But maybe it's more intuitive with the old names. I can change it back.

KerstinFaye:
@ptbrowne Thank you for your feedback and input.

**UI text for preview button:**
I appreciate your ideas. I've already done the changes in Accent. 
( However, the French translations are still pending. Since I translated the German version slightly differently, I would welcome your input on the French translation.)

**Scrollbar that appears when the current screen size is too small for the preview:**
I discussed an improvement with @bprusinowski to place the scrollbar at the bottom of the window and create a sticky segment for the "show sidebar" and preview buttons at the top.  I think your suggestion to display pixel dimensions is interesting, but I'm not sure it's necessary given our 4 available screen preview sizes. (but I keep it in mind for the next iteration)

**Making the preview mode more explicit and having the UI elements in the preview mode closer together:**
The preview mode still allows users to edit. To me hiding the sidebar rather than having an "exit preview mode" option would be the correct element for this case. Users can still edit charts and titles in this mode.

**4 preview sizes & and edit independently:**
I find the option to edit and make changes for each size independently valuable. 
And to be mentioned as well that it is not an obligation, since they can simply proceed without making any changes and leave the default version.  The default version already provides good results, which can often be published as-is. (This is different from a scenario where users would have to create everything from scratch.)
I agree with Bartosz that a reset button to revert to the default version could be useful. I'll add this to the list for the next iteration. We've also considered an option to "sync layout based on next bigger size," but this enhancement can be discussed later. 
For now, I don't think it takes users much time to scan through the preview sizes and check if it looks good. 
I wouldn't automatically adapt the smaller size to changes made in the bigger size, for the reason Bartosz mentioned. I believe it's easier for users to understand that each size can be edited individually.
I'd like to keep the 4 different sizes, since I realized in the usability test, that small desktop screen is often used as well.

PR #1745, opened 2024-09-13
style: Improved dashboard embed style & Overlapping Axis Title for Single Line Charts

noahonyejese:


app/components/chart-published.tsx
```diff
@@ -53,6 +53,7 @@ import {
   InteractiveFiltersChartProvider,
   InteractiveFiltersProvider,
 } from "@/stores/interactive-filters";
+import { theme } from "@/themes/federal";
```
bprusinowski, line 56: Generally `theme` shouldn't be imported directly, but rather used through:

```ts
import { useTheme } from "@mui/material";

const Component = () => {
  const theme = useTheme();
}
```

see https://v5.mui.com/material-ui/customization/theming/#accessing-the-theme-in-a-component 👀 

With this, in case there's e.g. a black theme in the future, you do not need to edit imports, but rather it would be handled automatically through [ThemeProvider](https://v5.mui.com/material-ui/customization/theming/#theme-provider) 👍

app/components/chart-published.tsx
```diff
@@ -121,11 +127,22 @@ export const ChartPublished = (props: ChartPublishedProps) => {
...
     <MetadataPanelStoreContext.Provider value={metadataPanelStore}>
       <InteractiveFiltersProvider chartConfigs={state.chartConfigs}>
         {state.layout.type === "dashboard" ? (
-          <>
+          <Box
+            style={{
+              padding: getPadding(),
+              backgroundColor: theme.palette.grey[200],
+            }}
+          >
```
bprusinowski, line 145: I think it would make sense to either use:

```ts
<Box sx={{ p: getPadding(), backgroundColor: "grey.200" }}>...</Box>
```

or

```ts
<div style={{ padding: getPadding(), backgroundColor: theme.palette.grey[200] }}>...</div>
```

or even extract styles to `makeStyles` to decouple the styles definitions from the component body altogether 👀 

The benefit of using `Box` over `div` is that you can use the [sx](https://mui.com/system/getting-started/the-sx-prop/) prop that doesn't require importing `theme` upfront – as you already have it imported for the `isMobile`, ... hooks, you could fall back to a `div` here 👍

app/components/chart-published.tsx
```diff
@@ -142,13 +159,22 @@ export const ChartPublished = (props: ChartPublishedProps) => {
...
-              renderChart={renderChart}
-            />
-          </>
+            <Flex
+              style={{
+                gap: 16,
+              }}
+              sx={{
+                flexDirection: "column",
+              }}
```
bprusinowski, line 168: I think we shouldn't mix up `style` and `sx` props 👀 You could change it to `<Flex flexDirection="column" gap={4}>`

also see [theme spacings](https://github.com/visualize-admin/visualization-tool/blob/03f3eef58b288fe25dc8784b55dee54763550c1c/app/themes/federal.tsx#L110) for shortcut values, based on index 👍

app/components/chart-published.tsx
```diff
@@ -142,13 +159,22 @@ export const ChartPublished = (props: ChartPublishedProps) => {
...
+            >
+              <ChartPanelLayout
+                layoutType={state.layout.layout}
+                chartConfigs={state.chartConfigs}
+                renderChart={renderChart}
+              />
+              <VisualizeLink />
```
bprusinowski, line 175: I think we should only add link when we are previewing a dashboard, see that it's already shown in footnotes in case we look at a single chart: https://github.com/visualize-admin/visualization-tool/blob/03f3eef58b288fe25dc8784b55dee54763550c1c/app/components/chart-published.tsx#L421

app/components/chart-published.tsx
```diff
@@ -107,6 +108,11 @@ export const ChartPublished = (props: ChartPublishedProps) => {
   const { dataSource } = state;
   const locale = useLocale();
   const metadataPanelStore = useMemo(() => createMetadataPanelStore(), []);
+
+  const isMobile = useMediaQuery(theme.breakpoints.down("md"));
+  const isDesktop = useMediaQuery(theme.breakpoints.up("lg"));
+  const isTablet = !isMobile && !isDesktop;
```
bprusinowski, line 114: Nit: We already use the `isMobile` logic in chart panel layout components; I think it would be cool to extract them to some utils file 🥹

bprusinowski (commented):
Thanks @noahonyejese! Generally looks good, but I think some things should be improved before merging, especially the `VisualizeLink` that should only be shown in case of dashboards and some smaller style-related adjustments 🎨

app/components/chart-published.tsx
noahonyejese: > I think we shouldn't mix up `style` and `sx` props 👀 You could change it to `<Flex flexDirection="column" gap={4}>`
> 
> 
> 
> also see [theme spacings](https://github.com/visualize-admin/visualization-tool/blob/03f3eef58b288fe25dc8784b55dee54763550c1c/app/themes/federal.tsx#L110) for shortcut values, based on index 👍 

Thanks, I'm not too familiar with mui, have been using other libs in the past.

app/components/chart-published.tsx
noahonyejese: Isn't state.layout.type === "dashboard" checking if it's dashboard?

app/components/chart-published.tsx
bprusinowski: Sure, np, it's not a huge thing, but it's helpful to know some of the benefits we get through it 👍

app/components/chart-published.tsx
bprusinowski: Ah, my bad, I didn't see that there's code below that renders non-dashboard charts 😅 Still, we could have a situation of a dashboard with one chart, e.g. when we edit a layout with two charts, and then remove one while in layouting step (see that in the screenshot below the chart already contains a link).

Screenshot 2024-09-13 at 14 20 06

So maybe in fact, there should be a safe-guard added to `showVisualizeLink={state.chartConfigs.length === 1}`, like:

```ts
showVisualizeLink={state.chartConfigs.length === 1 && state.layout.type !== "dashboard"}
```

so in such cases we only show the text link that you added in this PR, outside of the chart itself?

noahonyejese: At least in Kerstins design for dashboards it's not in the individual charts.

app/charts/shared/axis-height-linear.tsx
```diff
@@ -61,7 +64,10 @@ export const AxisHeightLinear = () => {
           {state.yAxisLabel}
         </text>
       ) : (
-        <foreignObject width={axisTitleWidth} height={axisLabelFontSize * 2}>
+        <foreignObject
+          width={textIsOverlapping ? state.bounds.chartWidth : axisTitleWidth}
+          height={axisLabelFontSize * overlappingAmount * 2}
+        >
```
bprusinowski, line 70: @noahonyejese thanks for improving this 💯

I think there's still a need to increase the chart height / margin between the axis and the graphical elements – now it's already better, as it wraps, but we can end up with a situation like this:

Screenshot 2024-09-16 at 08 58 56

bprusinowski: @noahonyejese I am not sure if this fixed the problem, I think you need to also update chart bounds to take into account the extended chart height, like you did for [combo charts](https://github.com/visualize-admin/visualization-tool/blob/f18afecffe67893e79010db1779f223deaca82ac/app/charts/combo/combo-line-dual-state.tsx#L215-L217)?

Screenshot 2024-09-22 at 13 04 37

PR #1746, opened 2024-09-13
fix: Possible filters that make sense (in context of config filters)

bprusinowski:


app/charts/shared/chart-data-filters.tsx
```diff
@@ -379,6 +370,39 @@ const DataFilter = ({
...
+          position: i + Object.keys(orderedNotInteractiveFilters).length,
+        },
+      ];
+    })
+  );
+  return {
+    ...orderedNotInteractiveFilters,
+    ...orderedInteractiveFilters,
+  };
+};
```
ptbrowne, line 404: Could be a bit simplified using pickBy and mapValues.

```
const nonInteractiveFilters = pickBy(filters, (value, key) => !(key in interactiveFilters))

let i = 0
return mapValues({
  ...nonInteractiveFilters,
  ...interactiveFilters,
}, (value) => ({ ...value, position: i++ }));
```

ptbrowne (approved):
Thanks 👍

Rdataflow:
🎉

PR #1753, opened 2024-09-23
fix: Only show tooltip if chart title is truncated

### LINKED ISSUE #1750: User Profile: Align Chart Title Tooltip
In Sprint 5.11 we introduced the enhancement where the full chart title is shown on hover on the User Profile Landing Page. The following enhancements are still needed. 

- [ ] Only show the full title on hover, for long titles, e.g. Titles that are truncated in the list. 
Image

- [ ] Align the tooltip with the Chart Title (it is currently placed to the right of the title) -> assumed to resolve itself if we only show the tooltip if the title has been truncated 

Image

- [ ] Apply the same behaviour and show a a tooltip if the dataset name is truncated 

Image

bprusinowski:
Closes #1750

## How to test
Will be possible to test once we deploy to TEST.


noahonyejese (approved):
LGTM 🎉

Issue #1809, opened 2024-10-31
✅ Add Visualize Change Log


sosiology:
**Is your feature request related to a problem? Please describe.**
Add a link to the changelog so users can quickly inform themselves of the current changes, new features, and bug fixes. 

**Describe the solution you'd like**
We add a link to the change log in github in the footer. Replace the version number in parenthesis with "Change Log" and link to https://github.com/visualize-admin/visualization-tool/blob/main/CHANGELOG.md 
![Image](https://github.com/user-attachments/assets/8f49382e-5575-4372-8ea2-8528a1d4440f)

**Describe alternatives you've considered**
We add a dedicated "What's New" page as designed as part of the landing page redesign (dependency to #1807)
![Image](https://github.com/user-attachments/assets/f2dd7944-f691-4c45-8322-2bf23fbbfa24)
Pros: A nice interface, fully integrated without having to send users to another page (github)
Cons: A higher maintenance effort, as we would most likely have to add the content prior to a release, and translate it in DE, FR, IT

**Usecases and impact**
Users can quickly inform themselves of new features or changes in visualize. 

**Additional context**
Gitlab Issue: https://gitlab.ldbar.ch/bafu/visualize/-/issues/693

Issue #1794, opened 2024-10-22
Add Option to remove border for embedded charts


sosiology:
**Is your enhancement related to a problem? Please describe.**
At the moment, the border can be removed from the embedded charts by manually adjusting the iframe code and enhancing it with `?disableBorder=true`. 

**Describe the solution you'd like**
The option to remove the border of an iframe  should be made available to all users directly in the UI, so that they can choose to directly copy the iframe code with our without border. 

**Describe alternatives you've considered**
Leave as is.

**Usecases and impact**
This is becoming increasingly relevant for users who want to embed charts into their website. The initial request came from zazuko as it was needed for a BLW project https://blv-dashboard.zazukoians.org/food_and_nutrition/meat-inspection-statistics/meat-inspection-statistics-ch

**Additional context**
Gitlab Issue: https://gitlab.ldbar.ch/bafu/visualize/-/issues/701

sosiology:
Decision: We will only have the "remove border" option for single charts
@noahonyejese

sosiology:
Thanks for the refinement @noahonyejese LGTM! 👏

sosiology:
@noahonyejese when testing on INT the border option is still available for tabs and dashboards

Tab example: https://int.visualize.admin.ch/en/v/CjRciHD7YP7S?dataSource=Prod

Dashboard example: https://int.visualize.admin.ch/en/v/F6FU0bjgBEg0?dataSource=Prod

not sure if the wrong PR got merged, please check

PR #1748, opened 2024-09-16
feat: Embed mode improvements

noahonyejese:
Closes #1747

This PR:
- adds a responsive iFrame mode (with a small JS script that listens for a custom resize event and adjusts iFrame height accordingly),
- adds a way to remove chart border via URL query param (`?disableBorder=true`). This is not yet possible to set via UI, but this will be done in a next iteration (the feature is needed now for a BLW dashboard).

## How to reproduce
1. Go to [this link](https://seleniumbase.io/w3schools/iframes).
2. Replace the HTML content with below snippet (copied from [this link](https://test.visualize.admin.ch/en/v/lQjE19b_FBql?dataSource=Prod) [and modified a bit to show the iframe border, so it's visible that it's not responsive]).
```html
<!DOCTYPE html>
<html>
<body>

<iframe src="https://test.visualize.admin.ch/en/embed/lQjE19b_FBql" style="border:1px solid #000; max-width: 100%" name="visualize.admin.ch" scrolling="no" frameborder="1" marginheight="0px" marginwidth="0px" height="640px" width="600px" allowfullscreen></iframe>

</body>
</html>
```
3. ❌ Resize the panel and see that the iframe does not resize dynamically.
4. ❌ See that the chart has a gray border.

## How to test
1. Go to [this link](https://seleniumbase.io/w3schools/iframes).
2. Replace the HTML content with below snippet (copied from [this link](https://visualization-tool-dhein6se5-ixt1.vercel.app/en/v/-Gs1rzUcZFz7?dataSource=Prod) [responsive embed mode, modified a bit to show the iframe border, so it's visible that it's responsive]).
```html
<!DOCTYPE html>
<html>
<body>

<iframe src="https://visualization-tool-dhein6se5-ixt1.vercel.app/en/embed/-Gs1rzUcZFz7?disableBorder=true" width="100%" height="0px" style="border:1px solid #000;" name="visualize.admin.ch"></iframe>
<script type="text/javascript">!function(){window.addEventListener("message", function (e) { if (e.data.type === "VISUALIZE-CHART-RESIZE") { document.querySelectorAll("iframe").forEach((iframe) => { if (iframe.contentWindow === e.source) { iframe.style.height = e.data.height + "px"; } }); } })}();</script>

</body>
</html>
```
3. ✅ Resize the panel and see that the iframe resizes dynamically.
4. Modify the style property to change the border to the original one: `border:1px #ffffff none;`.
5. ✅ See that the chart misses a border.

PR #1812, opened 2024-11-05
Complete smaller enhancements

noahonyejese:
Let know if this is works technically @bprusinowski, unfortunately I had difficulties spotting the border in the first place. Once the technically side is clear I will tag Kerstin to Review the design. - Currently don't know how to test exactly because I haven't seen a border in the first place...

### This PR: 
1. This PR allows the users to remove the border during Embed Selection 
2. This PR tackles the issue #1809 

---

### How to Reproduce (#1809)

1. Go to this [link](https://visualize.admin.ch) 
2. See how in the footer there is the id of the last commit ❌
4. Click on the Link get navigated to last commit ❌ 

### How to Test (#1809)
1. Go to this [link](https://visualization-tool-c62th0teo-ixt1.vercel.app)
2. See how in the footer it now says "Change Log" translated in all languages ✅
3. Click on the link get navigated to the Change log ✅

### How to Test (#1794)
1. Go to this [link](https://visualization-tool-c62th0teo-ixt1.vercel.app)
2. Create a Chart and Publish it 
3. Try to Share using the Embed button see that now you can Embed without the border ✅

@KerstinFaye this PR is almost complete please give me insights on the design side of things, related to the switch use the link above, also @sosiology and @adintegra let us know what you think about the proposal from @bprusinowski adding a separate Link for the `Changelog` and keeping the most recent commit there. - Thanks guys

### DIFF
--- a/app/components/footer.tsx
+++ b/app/components/footer.tsx
@@ -6,22 +6,28 @@ import { forwardRef, ReactNode } from "react";
 
 import Flex, { FlexProps } from "@/components/flex";
 import contentRoutes from "@/content-routes.json";
-import { BUILD_VERSION } from "@/domain/env";
+import { BUILD_COMMIT, BUILD_GITHUB_REPO, BUILD_VERSION } from "@/domain/env";
 import { useLocale } from "@/locales/use-locale";
 
 const Version = () => {
-  const commitLink = (
-    <>
-      (
-      <Link
-        color="primary"
-        href={`https://github.com/visualize-admin/visualization-tool/blob/main/CHANGELOG.md`}
-      >
-        Changelog
-      </Link>
-      )
-    </>
-  );
+  let commitLink = null;
+
+  if (BUILD_GITHUB_REPO && BUILD_COMMIT) {
+    commitLink = (
+      <>
+        (
+        <Link
+          color="primary"
+          href={`${BUILD_GITHUB_REPO}/commit/${BUILD_COMMIT}`}
+        >
+          {BUILD_COMMIT.substr(0, 7)}
+        </Link>
+        )
+      </>
+    );
+  } else if (BUILD_COMMIT) {
+    commitLink = `(${BUILD_COMMIT.substr(0, 7)})`;
+  }
 
   return (
     <>
@@ -121,9 +127,13 @@ export const Footer = ({ sx }: { sx?: FlexProps["sx"] }) => {
           >
             <Version />
           </Typography>
+          <FooterLinkBottom href="https://github.com/visualize-admin/visualization-tool/blob/main/CHANGELOG.md">
+            Changelog
+          </FooterLinkBottom>
           <FooterLinkBottom href="https://www.youtube.com/channel/UCNK0uJTJ74kbv3jmNtZfgOw">
             <Trans id="footer.tutorials">Tutorials</Trans>
           </FooterLinkBottom>
+
           <FooterLinkBottom href="https://visualization-tool.status.interactivethings.io/">
             <Trans id="footer.status">Status</Trans>
           </FooterLinkBottom>
--- a/app/components/footer.tsx
+++ b/app/components/footer.tsx
@@ -17,7 +17,7 @@ const Version = () => {
         color="primary"
         href={`https://github.com/visualize-admin/visualization-tool/blob/main/CHANGELOG.md`}
       >
-        <Trans id="footer.version.change-log">Change Log</Trans>
+        Changelog
       </Link>
       )
     </>
--- a/app/components/footer.tsx
+++ b/app/components/footer.tsx
@@ -6,28 +6,22 @@ import { forwardRef, ReactNode } from "react";
 
 import Flex, { FlexProps } from "@/components/flex";
 import contentRoutes from "@/content-routes.json";
-import { BUILD_COMMIT, BUILD_GITHUB_REPO, BUILD_VERSION } from "@/domain/env";
+import { BUILD_VERSION } from "@/domain/env";
 import { useLocale } from "@/locales/use-locale";
 
 const Version = () => {
-  let commitLink = null;
-
-  if (BUILD_GITHUB_REPO && BUILD_COMMIT) {
-    commitLink = (
-      <>
-        (
-        <Link
-          color="primary"
-          href={`${BUILD_GITHUB_REPO}/commit/${BUILD_COMMIT}`}
-        >
-          {BUILD_COMMIT.substr(0, 7)}
-        </Link>
-        )
-      </>
-    );
-  } else if (BUILD_COMMIT) {
-    commitLink = `(${BUILD_COMMIT.substr(0, 7)})`;
-  }
+  const commitLink = (
+    <>
+      (
+      <Link
+        color="primary"
+        href={`https://github.com/visualize-admin/visualization-tool/blob/main/CHANGELOG.md`}
+      >
+        <Trans id="footer.version.change-log">Change Log</Trans>
+      </Link>
+      )
+    </>
+  );
 
   return (
     <>
--- a/app/components/publish-actions.tsx
+++ b/app/components/publish-actions.tsx
@@ -9,7 +9,6 @@ import {
   Switch,
   Tooltip,
   Typography,
-  useEventCallback,
 } from "@mui/material";
 import { ChangeEvent, ReactNode, RefObject, useEffect, useState } from "react";
 
@@ -67,7 +66,7 @@ const Embed = ({ chartWrapperRef, configKey, locale }: PublishActionProps) => {
   const [embedUrl, setEmbedUrl] = useState("");
   const [embedAEMUrl, setEmbedAEMUrl] = useState("");
   const [isResponsive, setIsResponsive] = useState(true);
-  const [withoutBorder, setWithoutBorder] = useState(false);
+  const [isWithoutBorder, setIsWithoutBorder] = useState(false);
   const [iframeHeight, setIframeHeight] = useState(0);
 
   const handlePopoverOpen = useEvent(() => {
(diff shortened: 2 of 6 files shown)


app/components/publish-actions.tsx
```diff
@@ -65,6 +67,7 @@ const Embed = ({ chartWrapperRef, configKey, locale }: PublishActionProps) => {
   const [embedUrl, setEmbedUrl] = useState("");
   const [embedAEMUrl, setEmbedAEMUrl] = useState("");
   const [isResponsive, setIsResponsive] = useState(true);
+  const [withoutBorder, setWithoutBorder] = useState(false);
```
bprusinowski, line 70: Nit: `isWithoutBorder`? To be consistent with the top variables :)

app/components/publish-actions.tsx
```diff
@@ -78,14 +81,22 @@ const Embed = ({ chartWrapperRef, configKey, locale }: PublishActionProps) => {
...
-    setEmbedUrl(`${origin}/${locale}/embed/${configKey}`);
-    setEmbedAEMUrl(`${origin}/api/embed-aem-ext/${locale}/${configKey}`);
-  }, [configKey, locale]);
+    setEmbedUrl(
+      `${origin}/${locale}/embed/${configKey}${withoutBorder ? "?disableBorder=true" : ""}`
+    );
+    setEmbedAEMUrl(
+      `${origin}/api/embed-aem-ext/${locale}/${configKey}${withoutBorder ? "?disableBorder=true" : ""}`
+    );
+  }, [configKey, locale, withoutBorder]);
```
bprusinowski, line 90: 👍

app/components/publish-actions.tsx
```diff
@@ -78,14 +81,22 @@ const Embed = ({ chartWrapperRef, configKey, locale }: PublishActionProps) => {
...
 
   const handleChange = useEvent((e: ChangeEvent<HTMLInputElement>) => {
     setIsResponsive(e.target.value === "responsive");
   });
 
+  const handleStylingChange = useEventCallback((checked: boolean) => {
+    setWithoutBorder(checked);
+  });
```
bprusinowski, line 98: Nit: As now we have two callbacks, I'd change the name of `handleChange` to `handleResponsiveChange`, now it looks quite generic. Also, there's `useEvent` and `useEventCallback` hooks used after each one, while they have the same purpose – I'd settle on using only one of them 👀

app/components/publish-actions.tsx
```diff
@@ -149,9 +160,23 @@ const Embed = ({ chartWrapperRef, configKey, locale }: PublishActionProps) => {
                   "For embedding visualizations in systems without JavaScript support (e.g. WordPress).",
               })}
             />
+            <EmbedToggleSwitch
+              value="without-border"
+              checked={withoutBorder}
+              onChange={(_, checked) => handleStylingChange(checked)}
```
bprusinowski, line 166: At the top, `handleStylingChange` is wrapped in `useEventCallback`, but here we have a function that's re-created on each render (`(_, checked) => ...`). In order to improve the caching, I'd change the `handleStylingChange` callback to accept an event as the first function argument, and only pass `onChange={handleStylingChange}` here (see how it's used in the line 152 in `handleChange`, where we don't create another arrow function) 👍

app/components/publish-actions.tsx
```diff
@@ -149,9 +160,23 @@ const Embed = ({ chartWrapperRef, configKey, locale }: PublishActionProps) => {
...
+              label={t({
+                id: "publication.embed.iframe.remove-border",
+                message: "Iframe without border",
```
bprusinowski, line 169: Nit: Maybe simply "Remove border" would suffice? I'd also move this to another line, as it's not about the same "category" as being responsive or not, but maybe @KerstinFaye has an opinion there 😅

app/components/publish-actions.tsx
```diff
@@ -149,9 +160,23 @@ const Embed = ({ chartWrapperRef, configKey, locale }: PublishActionProps) => {
...
+              })}
+              infoMessage={t({
+                id: "publication.embed.iframe.remove-border.warn",
+                message:
+                  "For embedding visualizations in systems without a Border.",
```
bprusinowski, line 174: Nit: `without a border` (lowercase) :)

app/components/publish-actions.tsx
```diff
@@ -149,9 +160,23 @@ const Embed = ({ chartWrapperRef, configKey, locale }: PublishActionProps) => {
...
+              })}
+            />
           </Flex>
           <CopyToClipboardTextInput
-            content={`<iframe src="${embedUrl}" width="100%" style="${isResponsive ? "" : `height: ${iframeHeight || 640}px; `}border:0px #ffffff none;" name="visualize.admin.ch"></iframe>${isResponsive ? `<script type="text/javascript">!function(){window.addEventListener("message", function (e) { if (e.data.type === "${CHART_RESIZE_EVENT_TYPE}") { document.querySelectorAll("iframe").forEach((iframe) => { if (iframe.contentWindow === e.source) { iframe.style.height = e.data.height + "px"; } }); } })}();</script>` : ""}`}
+            content={`<iframe src="${embedUrl}" width="100%" style="${isResponsive ? "" : `height: ${iframeHeight || 640}px; `}border: ${withoutBorder ? "0px #ffffff none;" : "1px #ffffff;"}" name="visualize.admin.ch"></iframe>${isResponsive ? `<script type="text/javascript">!function(){window.addEventListener("message", function (e) { if (e.data.type === "${CHART_RESIZE_EVENT_TYPE}") { document.querySelectorAll("iframe").forEach((iframe) => { if (iframe.contentWindow === e.source) { iframe.style.height = e.data.height + "px"; } }); } })}();</script>` : ""}`}
```
bprusinowski, line 179: Here we don't need to modify this, only `disableBorder=true` URL param is needed (iframe itself needs to not have a border at all times).

app/locales/en/messages.po
```diff
@@ -1511,6 +1511,14 @@ msgstr "Embed Code"
 msgid "publication.embed.iframe.caption"
 msgstr "Use this link to embed the chart in any webpage."
 
+#: app/components/publish-actions.tsx
+msgid "publication.embed.iframe.remove-border"
+msgstr "Iframe without border"
+
+#: app/components/publish-actions.tsx
+msgid "publication.embed.iframe.remove-border.warn"
+msgstr "For embedding visualizations in systems without a Border."
```
bprusinowski, line 1524: Nit: ...`without a border.`

app/components/footer.tsx
```diff
@@ -6,28 +6,22 @@ import { forwardRef, ReactNode } from "react";
...
-    commitLink = `(${BUILD_COMMIT.substr(0, 7)})`;
-  }
+  const commitLink = (
+    <>
+      (
+      <Link
+        color="primary"
+        href={`https://github.com/visualize-admin/visualization-tool/blob/main/CHANGELOG.md`}
+      >
+        <Trans id="footer.version.change-log">Change Log</Trans>
```
bprusinowski, line 20: I'd use a `Changelog` convention, without a space (see how it's defined in the actual changelog: https://github.com/visualize-admin/visualization-tool/blob/main/CHANGELOG.md) 👀

bprusinowski (commented):
Thanks @noahonyejese, generally looks good, but it would be good to take a look at few comments I left 👍 

Also, I found it useful to have a reference to a most recent commit at least a few times – as we don't have access to INT or PROD deployment pipelines, it's sometimes not obvious if the new version has already been deployed, or not, and this is an easy way to make sure. I remember discovering an issue with Abraxas' automatic deployment system in the past through seeing that the commit number was not updated for longer that expected.

I'd personally vote to keep it, as otherwise we'd need to manually check what is a fix or feature we recently deployed to INT, and going through the application to see if this version is already up, or not, by trying to reproduce the issue. Maybe we can simply have the `Changelog` link as another entry in the footer, but keep the current commit hash? cc @sosiology @KerstinFaye @adintegra

bprusinowski:
@noahonyejese for testing the iframe's border, you can take a look at instructions from https://github.com/visualize-admin/visualization-tool/pull/1748 👍

app/components/publish-actions.tsx
noahonyejese: Ah okay alright, tried to copy and example from an other instance using Switch

app/components/footer.tsx
noahonyejese: @bprusinowski what about other languages?

app/components/publish-actions.tsx
noahonyejese: Yeah true, I first had is as radio button so then the text made more sense like so

app/components/footer.tsx
bprusinowski: Not 100% sure about that 😅 I think that `Changelog` is a bit of an international word, but maybe there are dedicated words in other languages that don't sounds weird (e.g. in Polish, I wouldn't use "rejestr zmian" for that, but keep "Changelog") 😅

noahonyejese: In German it would be `Änderungsprotokoll`, I personally think `Changelog` is better, is this our call?

bprusinowski: I'd say we go with `Changelog`, we can adjust this easily going forward 👍

adintegra:
Agree with all points here 🎉 Thank you both, @noahonyejese and @bprusinowski!

app/locales/it/messages.po
```diff
@@ -1274,7 +1274,7 @@ msgstr "Tutorials"
 
 #: app/components/footer.tsx
 msgid "footer.version.change-log"
-msgstr "Registro modifiche"
+msgstr ""
```
bprusinowski, line 1278: Nit: I think all translations strings for the `Changelog` can be removed :)

noahonyejese: missed that one thanks

noahonyejese:
> Thanks @noahonyejese, generally looks good, but it would be good to take a look at few comments I left 👍
> 
> Also, I found it useful to have a reference to a most recent commit at least a few times – as we don't have access to INT or PROD deployment pipelines, it's sometimes not obvious if the new version has already been deployed, or not, and this is an easy way to make sure. I remember discovering an issue with Abraxas' automatic deployment system in the past through seeing that the commit number was not updated for longer that expected.
> 
> I'd personally vote to keep it, as otherwise we'd need to manually check what is a fix or feature we recently deployed to INT, and go through the application to see if this version is already up, or not, by trying to reproduce the issue. Maybe we can simply have the `Changelog` link as another entry in the footer, but keep the current commit hash? cc @sosiology @KerstinFaye @adintegra

This has been changed now to use a separate footer link I think this makes more sense as well

bprusinowski (approved):
LGTM, thanks!

KerstinFaye:
Hi @noahonyejese. Thanks for the fast work. Regarding the #1794. It looks good to me and I agree that the toggle is the correct element here. This would also visually distinguish it from the other options, which is good. Could we add in the tooltip border 'line' or boundary line instead of only border? Not that users confuse the border with the space around the chart instead of only the line.

KerstinFaye:
@noahonyejese @bprusinowski Regarding #1809. I agree of having an alternative link with ChangeLog. That's a good thought that you had @bprusinowski. I think these changes are good in anyways, even if we are not immediately working on the landing page section that communicates the changes. (Even though I'd love to push this.)

noahonyejese:
@KerstinFaye nope still had the wrong link somehow here you go, try here https://visualization-tool-32ubsc2ls-ixt1.vercel.app

sosiology:
Nice - LGTM @noahonyejese 👏

sosiology:
@noahonyejese 

I had a look on test and it works well for single charts, but i noticed the following: 

Tab Layout: see https://test.visualize.admin.ch/en/embed/eznZX6RN-XlV?disableBorder=true To me the tabs look a bit lost - maybe here we need to force a border? 

Dashboards: We don't support this, correct? can we not enable the option only for dashboards?

@KerstinFaye would be interested in your feedback. Lets to discuss on Monday :)

KerstinFaye:
@noahonyejese @sosiology as discussed today in the daily, we only allow this option for singe charts.

Issue #1813, opened 2024-11-05
✅ "Created with visualize.admin.ch" missing from dashboards


sosiology:
**Description**
The note that a chart was created with visualize.admin.ch is missing when multiple charts are published using the layout options: Tab layout, dashboard)

**Steps to Reproduce**
1. Go to: https://visualize.admin.ch/v/TVNwmsKWrtx5?dataSource=Prod
2. See that at the bottom of the chart, the hint "Erstellt mit: visualize.admin.ch" is missing 
3. Copy and Edit the Chart 
4. Select a dashboard layout (any)
5. See that in editing mode, but also in the published chart the note "Erstellt mit visualize.admin.ch" is missing. 

**Screenshots and Examples**
Tab Layout | https://visualize.admin.ch/v/TVNwmsKWrtx5?dataSource=Prod
![Image](https://github.com/user-attachments/assets/beb49b32-99c4-48d8-b822-ff065b83e4af)

Dashboard Layout | https://visualize.admin.ch/v/4yHJ0msuFuY2?dataSource=Prod 
![Image](https://github.com/user-attachments/assets/0d1ad2cd-c939-4d57-901e-d97d5a740763)

**Expected Result**
The note is shown below each chart in the tab layout, or at least once at the bottom of a dashboard. 
See example here (single chart): https://visualize.admin.ch/v/7PyLLMXyF4PW?dataSource=Prod

![Image](https://github.com/user-attachments/assets/ff5b59d5-7f89-4194-8596-a85eeed3e87d)

noahonyejese:
@bprusinowski and @sosiology this is an issue I've worked on the past the idea was to move the "Created with" part onto the bottom when creating Dashboards with multiple Charts. (I now fixed the missing "Created with" on the single Chart). However how should it be on the Dashboard Charts with multiple Charts during Edit mode should it be for every Chart or just once like it is in published mode?

bprusinowski:
Hi @noahonyejese, I am not sure if this is a behavior we all have in mind, but I think the labels should be there in single charts and tab layout, and in case of dashboards, we should only have one entry at the bottom and not include the labels in individual charts. @KerstinFaye maybe we have a design for this? 🤔

sosiology:
@noahonyejese the goal is to have a solution for dashboards - where, as @bprusinowski said, it is shown in one place (not per chart). The single chart types should already have the "Created with" ....so i am not 100% sure what you fixed?

[...]

PR #1816, opened 2024-11-06
fix: Added Links for all Charts in Preview & Published Mode

noahonyejese:
This PR fixes Issue #1813 

### How to Test
1. Go to this [Link](https://visualization-tool-azw7yjcr8-ixt1.vercel.app) 
2. Publish, a tab layout see how "Created with..." is visible ✅
3. Publish a Dashboard Layout with 1 Chart see how "Created with..." is in the Chart ✅
4. Publish a Dashboard Layout with more than 1 Chart see how the "Created with..." is only at the bottom of the layout ✅

### DIFF
--- a/app/components/chart-preview.tsx
+++ b/app/components/chart-preview.tsx
@@ -568,6 +568,7 @@ const ChartPreviewInner = ({
                   chartConfig={chartConfig}
                   dashboardFilters={state.dashboardFilters}
                   components={allComponents}
+                  showVisualizeLink
                 />
                 {/* Wrap in div for subgrid layout */}
                 <div className="debug-panel">
--- a/app/components/chart-preview.tsx
+++ b/app/components/chart-preview.tsx
@@ -568,7 +568,6 @@ const ChartPreviewInner = ({
                   chartConfig={chartConfig}
                   dashboardFilters={state.dashboardFilters}
                   components={allComponents}
-                  showVisualizeLink
                 />
                 {/* Wrap in div for subgrid layout */}
                 <div className="debug-panel">
--- a/app/components/chart-published.tsx
+++ b/app/components/chart-published.tsx
@@ -11,7 +11,7 @@ import { LoadingStateProvider } from "@/charts/shared/chart-loading-state";
 import { isUsingImputation } from "@/charts/shared/imputation";
 import { CHART_RESIZE_EVENT_TYPE } from "@/charts/shared/use-size";
 import { ChartErrorBoundary } from "@/components/chart-error-boundary";
-import { ChartFootnotes } from "@/components/chart-footnotes";
+import { ChartFootnotes, VisualizeLink } from "@/components/chart-footnotes";
 import { ChartPanelLayout, ChartWrapper } from "@/components/chart-panel";
 import {
   ChartControls,
@@ -155,6 +155,7 @@ export const ChartPublished = ({
                 chartConfigs={state.chartConfigs}
                 renderChart={renderChart}
               />
+              <VisualizeLink />
             </Box>
           ) : (
             <>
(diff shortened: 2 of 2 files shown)


app/components/chart-preview.tsx
```diff
@@ -568,6 +568,7 @@ const ChartPreviewInner = ({
                   chartConfig={chartConfig}
                   dashboardFilters={state.dashboardFilters}
                   components={allComponents}
+                  showVisualizeLink
```
bprusinowski, line 571: @noahonyejese I think we should only show this when a chart is published, the current behavior here is correct; see how it looks in the design, e.g. here: https://www.figma.com/design/Hn9wvuEYLUmdJtg4KwZYUx/visualize.admin.ch---design?node-id=2805-554265&m=dev

app/components/chart-published.tsx
```diff
@@ -443,8 +447,9 @@ const ChartPublishedInnerImpl = (props: ChartPublishInnerProps) => {
               dashboardFilters={state.dashboardFilters}
               components={allComponents}
               showVisualizeLink={
-                state.chartConfigs.length === 1 &&
-                state.layout.type !== "dashboard"
+                state.layout.type !== "dashboard" ||
+                (state.layout.type === "dashboard" &&
+                  state.chartConfigs.length === 1)
```
bprusinowski, line 452: 👍

bprusinowski (commented):
Thanks @noahonyejese, I think we shouldn't show the `Created with Visualize` link during editing as is the case now on PROD 👀 

Also, the` How to test` link doesn't work (I think you copied a `/create` link that points to a configurator state stored in your `localStorage`) – here one trick is to copy a link as shown in the screenshot that opens the editor mode for a given dataset 👍 

Screenshot 2024-11-06 at 12 48 24

noahonyejese:
@bprusinowski So not show at all for any chart during preview / edit mode ?

bprusinowski:
Yes, the issue was only that the link was missing in case of published dashboards 👍 So in editing mode we never show the link, and in published mode we show it inside individual chart if it's not a dashboard or a dashboard with one chart, or outside of dashboard group for dashboards with more than one chart 😄

bprusinowski:
I see that @sosiology mentioned in the issue that it's not shown in the editing mode, but I am pretty sure we have never shown it here, and that could lead to some issues – if you click it, you'd go out of the editing mode and go back to the home page? 🤔 And we also don't have it in the design, so I'd scope this only to published charts (or we can discuss together 👍)

bprusinowski (approved):
LGTM, thanks! 💯 

Btw, I noticed that you used a deployment link based on a commit in `How to test` instructions (https://visualization-tool-azw7yjcr8-ixt1.vercel.app/) – to make sure it stays up to date, you should use a deployment link based on a PR (https://visualization-tool-git-fix-created-reference-dashboards-ixt1.vercel.app/en) that updates automatically (I tried to test again after the most recent commit, but the deployment was not updated) :) This might be important for other people that test after some time 👍 

Screenshot 2024-11-06 at 13 11 14

noahonyejese:
> LGTM, thanks! 💯
> 
> Btw, I noticed that you used a deployment link based on a commit in `How to test` instructions (https://visualization-tool-azw7yjcr8-ixt1.vercel.app/) – to make sure it stays up to date, you should use a deployment link based on a PR (https://visualization-tool-git-fix-created-reference-dashboards-ixt1.vercel.app/en) that updates automatically (I tried to test again after the most recent commit, but the deployment was not updated) :) This might be important for other people that test after some time 👍
> 
> Screenshot 2024-11-06 at 13 11 14

oh damn, thanks a lot!

PR #1849, opened 2024-11-12
feat: Introduce component ids

### LINKED ISSUE #1696: Technical clean up of merging of cubes feature
- Index several properties in the chart config not just by `dimensionIri`, but also by `cubeIri`, as when merging two cubes, there could be overlaps in dimension iris,
- Properly handle cases in the code where we currently take the first cube, instead of multiple
- ...

bprusinowski:
Closes #1696

This PR introduces a new, **component id** concept to Visualize, which replaces the current _component iri_ convention. 

**Component id** is made by combining `unversioned cube iri` with `unversioned component iri`, which is a unique combination even in case we merged two cubes with the same `component iris`. This means that we can handle these cases correctly, while leaving our business logic mostly intact (and in some cases, even simpler, as we don't need to filter by `cube iri` and `component iri` anymore, but instead simply use **component id**).

This combination of `iris` is done on the server side and sent to the client, which doesn't need to know about this fact, but simply keep the current logic (chart config migrations are the only exception, as we need to introduce `unversioned cube iris` to relevant parts of the config dynamically to make sure old charts work).

By proxy, it was necessary to make the migrations async, so we can fetch `unversioned cube iri` when needed – as [immer doesn't support / recommend asynchronous operations](https://immerjs.github.io/immer/async/), I removed it.

ℹ️ [Dimension id docs](https://www.notion.so/interactivethings/Dimensions-87f4b91b502045ba8f1486439fdb1691?pvs=4#14430da54302800087f1ef1364b23d55)

🔎 [Visual regression testing docs](https://www.notion.so/interactivethings/Visual-regression-testing-14430da5430280cb83e0d0ccbe657a88)

### How to test
🚨 **Preferably, the implementation should be tested holistically, as essentially every part of Visualize was affected in one or another way.**

1. Go to [this link](https://visualization-tool-git-feat-finalize-merging-of-cubes-2-ixt1.vercel.app/en/create/new?cube=https://agriculture.ld.admin.ch/foag/cube/MilkDairyProducts/Consumption_Price_Month&dataSource=Prod).
2. Click on `Add dataset`.
3. Add `Market figures milk and dairy products - Monthly producer prices` dataset.
4. Open `Vertical axis`.
5. ✅ See that it's possible to toggle between two `Price` dimensions that have the same `iri`s and names, but come from different cubes.
6. Change chart type to `Table`.
7. ✅ See that there are duplicated dimensions visible in the `Table` (that have the same `iri`s and names, but come from different cubes).

## How to reproduce
Repeat the above steps on PROD environment and see that ✅ points are broken ❌.

---
- [x] Add docs to feature pages in Notion (both for `componentId` and visual regression testing)
- [x] Add tests
- [x] Check E2E tests locally
- [x] Check if all old charts work (checked local `/_charts` page and cross-compared 500 PROD charts on `main` vs this branch)


app/pages/preview_post.tsx
```diff
@@ -58,20 +59,31 @@ export const getServerSideProps: GetServerSideProps<PageProps> = async ({
...
+  useEffect(() => {
+    const run = async () => {
+      setMigrated({
+        ...(await migrateConfiguratorState(configuratorState)),
+        state: "PUBLISHED",
+      } as ConfiguratorState);
+    };
+
+    run();
+  }, [configuratorState]);
```
ptbrowne, line 72: Nit: Maybe we should have a useMigratedConfiguratorState hook here. It seems that this pattern will appear multiples (already is there from the homepage), and it seems we could be in need of having an object like { data, isLoading } similar to react useQuery. Having a central place to put this pattern could be good.

app/graphql/resolvers/rdf.ts
```diff
@@ -211,6 +219,8 @@ const getResolvedDimension = async (
     cache,
   });
 
+  console.log("iri", iri);
```
ptbrowne: 🦅

app/charts/index.spec.ts
```diff
@@ -69,12 +69,7 @@ describe("initial config", () => {
   it("should create an initial table config with column order based on dimension order", () => {
     const config = getInitialConfig({
       chartType: "table",
-      iris: [
-        {
-          iri: "https://environment.ld.admin.ch/foen/nfi",
-          publishIri: "https://environment.ld.admin.ch/foen/nfi",
-        },
-      ],
+      iris: [{ iri: "https://environment.ld.admin.ch/foen/nfi" }],
```
ptbrowne, line 72: If I understand well, the componentIds will already include the "publishIri" (I am still fuzzy on why we need publish iris 🙈 ) ?

app/rdf/query-dimension-values.ts
```diff
@@ -545,7 +545,7 @@ export const loadMinMaxDimensionValues = async ({
 }) => {
   const query = SELECT`(MIN(?value) as ?minValue) (MAX(?value) as ?maxValue)`
     .WHERE`${iriToNode(datasetIri)} ${ns.cube.observationSet}/${ns.cube.observation} ?observation .
-?observation <${dimensionIri}> ?value .
+?observation ${dimensionIri} ?value .
```
ptbrowne, line 548: How come it already contains the brackets ?

app/graphql/resolvers/rdf.ts
```diff
@@ -242,6 +240,8 @@ export const possibleFilters: NonNullable<
     return [];
   }
 
+  const filters = getFiltersWithSplitIris(_filters);
```
ptbrowne: Wondering on the naming here.

Wondering if we should have 

dimensionIri = dimension URL without the cube
dimensionId = cubeIRi:dimensionIRI

then the method becomes getFiltersComponentIris.

I think the fact that it's two strings joined is an implementation detail.

app/utils/chart-config/versioning.ts
```diff
@@ -1,15 +1,22 @@
 import { DEFAULT_OTHER_COLOR_FIELD_OPACITY } from "@/charts/map/constants";
 import { ChartConfig, ConfiguratorState } from "@/config-types";
 import { mapValueIrisToColor } from "@/configurator/components/ui-helpers";
+import { FIELD_VALUE_NONE } from "@/configurator/constants";
 import {
   LEGACY_PROD_DATA_SOURCE_URL,
   PROD_DATA_SOURCE_URL,
 } from "@/domain/datasource/constants";
+import { client } from "@/graphql/client";
+import { joinIris } from "@/graphql/resolvers/rdf";
```
ptbrowne: I think the joinIris, splitIris methods should not come directly from the resolvers, this would be a concept that would be source agnostic, no ?

app/charts/shared/chart-state.ts
```diff
@@ -298,7 +304,9 @@ export const useNumericalYVariables = (
 ): NumericalYVariables => {
   const yMeasure = measuresByIri[y.componentIri];
   if (!yMeasure) {
-    throw Error(`No dimension <${y.componentIri}> in cube!`);
+    throw Error(
+      `No dimension <${y.componentIri}> in cube! (useNumericalYVariables)`
+    );
```
ptbrowne, line 307: Shouldn't there be a `new` ? I am not sure what difference it makes actually.

app/configurator/configurator-state/reducer.spec.tsx
```diff
@@ -772,6 +772,18 @@ describe("filtering", () => {
...
+    (d) => ({
+      ...d,
+      // Adding cube prefix.
+      iri: `foo___${d.iri}`,
+    })
+  );
+  const measures = (dataSetMetadata.measures as any as Measure[]).map((m) => ({
+    ...m,
+    // Adding cube prefix.
+    iri: `foo___${m.iri}`,
```
ptbrowne: Using `joinIri` could be better here ?

app/utils/chart-config/versioning.ts
```diff
@@ -1014,23 +1012,29 @@ export const chartConfigMigrations: Migration[] = [
...
             joinIris({
               unversionedCubeIri,
               dimensionIri: iri,
             })
           );
+        } else if ("componentIri" in v) {
+          v.componentIri = joinIris({
+            unversionedCubeIri,
+            dimensionIri: v.componentIri,
+          });
```
ptbrowne: I see no tests updated for this commit (11284bf9d0f2ee291653a0a635c5c27a0f9875c6), do we need one ?

app/rdf/parse-search-results.ts
```diff
@@ -105,8 +105,10 @@ function buildSearchCubes(
           const dim = bySubjectAndPredicate.get(x.object.value);
           return {
             iri: joinIris({
-              // We don't need to unversion the cube iri here, as search cubes are
-              // temporary and dimensions coming from here aren't stored in chart config.
+              // TODO
```
ptbrowne, line 108: Comment to see if the TODO has been done :)

app/graphql/resolvers/rdf.ts
```diff
@@ -40,27 +40,33 @@ import { getPossibleFilters } from "@/rdf/query-possible-filters";
...
-  const [unversionedCubeIri, dimensionIri] = iri.split(IRI_SEPARATOR);
+export const splitComponentId = (iri: string) => {
+  const [unversionedCubeIri, unversionedComponentIri] = iri.split(ID_SEPARATOR);
 
-  return { unversionedCubeIri, dimensionIri: dimensionIri || undefined };
+  return {
+    unversionedCubeIri,
+    unversionedComponentIri: unversionedComponentIri || undefined,
+  };
 };
```
ptbrowne, line 60: This two operations could be seen as "parse" and "stringify".

ptbrowne (approved):
Thanks a lot Bartosz for this massive effort 💪 

I see no blockers here, I have mostly put nits and comments.

- The use of branded types would make sure we do not pass a dimension iri where we expect a dimension id.
- I think it would be a good idea to perform screenshot testing against production charts to make sure they are ok (load the production charts into local database, screenshot the chart on `main`, switch to this branch rescreenshot, check if they are the same)

app/rdf/queries.ts
```diff
@@ -342,13 +348,19 @@ export const getCubeObservations = async ({
...
   const cubeView = View.fromCube(cube, false);
-  const allResolvedDimensions = await getCubeDimensions({
-    cube,
-    locale,
-    sparqlClient,
-    cache,
-  });
+  const [unversionedCubeIri = cubeIri, allResolvedDimensions] =
+    await Promise.all([
+      queryCubeVersionHistory(sparqlClient, cubeIri),
```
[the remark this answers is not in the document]
bprusinowski: Yes, I think it would be a better name (I settled for that as it has a `versionHistory` predicate that returns one object from SPARQL query) 👍

app/charts/shared/use-sync-interactive-filters.spec.tsx
```diff
@@ -74,24 +74,24 @@ const interactiveFiltersConfig: InteractiveFiltersConfig = {
...
+      ...fixture.data.chartConfig,
+      interactiveFiltersConfig,
+    },
+    {
+      migrationProps: {
+        meta: {},
+        dataSet: "foo",
+      },
+    }
+  )) as ChartConfig;
```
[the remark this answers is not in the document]
bprusinowski: I think the problem with this is that it's only a `ChartConfig` if we migrate to latest version (which we usually do). Maybe it would be possible to narrow it down somehow 🤔

app/charts/index.spec.ts
bprusinowski: Yeah, I think it was added just in case – what happens now is that when we load a config, we upgrade the cube iri to latest one and overwrite it on the client-side. So `publishIri` is always stored in a db, and we know it before it gets upgraded, so I removed the `publishIri` 👍

bprusinowski: We could need them in case there's a feature to "pin" cube version and disable auto-upgrade behavior, but it's not something that's coming soon (or ever?) :)

app/charts/shared/use-sync-interactive-filters.spec.tsx
```diff
@@ -51,7 +51,7 @@ const getLegacyInteractiveFiltersConfig = () => ({
   },
   dataFilters: {
     active: true,
-    componentIds: [
+    componentIris: [
```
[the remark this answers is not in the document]
ptbrowne: OK understood, it's because it's migrated below. Thanks.

app/graphql/resolvers/rdf.ts
```diff
@@ -32,13 +32,38 @@ import {
...
 
+const IRI_SEPARATOR = "___";
+
+export const joinIris = ({
+  unversionedCubeIri,
+  dimensionIri,
+}: {
+  unversionedCubeIri: string;
+  dimensionIri: string;
+}) => `${unversionedCubeIri}${IRI_SEPARATOR}${dimensionIri}`;
```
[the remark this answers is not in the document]
bprusinowski: Thanks for this suggestion @ptbrowne, I tried it out in https://github.com/visualize-admin/visualization-tool/pull/1849/commits/9de28d57fc52e9f1562302898e879f05579f5d4a and this helped to find a few places and edge cases (that e.g. `joinBy` ids need to be a `string` :)).

app/pages/preview_post.tsx
bprusinowski: I checked the code and looks like in the end this is the only one place that uses this in a React-component context (except of Examples that use a bit different pattern in a component context), all the other ones are inside regular functions, maybe it's fine to not introduce a new concept just for this one page?

app/rdf/query-dimension-values.ts
bprusinowski: It was a `Term`, which gets converted to a `<>` pattern automatically when in `SELECT` context. I refactored this to pass string, so it's more clear 👍

app/charts/index.spec.ts
ptbrowne: But then since we already have the non upgrade cube iri, it would not be an additional field, but more of a boolean "doNotUpgrade" right ?

app/rdf/parse-search-results.ts
bprusinowski: Yeah, maybe this should be a `MAYBE TODO` :) I am not sure (also from performance reasons) if we should fetch additional triple and unversion cube iri while knowing it's not really needed.

I know it would be a "correct" approach, but I am not 100% sure if it's worth pursuing, what do you think?

app/charts/index.spec.ts
bprusinowski: Exactly, I think this would be more clear 👍

app/rdf/parse-search-results.ts
ptbrowne: Ah yes OK I see. I agree, then I think you can remove the TODO and include the "performance reason" bit. 🙇

app/utils/chart-config/versioning.ts
```diff
@@ -1801,6 +1825,26 @@ export const configuratorStateMigrations: Migration[] = [
...
+        dataFilters.filters[
+          splitComponentId(fId).unversionedComponentIri as string
+        ] = dataFilters.filters[fId];
+        delete dataFilters.filters[fId];
+      }
+
+      dataFilters.componentIris = dataFilters.componentIds;
+      delete dataFilters.componentIds;
+      newConfig.dashboardFilters.dataFilters = dataFilters;
+
```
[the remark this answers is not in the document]
bprusinowski: Yeah, I don't like this either, but I think abstracting this would just move this (or very similar) logic to a different place in the codebase? Maybe there could be a visitor pattern that would go through every possible field and return a `[key, value]` pair in case the property contains "componentIri"?

But as this is generally one-off thing and I feel like abstracting this would be a bit complex, maybe it's fine to leave it as it is? 🤔

app/utils/chart-config/versioning.ts
```diff
@@ -916,11 +918,26 @@ export const chartConfigMigrations: Migration[] = [
...
     to: "4.0.0",
     up: async (config, configuratorState) => {
-      let unversionedCubeIri = "";
+      const unversionedCubeIris: string[] = [];
+      const getClosestUnversionedCubeIri = (id: string) => {
+        let score = 0;
+        let unversionedIri = "";
+
+        for (const iri of unversionedCubeIris) {
+          const s = stringSimilarity(id, iri);
```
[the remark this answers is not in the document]
bprusinowski: Thanks @ptbrowne, in fact we don't need to load the charts to the database but can use preview via API feature. I included the testing logic in https://github.com/visualize-admin/visualization-tool/commit/940e1f7582086c5007c99e7a9f889bc95c05b454 – there are a few cases of broken charts that I discover based on this and will fix them tomorrow 🥹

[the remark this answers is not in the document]
bprusinowski: (I ran the testing based on around 500 PROD charts, out of the total of ~2000)

noahonyejese (approved):
@bprusinowski nice work, couldn't find any additional nits 👍

PR #1848, opened 2024-11-12
chore: Clean up of GitHub-related things

bprusinowski:
This PR:
- removes unused Docker-build GitHub Actions,
- corrects README file,
- updates GitHub issue templates,
- updates CODEOWNERS.

cc @adintegra

PR #1855, opened 2024-11-13
Feat/extend supported datasets

### LINKED ISSUE #1799: Inconsistent horizontal axis values in Column-Line Chart type
**Describe the bug** 

When attempting to create a Column-Line graph, the x-axis (time component) displays incorrect data. This issue does not occur with other types of graphs using the same x-axis (time component), suggesting that the data itself is not the problem.
Also, when you hover the mouse over the graph, something (tooltip?) appears to the left but outside the frame and is therefore cut off.

**To Reproduce** 

1. Go to [project](https://visualize.admin.ch/en/browse?dataset=https%3A%2F%2Fhealth.ld.admin.ch%2Ffsvo%2Fmoderhinke_entwicklung%2F2&dataSource=Prod) and start a new visualization.
2. Select 'Column-Line' as chart type
3. Set 'Woche' as the horizontal axis value.
4. Observe the x-axis data.
5. Try creating another type of chart with 'Woche' as the x-axis.
6. Compare the x-axis values in both graphs.

**Expected behavior** 
The x-axis should display the same data across different types of graphs

**Screenshots or video** 

Image
Image

- Visualize environment and version: PROD v4.9.2 and INT 
- Browser and version Microsoft Edge v130.0.2849.46

noahonyejese:
**This PR**
- This PR introduces additional support for new time formats.
- This PR resolves #1799 

**Description**
Our functions only allowed for DD.MM.YY supported dates on datasets causing issue's w/ the dataset mentioned in #1799 . This PR introduces 10 new supported Time formats increasing supported time formats from 5 to 15. Additionally a lot of components where not `formatDate` resulting everything to be stacked on top of each other or not visible.

---
### How to test
1. Since this Dataset is only available on Int or `ssl` development to Int you can only test this PR once deployed
2. Go to this [link](https://int.visualize.admin.ch/en/create/4FPpIV3fglSp?dataSource=Int)
3. Select Column-Line Chart See how we have a bunch of `NaN.NaN.NaN` dates
4. Select they are gone now view image 

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,14 +71,8 @@ yarn dev:ssl # If you are working with the login process
 > redirection, or use the `yarn dev:ssl` command to use HTTPs for the
 > development server. Also, make sure to set the `NEXTAUTH_URL` environment
 > variable to `https://localhost:3000` in your `.env.local` file
-
-> Authentication Requirements Checklist
->
-> - Create .env.loval file
-> - NEXTAUTH_URL=https://localhost:3000
-> - NEXTAUTH_SECRET=Some Secret
-> - Don't forget to import all ADFS\_ from
->   `https://vercel.com/ixt/visualization-tool/settings/environment-variables`
+> make sure to fill in every variable from .env.example to make sure the
+> authentication works
 
 > 👉 In [Visual Studio Code](https://code.visualstudio.com/), you also can run
 > the **default build task** (CMD-SHIFT-B) to start the dev server, database
--- a/README.md
+++ b/README.md
@@ -70,7 +70,15 @@ yarn dev:ssl # If you are working with the login process
 > the app after login. You can either remove the trailing s in the URL after the
 > redirection, or use the `yarn dev:ssl` command to use HTTPs for the
 > development server. Also, make sure to set the `NEXTAUTH_URL` environment
-> variable to `https://localhost:3000` in your `.env.local` file.
+> variable to `https://localhost:3000` in your `.env.local` file
+
+> Authentication Requirements Checklist
+>
+> - Create .env.loval file
+> - NEXTAUTH_URL=https://localhost:3000
+> - NEXTAUTH_SECRET=Some Secret
+> - Don't forget to import all ADFS\_ from
+>   `https://vercel.com/ixt/visualization-tool/settings/environment-variables`
 
 > 👉 In [Visual Studio Code](https://code.visualstudio.com/), you also can run
 > the **default build task** (CMD-SHIFT-B) to start the dev server, database
--- a/app/charts/column/overlay-columns.tsx
+++ b/app/charts/column/overlay-columns.tsx
@@ -3,9 +3,11 @@ import { ComboLineColumnState } from "@/charts/combo/combo-line-column-state";
 import { useChartState } from "@/charts/shared/chart-state";
 import { useInteraction } from "@/charts/shared/use-interaction";
 import { Observation } from "@/domain/data";
+import { useFormatFullDateAuto } from "@/formatters";
 
 export const InteractionColumns = () => {
   const [, dispatch] = useInteraction();
+  const formatDate = useFormatFullDateAuto();
 
   const { chartData, bounds, getX, xScaleInteraction } = useChartState() as
     | ColumnsState
@@ -28,7 +30,7 @@ export const InteractionColumns = () => {
       {chartData.map((d, i) => (
         <rect
           key={i}
-          x={xScaleInteraction(getX(d)) as number}
+          x={xScaleInteraction(formatDate(getX(d))) as number}
           y={0}
           width={xScaleInteraction.bandwidth()}
           height={Math.max(0, chartHeight)}
--- a/app/charts/combo/combo-line-column-state.tsx
+++ b/app/charts/combo/combo-line-column-state.tsx
@@ -160,7 +160,8 @@ const useComboLineColumnState = (
   // Tooltip
   const getAnnotationInfo = (d: Observation): TooltipInfo => {
     const x = getX(d);
-    const xScaled = (xScale(x) as number) + xScale.bandwidth() * 0.5;
+    const xScaled =
+      (xScale(formatDate(x)) as number) + xScale.bandwidth() * 0.5;
     const values = [variables.y.left, variables.y.right]
       .map(({ orientation, getY, label, chartType }) => {
         const y = getY(d);
--- a/app/charts/combo/combo-line-column.tsx
+++ b/app/charts/combo/combo-line-column.tsx
@@ -9,6 +9,7 @@ import { ComboLineColumnState } from "@/charts/combo/combo-line-column-state";
 import { useChartState } from "@/charts/shared/chart-state";
 import { renderContainer } from "@/charts/shared/rendering-utils";
 import { Observation } from "@/domain/data";
+import { useFormatFullDateAuto } from "@/formatters";
 import { useTransitionStore } from "@/stores/transition";
 
 const Columns = () => {
@@ -26,6 +27,7 @@ const Columns = () => {
   const ref = useRef<SVGGElement>(null);
   const enableTransition = useTransitionStore((state) => state.enable);
   const transitionDuration = useTransitionStore((state) => state.duration);
+  const formatDate = useFormatFullDateAuto();
   const bandwidth = xScale.bandwidth();
   const yColumn = y.left.chartType === "column" ? y.left : y.right;
   const yScale =
@@ -36,7 +38,7 @@ const Columns = () => {
   const renderData: RenderColumnDatum[] = useMemo(() => {
     return chartData.map((d) => {
       const key = getRenderingKey(d);
-      const xScaled = xScale(getX(d)) as number;
+      const xScaled = xScale(formatDate(getX(d))) as number;
       const y = yColumn.getY(d) ?? NaN;
       const yScaled = yScale(y);
       const yRender = yScale(Math.max(y, 0));
@@ -62,6 +64,7 @@ const Columns = () => {
     y0,
     colors,
     bandwidth,
+    formatDate,
   ]);
 
   useEffect(() => {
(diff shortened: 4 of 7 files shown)


bprusinowski:
@noahonyejese dataset selection generally doesn't depend on environment, except we limit data sources to INT and PROD on INT, and to PROD on PROD – every other environment (local, Vercel, TEST) has access to every cube. The one from the issue can be checked with this PR via https://visualization-tool-git-feat-extend-supported-datasets-ixt1.vercel.app/en/browse?previous=%7B%22order%22%3A%22SCORE%22%2C%22search%22%3A%22moderhinke%22%2C%22includeDrafts%22%3Atrue%7D&dataset=https%3A%2F%2Fhealth.ld.admin.ch%2Ffsvo%2Fmoderhinke_entwicklung%2F1&dataSource=Int (you need to enable draft datasets and change data source to INT to find it) :)

README.md
```diff
@@ -70,7 +70,15 @@ yarn dev:ssl # If you are working with the login process
...
-> variable to `https://localhost:3000` in your `.env.local` file.
+> variable to `https://localhost:3000` in your `.env.local` file
+
+> Authentication Requirements Checklist
+>
+> - Create .env.loval file
+> - NEXTAUTH_URL=https://localhost:3000
+> - NEXTAUTH_SECRET=Some Secret
+> - Don't forget to import all ADFS\_ from
+>   `https://vercel.com/ixt/visualization-tool/settings/environment-variables`
```
bprusinowski, line 81: @noahonyejese as discussed briefly in https://github.com/visualize-admin/visualization-tool/pull/1848#discussion_r1837928507, I'd try to not include IXT-specific information in README. For these variables, I added them to `.env.example` already 👍 I think we might have other places where we don't follow this practice, but we'll need to extract them going forward, to make it more open-source :)

So here I'd rephrase it to something like "make sure to fill in every variable from .env.example to make sure the authentication works"?

app/charts/column/overlay-columns.tsx
```diff
@@ -28,7 +30,7 @@ export const InteractionColumns = () => {
       {chartData.map((d, i) => (
         <rect
           key={i}
-          x={xScaleInteraction(getX(d)) as number}
+          x={xScaleInteraction(formatDate(getX(d))) as number}
```
bprusinowski, line 33: @noahonyejese I might be wrong, but I am pretty sure we shouldn't do anything with dates here, `getX` should correctly return an X-position, I think the error might be somewhere else 👀 Column chart can also not be date-based, so then it would also fail.

Also, it looks like hovering is broken [here](https://visualization-tool-git-feat-extend-supported-datasets-ixt1.vercel.app/en/create/new?cube=https://health.ld.admin.ch/fsvo/moderhinke_entwicklung/1&dataSource=Int), see that the tooltip only works on the first bar, and displays in a wrong place 😱

app/charts/combo/combo-line-column-state.tsx
```diff
@@ -160,7 +160,8 @@ const useComboLineColumnState = (
   // Tooltip
   const getAnnotationInfo = (d: Observation): TooltipInfo => {
     const x = getX(d);
-    const xScaled = (xScale(x) as number) + xScale.bandwidth() * 0.5;
+    const xScaled =
+      (xScale(formatDate(x)) as number) + xScale.bandwidth() * 0.5;
```
bprusinowski, line 164: 👍

app/charts/combo/combo-line-column.tsx
```diff
@@ -36,7 +38,7 @@ const Columns = () => {
   const renderData: RenderColumnDatum[] = useMemo(() => {
     return chartData.map((d) => {
       const key = getRenderingKey(d);
-      const xScaled = xScale(getX(d)) as number;
+      const xScaled = xScale(formatDate(getX(d))) as number;
```
bprusinowski, line 41: 💯

app/charts/combo/combo-line-dual-state.tsx
```diff
@@ -146,6 +146,7 @@ const useComboLineDualState = (
   const getAnnotationInfo = (d: Observation): TooltipInfo => {
     const x = getX(d);
     const xScaled = xScale(x);
+    console.log(x, xScaled);
```
bprusinowski, line 149: 🔍

bprusinowski (commented):
Thanks @noahonyejese, good catch with the `formatDate` thing! 📆 

There is one thing I think we need to fix before merging (column charts tooltip, see my comment).

I also wonder if we needed to extend the dates formats, was there a problem with them? If not, maybe we don't need to extend them just for the sake of doing so, I believe they come in a standard format from the Cube Creator 🤔

noahonyejese:
> @noahonyejese dataset selection generally doesn't depend on environment, except we limit data sources to INT and PROD on INT, and to PROD on PROD – every other environment (local, Vercel, TEST) has access to every cube. The one from the issue can be checked with this PR via https://visualization-tool-git-feat-extend-supported-datasets-ixt1.vercel.app/en/browse?previous=%7B%22order%22%3A%22SCORE%22%2C%22search%22%3A%22moderhinke%22%2C%22includeDrafts%22%3Atrue%7D&dataset=https%3A%2F%2Fhealth.ld.admin.ch%2Ffsvo%2Fmoderhinke_entwicklung%2F1&dataSource=Int (you need to enable draft datasets and change data source to INT to find it) :)

Ah okay thanks @bprusinowski didn't know that!

app/charts/column/overlay-columns.tsx
noahonyejese: getX returns date as 2024-10-20 instead of 2024.10.20

bprusinowski: Yes, I am not sure if this would work, but you may also change how getX is created for combo charts, so that it also includes `formatDate`. This way you don't need conditional logic, and can remove all of the formatDate calls in other places, I think it would be better 🤔 

You can see e.g. https://github.com/visualize-admin/visualization-tool/blob/e5170fa14a7f3f23d9da55ff5d059d7bce3f2ae0/app/charts/combo/combo-line-column-state-props.ts#L61-L64 – maybe by extracting `getX` from here like

```typescript
  const { getX: _getX, ...bandXVariables } = useBandXVariables(x, {
    dimensionsByIri,
    observations,
  });
  const formatDate = useFormatFullDateAuto();
  const getX = useCallback((d) => formatDate(_getX(d)), [_getX, formatDate]);
```

? There might be other things to change, maybe also in https://github.com/visualize-admin/visualization-tool/blob/e5170fa14a7f3f23d9da55ff5d059d7bce3f2ae0/app/charts/combo/combo-state.ts#L23-L32, but might be worth trying :)

bprusinowski: In worst case, I guess we could have` chartType === "combo" ? formatDate(getX(d)) : getX(d)`, but it would be good to explore better solutions first 🥹

bprusinowski:
@noahonyejese are you sure it was expecting MM.DD.YY? I think the issue was that we mapped `formatDate` over `xDomain`, and afterwards didn't format the date (thus the issue with dots and dashes), but I'd be surprised if there was a US-date format that was expected 😱

app/charts/column/overlay-columns.tsx
noahonyejese: Yes alright I will try to go this direction

noahonyejese:
> @noahonyejese are you sure it was expecting MM.DD.YY? I think the issue was that we mapped `formatDate` over `xDomain`, and afterwards didn't format the date (thus the issue with dots and dashes), but I'd be surprised if there was a US-date format that was expected 😱


@bprusinowski  I mean I logged everything and I saw it there. Pick Combo-line Chart [here](https://int.visualize.admin.ch/en/create/4FPpIV3fglSp?dataSource=Int) and you'll see some dates are working and some are `NaN.NaN.NaN` if you look closely you don't even have to use `console.log`, the last working date is `10.12.2024` and from `10.13.2024` it says `NaN.NaN.NaN` so I assumed it was having trouble because there is no 13th month, also 

It works now (view the picture above)

bprusinowski:
@noahonyejese just a small FYI, you can't share /create links, you'd need to publish the chart so I can access it (you can try to go to this link in a private browser) 👀 

I hope I'm not too picky, but it looks like it's correct in the data, see a SPARQL query with the full dataset: https://s.zazuko.com/22nfxhH, that's why I doubt that the US-date format can appear here. Did you `console.log `the `observations` to see how it's formatted? Maybe we do something on the server side that makes it fail. It's just that I've never seen a problem like that in Visualize, and here also it looks like the date format is correct (EU) on the data side, so maybe it's another bug we need to find?

PS. It seems to work correctly with the current logic in other places (see the image below), that's why I doubt it's a new formatting issue 🤔 

Screenshot 2024-11-13 at 11 59 43

bprusinowski:
I think I'd try to see why column chart X axis works correctly (without new date formats), and then see what's different in combo charts, to narrow down the issue 👀

noahonyejese:
@bprusinowski Thanks for your insights, maybe that's true I will try to compare it. Maybe it's another issue, but that's still quite a weird because I `console.log(x)` the output was `10.10.2024`, `10.11.2024`

 etc 

Then I console.logged `xScaled` which was `undefined` 

```javascript
 const x = getX(d);
    const xScaled =
      (xScale(formatDate(x)) as number) + xScale.bandwidth() * 0.5;
```

also xScaled was only undefined for values above `10.12.2024` I will further investigate this tho.

bprusinowski:
Thanks @noahonyejese, yeah, but 10.10.2024 and 10.11.2024 are still valid dates with our current formatting, right? I think it would be super weird if this bug was related to US-date formatting, but only for combo charts, that why I was a bit pushing against adding a new date format options 😅

I think with the comment to add more formats, I was thinking about more formats like "decade", "microsecond", etc, but not about introducing MM.DD.YYYY ones, unless we really see that the data has it defined in such way.

noahonyejese:
> Thanks @noahonyejese, yeah, but 10.10.2024 and 10.11.2024 are still valid dates with our current formatting, right? I think it would be super weird if this bug was related to US-date formatting, but only for combo charts, that's why I was a bit pushing against adding a new date format options 😅
> 
> I think with the comment to add more formats, I was thinking about more formats like "decade", "microsecond", etc, but not about introducing MM.DD.YYYY ones, unless we really see that the data has it defined in such way.

I see, I will investigate this

PR #1879, opened 2024-11-14
feat: Implement Share and Embed on Profile Page

### LINKED ISSUE #1758: Allow User to Copy Share Link from User Profile
At the moment, if the user selects "share" on a published chart in the user profile, they are taken to the published view where they can then click the "share" button, and copy the URL. 
I think this behaviour could be confusing for someone who is using visualize for the first time, as they then have to find the share button in UI of the published view to copy the URL. I think it would be a nice enhancement to allow the user to copy the published chart link,  and possibly even the embedd code directly from the user profile, as oppose to having them first open the published view. 

https://github.com/user-attachments/assets/53e9b63d-59fb-41aa-b4dc-99d5acb0aa85

noahonyejese:
**This PR:** 
- This PR adds the possibility for embedding and sharing on the user profile page e
- This PR resolves #1758 

**Review Hints**
@bprusinowski I am not 100% certain if this is the best way of implementing it. It heavily depends on custom positioning, because the anchor element by itself would not detect the position correctly and place it at the top left of the screen.

- Perhaps there is a better place also to put the `adjustPopoverPosition`.

---

### How to test 
1. This feature can only be tested on Prod
2. Go to User Profile
3. Click the 3 dots for more on a Published Chart 
4. See that there is a Share button✅ 
5. See that there is a Embed button ✅ 
6. On click of either of those buttons, see how the same popup from the "Published Page" popup ✅ 

---
<details>
  <summary><strong>Images from my Local Machine</strong></summary>

image
image
image

</details>

### DIFF
--- a/app/components/publish-actions.tsx
+++ b/app/components/publish-actions.tsx
@@ -56,12 +56,13 @@ export const TriggeredPopover = (props: TriggeredPopoverProps) => {
     setAnchorEl(trigger);
   }, [trigger]);
 
-  const [ref, width, height] = useResizeObserver<HTMLElement>();
+  const [ref, width, height] = useResizeObserver<HTMLDivElement>();
 
   return (
     <>
       {renderTrigger && renderTrigger(setAnchorEl)}
       <Popover
+        ref={ref}
         open={!!anchorEl}
         anchorEl={anchorEl}
         {...popoverProps}
@@ -75,7 +76,7 @@ export const TriggeredPopover = (props: TriggeredPopoverProps) => {
         }
         onClose={() => setAnchorEl(undefined)}
       >
-        <Box ref={ref}>{children}</Box>
+        {children}
       </Popover>
     </>
   );
--- a/app/components/publish-actions.tsx
+++ b/app/components/publish-actions.tsx
@@ -56,7 +56,7 @@ export const TriggeredPopover = (props: TriggeredPopoverProps) => {
     setAnchorEl(trigger);
   }, [trigger]);
 
-  const [ref, width, height] = useResizeObserver<HTMLElement>();
+  const [ref, width, height] = useResizeObserver<HTMLDivElement>();
 
   return (
     <>
--- a/app/components/publish-actions.tsx
+++ b/app/components/publish-actions.tsx
@@ -56,13 +56,12 @@ export const TriggeredPopover = (props: TriggeredPopoverProps) => {
     setAnchorEl(trigger);
   }, [trigger]);
 
-  const [ref, width, height] = useResizeObserver<HTMLDivElement>();
+  const [ref, width, height] = useResizeObserver<HTMLElement>();
 
   return (
     <>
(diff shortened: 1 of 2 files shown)


app/components/publish-actions.tsx
```diff
@@ -44,21 +45,37 @@ type TriggeredPopoverProps = {
...
 };
 
 export const TriggeredPopover = (props: TriggeredPopoverProps) => {
-  const { children, renderTrigger, popoverProps } = props;
+  const { children, renderTrigger, popoverProps, trigger } = props;
   const [anchorEl, setAnchorEl] = useState<Element | undefined>();
+
+  useEffect(() => {
+    setAnchorEl(trigger);
+  }, [trigger]);
```
bprusinowski, line 57: Nit: I think you can achieve the same behavior with an initializer function like:

```typescript
const [anchorEl, setAnchorEl] = useState<Element | undefined>(() => trigger);
```

or maybe even

```typescript
const [anchorEl, setAnchorEl] = useState<Element | undefined>(trigger);
```
?

app/login/components/profile-tables.tsx
```diff
@@ -408,3 +452,18 @@ const ProfileVisualizationsRow = (props: {
     </TableRow>
   );
 };
+
+const adjustPopoverPosition = (
```
bprusinowski, line 456: Nit: `getAdjustedPopoverPosition`?

app/login/components/profile-tables.tsx
```diff
@@ -190,6 +197,9 @@ const ProfileVisualizationsRow = (props: {
   const removeConfigMut = useMutate(removeConfig);
   const updateConfigMut = useMutate(updateConfig);
 
+  const [openShare, setOpenShare] = useState<HTMLElement | undefined>();
```
bprusinowski, line 200: Nit: `shareEl`, `embedEl`?

app/components/publish-actions.tsx
```diff
@@ -56,13 +56,12 @@ export const TriggeredPopover = (props: TriggeredPopoverProps) => {
     setAnchorEl(trigger);
   }, [trigger]);
 
-  const [ref, width, height] = useResizeObserver<HTMLDivElement>();
+  const [ref, width, height] = useResizeObserver<HTMLElement>();
```
bprusinowski, line 59: Nit: Box returns `HTMLDivElement` by default, should be safe to keep that type 👍

bprusinowski (approved):
LGTM, thanks! I couldn't test the preview, but the code looks good :)

app/components/publish-actions.tsx
noahonyejese: @bprusinowski Not quite sure why but I tested both before I created this PR and it didn't work

app/login/components/profile-tables.tsx
noahonyejese: thanks -

PR #1885, opened 2024-11-15
Combo Chart Axis Title improvements

### LINKED ISSUE #1878: ✓✓ Long Titles Wrap but Cover the Chart
To reproduce, see https://int.visualize.admin.ch/en/v/wVfk4PSYRdYk?dataSource=Prod on a smaller screen – the Y axis titles are wrapped, but the chart is not shifted to the bottom, so they are covering the chart.

Image

noahonyejese:
**This PR**
- This PR prevents all kind of overlapping of axis titles in combo Charts
- This includes Text overlapping and Label overlapping the chart itself
- This PR resolves #1878 

---

### How to test
1. Go to this link
2. Create a new visualization (f.e Einmalvergütung für Photovoltaikanlagen) - works better with this 
3. Select Column + Line Chart
4. Select these f.e
image
5. These are long Axis Title's allowing you to view the full potential 
6. Resize the screen to any size you'd like 
7. See how text is not overflowing the label ✅
8. See how label is not covering the chart ✅

---

### DIFF
--- a/app/charts/combo/combo-line-column-state.tsx
+++ b/app/charts/combo/combo-line-column-state.tsx
@@ -45,6 +45,8 @@ import { useIsMobile } from "@/utils/use-is-mobile";
 
 import { ChartProps } from "../shared/ChartProps";
 
+import { TITLE_VPADDING } from "./combo-line-container";
+
 export type ComboLineColumnState = CommonChartState &
   ComboLineColumnStateVariables &
   InteractiveXTimeRangeState & {
@@ -147,8 +149,31 @@ const useComboLineColumnState = (
         TICK_PADDING
     )
   );
+
+  const axisTitleLeft = variables.y.left.label;
+  const axisTitleRight = variables.y.right.label;
+  const axisTitleWidthLeft =
+    getTextWidth(axisTitleLeft, { fontSize: TICK_FONT_SIZE }) + TICK_PADDING;
+  const axisTitleWidthRight =
+    getTextWidth(axisTitleRight, { fontSize: TICK_FONT_SIZE }) + TICK_PADDING;
+  const overLappingTitles = axisTitleWidthLeft + axisTitleWidthRight > width;
+  const overLappingAmount = (axisTitleWidthLeft + axisTitleWidthRight) / width;
+
+  const axisTitleAdjustment =
+    (overLappingTitles
+      ? TICK_FONT_SIZE * Math.ceil(overLappingAmount)
+      : TICK_FONT_SIZE + TITLE_VPADDING) *
+      2 -
+    TICK_FONT_SIZE * 2;
+
+  const topMarginAxisTitleAdjustment = 50 + axisTitleAdjustment;
   const right = Math.max(maxRightTickWidth, 40);
-  const margins = getMargins({ left, right, bottom });
+  const margins = getMargins({
+    left,
+    right,
+    bottom,
+    top: topMarginAxisTitleAdjustment,
+  });
   const bounds = useChartBounds(width, margins, height);
   const { chartWidth, chartHeight } = bounds;
   const xScales = [xScale, xScaleTime, xScaleTimeRange];
--- a/app/utils/text-adjustments.ts
+++ b/app/utils/text-adjustments.ts
@@ -1,0 +1,46 @@
+import { getTextWidth } from "./get-text-width";
+
+const WORD_SPLITTING_THRESHOLD = 0.75;
+
+export const splitTextByWidth = (
+  text: string,
+  width: number,
+  options: { fontSize: number }
+): string => {
+  const { fontSize } = options;
+  const words = text.split(" ");
+  let currentLine = "";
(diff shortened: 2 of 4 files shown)


app/utils/text-adjustments.ts
```diff
@@ -0,0 +1,46 @@
+import { getTextWidth } from "./get-text-width";
+
+const WORD_SPLITTING_THRESHOLD = 0.75;
+
+export const splitTextByWidth = (
+  text: string,
+  width: number,
+  options: { fontSize: number }
+): string => {
```
bprusinowski, line 9: @noahonyejese I wonder if you considered adding a `word-break` (`break-word`, `break-all`) CSS property to the axis titles? As they are regular `span`s, it's possible to get this behavior without having a custom wrapping function 🤔

app/charts/combo/combo-line-column-state.tsx
```diff
@@ -147,8 +149,31 @@ const useComboLineColumnState = (
...
+
+  const axisTitleAdjustment =
+    (overLappingTitles
+      ? TICK_FONT_SIZE * Math.ceil(overLappingAmount)
+      : TICK_FONT_SIZE + TITLE_VPADDING) *
+      2 -
+    TICK_FONT_SIZE * 2;
+
+  const topMarginAxisTitleAdjustment = 50 + axisTitleAdjustment;
   const right = Math.max(maxRightTickWidth, 40);
```
bprusinowski, line 170: Nit: I think this could be extracted, as the logic seems to be the same between combo line column and combo line dual charts 👍

app/charts/combo/combo-line-column-state.tsx
```diff
@@ -147,8 +149,31 @@ const useComboLineColumnState = (
...
   );
+
+  const axisTitleLeft = variables.y.left.label;
+  const axisTitleRight = variables.y.right.label;
+  const axisTitleWidthLeft =
+    getTextWidth(axisTitleLeft, { fontSize: TICK_FONT_SIZE }) + TICK_PADDING;
+  const axisTitleWidthRight =
+    getTextWidth(axisTitleRight, { fontSize: TICK_FONT_SIZE }) + TICK_PADDING;
+  const overLappingTitles = axisTitleWidthLeft + axisTitleWidthRight > width;
+  const overLappingAmount = (axisTitleWidthLeft + axisTitleWidthRight) / width;
```
bprusinowski, line 160: Nit: I wonder if we couldn't get this in a more bullet-proof way by adding a `getTextHeight` function? Here we'd need to actually add an element to the DOM, constrain it and measure the height, as it's not really possible to do in the same way with `canvas` :/

bprusinowski (commented):
Thank you @noahonyejese for taking a look at this 🙇‍♂️ 

I checked the deployment preview and it looks like the titles are still overlapping the chart? 🤔 Also, maybe we can achieve the same behavior with CSS `break-word` property? Let me know what you think :)

Screenshot 2024-11-15 at 16 33 47

app/charts/combo/combo-line-column-state.tsx
noahonyejese: yup

sosiology:
@noahonyejese thank you for preparing. I check on the deployment preview https://visualization-tool-git-fix-fix-minor-bugs-ixt1.vercel.app/en/v/MNoS3_CeyAIv?dataSource=Prod and it appears like the bottom of the boxes are cut off. 
image

app/charts/combo/combo-line-column-state.tsx
noahonyejese: @bprusinowski I see, this makes total sense should we handle this in a different step or try to approach it in this PR?

noahonyejese:
> @noahonyejese thank you for preparing. I check on the deployment preview https://visualization-tool-git-fix-fix-minor-bugs-ixt1.vercel.app/en/v/MNoS3_CeyAIv?dataSource=Prod and it appears like the bottom of the boxes are cut off. image

@sosiology true I have noticed that at a certain screen size the bottom gets cut off a bit

app/charts/combo/combo-line-column-state.tsx
bprusinowski: I think it makes sense to do it in this PR to keep things close to each other, I see this as a part of implementation of the wrapping behavior improvements 👍

Depending on how complex this turns out to be, there could also be a new issue. I'd investigate for ~0.5h to see if it's a huge change and make a decision based on this, if you feel like diving deeper into this :)

bprusinowski (approved):
Cool, thank you @noahonyejese 🚀

sosiology:
Nice @noahonyejese LGTM! 

Tested with this chart and reduced browser window and labels still look good! https://visualization-tool-git-fix-fix-minor-bugs-ixt1.vercel.app/en/v/Ap4_IQYbBo0i?dataSource=Prod. Once the PR is merged i will check again on TEST

PR #1889, opened 2024-11-18
feat: Adds basic CSP configuration

adintegra:


adintegra:
@wereHamster, making these changes primarily due to the BAFU's report, but also generally, to improve overall security. Do you think the added headers make sense?

ptbrowne:
I tried with https://www.npmjs.com/package/graphql-constraint-directive but could not make it work and it seems the directive only worked if the field was nested inside an object, which would have required a refactor of places using datasourceUrl.

Instead, I used a custom scalar, which also can be used for providing additional validation rules on existing scalars (see https://www.apollographql.com/docs/apollo-server/schema/custom-scalars).

adintegra:
Thanks @ptbrowne! Looks like this would be a good solution 👍🏼

Did you have issues with `graphql-constraint-directive` because of the version maybe? I initially missed the note about having to use v2 with graphql v15.

ptbrowne:
I do not know, I did not see this note either, could be. Did it work for you ?

app/domain/datasource/urls.ts
```diff
@@ -0,0 +1,20 @@
...
+const allowedDataSourceUrls = allowedSources.map((o) => o.value.split("+")[1]);
+
+export type DataSourceUrl = string & {};
+
+export const isDataSourceUrlAllowed = (url: string): url is DataSourceUrl => {
+  if (typeof url === "string" && allowedDataSourceUrls.includes(url)) {
+    return true;
+  }
+  return false;
+};
```
bprusinowski, line 20: Super-nit: could be simplified to:

```typescript
export const isDataSourceUrlAllowed = (url: string): url is DataSourceUrl => {
   return typeof url === "string" && allowedDataSourceUrls.includes(url);
}
```

app/graphql/resolvers/index.ts
```diff
@@ -119,6 +122,34 @@ export const resolveMeasureType = (
...
+  parseValue: datasourceUrlValue,
+  serialize: datasourceUrlValue,
+  parseLiteral(ast) {
+    if (ast.kind === Kind.INT) {
+      return datasourceUrlValue(ast.value);
+    }
+
+    throw datasourceValidationError();
+  },
+});
```
bprusinowski, line 150: 👍

app/next.config.js
```diff
@@ -40,15 +40,38 @@ module.exports = withPreconstruct(
...
-              {
-                key: "X-Robots-Tag",
-                value: "noindex, nofollow",
-              },
-            ],
+          headers[0].headers.push({
+            key: "X-Robots-Tag",
+            value: "noindex, nofollow",
           });
         }
```
bprusinowski, line 76: @adintegra was this change to simplify the code, or needed to improve security? Just wondering out loud, as it looks that now we miss `"/:path*"` source :)

bprusinowski (approved):
Thanks @adintegra, I am not 100% sure how to test the change, but overall looks good!

One maintenance thing I'd consider would be to add a CHANGELOG entry, mentioning that the security of the application has been improved 🔒

app/graphql/resolvers/index.ts
adintegra: > 👍

This is all @ptbrowne 🎉

PR #1907, opened 2024-11-21
feat: Merging of cubes UI improvements

### LINKED ISSUE #1830: Combining Datasets: Implement Improvements (Usability Testing)
https://gitlab.ldbar.ch/bafu/visualize/-/issues/697

List of tasks for the improvements of the merging of cubes feature: 
https://www.notion.so/interactivethings/Merging-of-cubes-Iteration-3-13530da54302802b9073f45e58169c6d?pvs=4

- [x] Dimension dropdown grouped:
Image
- [x] Filter grouped per dataset:
Image
- [x] Delete “Contains values from Cantons” and termset value from the dropdown dimensions list to reduce complexity for the user:
Image

- [x] Dataset Section on top of the sidebar and rename to Dataset: 
Image

- [x] Explanation of “Concepts” Filter on the Search page: ”Concepts represent values that can be shared across different dimensions and datasets”.:
Image

- [x] Improve default settings of charts as soon as a second dataset is added. [Based on the discussion underneath](https://github.com/13530da54302802b9073f45e58169c6d?pvs=25#13530da54302803ebdc5fb1ef75f7808). 
→ If possible show dual axis as a default chart type as soon as a second dataset is selected?

@bprusinowski @sosiology I took the tasks in here.

bprusinowski:
Closes #1830

This PR:
- removes termsets from merging of cubes modal dropdown,
- renames "Contains values from" to "Join by" when hovering over a shared dimension tag in merging of cubes modal,
- prefers dual-line and column-line chart before multi-line chart and tries to initialize the dual-line chart with measures from different charts, if other conditions are met (e.g. that dimensions have different units),
- adds an info tooltip to explain what `Concepts` are on the browse page,
- groups filters and dropdown options by cube, in case we have more than one.

## How to test
1. Go to [this link](https://visualization-tool-git-feat-merging-of-cubes-ui-imp-b426b5-ixt1.vercel.app/en/browse?dataSource=Prod).
2. ✅ Scroll a bit down to see that there's an icon tooltip next to `Concepts` navigation menu.
3. Search for "photo" and select `Einmalvergütung für Photovoltaikanlagen` dataset.
4. Start a visualization.
5. ✅ Go to Vertical Axis menu and open a dropdown. See that the popup menu looks the same as before.
6. Go back to previous view (`Back to main`).
7. Click on `Add dataset`.
8. ✅ Open dimensions dropdown and see that termsets are not longer visible in the popup menu.
9. See a second row (`Median electricity tariff per canton` dataset) and hover over "Canton" chip.
10. ✅ See that the text says "Joined by" Cantons instead of "Contains values from" Cantons.
11. Scroll down a bit and proceed with merging the `International comparison: Expenditure of general government by function (COFOG)` dataset.
12. ✅ See that after merging, chart type changed to dual-axis line chart and that selected dimensions come from different cubes.
13. ✅ Scroll down the left panel and see that the filters are grouped by cubes.
14. Switch chart type to column chart.
15. Open Horizontal Axis field.
16. ✅ Open dropdown and see that dimensions are grouped by cube (including indication of `joinBy` dimensions).

### DIFF
--- a/app/charts/index.ts
+++ b/app/charts/index.ts
@@ -136,9 +136,9 @@ function getChartTypeOrder({ cubeCount }: { cubeCount: number }): ChartOrder {
     pie: 4,
     map: 5,
     table: 6,
-    comboLineSingle: 7 + multiCubeBoost,
-    comboLineDual: 8 + multiCubeBoost,
-    comboLineColumn: 9 + multiCubeBoost,
+    comboLineDual: 7 + multiCubeBoost,
+    comboLineColumn: 8 + multiCubeBoost,
+    comboLineSingle: 9 + multiCubeBoost,
   };
 }
 
@@ -1924,7 +1924,6 @@ const adjustSegmentSorting = ({
   return newSorting;
 };
 
-// Helpers
 export const getPossibleChartTypes = ({
   dimensions,
   measures,
@@ -1994,6 +1993,7 @@ export const getPossibleChartTypes = ({
   }
 
   const chartTypesOrder = getChartTypeOrder({ cubeCount });
+
   return chartTypes
     .filter(
       (d) =>
--- a/app/charts/index.ts
+++ b/app/charts/index.ts
@@ -1457,24 +1457,24 @@ const chartConfigsAdjusters: ChartConfigsAdjusters = {
           (d) => d.id === numericalMeasureIds[0]
         ) as NumericalMeasure;
         let rightMeasureId: string | undefined;
-        const getLeftMeasure = (id: string) => {
+        const getLeftMeasure = (preferredId: string) => {
+          const preferredLeftMeasure = numericalMeasures.find(
+            (d) => d.id === preferredId
+          ) as NumericalMeasure;
+
           if (isAddingNewCube) {
             const rightMeasure = numericalMeasures.find(
               (d) => d.id === rightMeasureId
             );
-
-            // We're trying to find a measure that comes from a different cube,
-            // to better showcase the fact that we have two cubes now.
-            return (numericalMeasures.find(
+            const overrideLeftMeasure = numericalMeasures.find(
               (d) =>
                 d.cubeIri !== rightMeasure?.cubeIri &&
                 d.unit !== rightMeasure?.unit
-            ) ??
-              numericalMeasures.find((d) => d.id === id)) as NumericalMeasure;
+            );
+
+            return overrideLeftMeasure ?? preferredLeftMeasure;
           } else {
-            return numericalMeasures.find(
-              (d) => d.id === id
-            ) as NumericalMeasure;
+            return preferredLeftMeasure;
           }
         };
 
--- a/app/charts/index.ts
+++ b/app/charts/index.ts
@@ -684,11 +684,13 @@ export const getChartConfigAdjustedToChartType = ({
   newChartType,
   dimensions,
   measures,
+  isAddingNewCube,
 }: {
   chartConfig: ChartConfig;
   newChartType: ChartType;
   dimensions: Dimension[];
   measures: Measure[];
+  isAddingNewCube?: boolean;
 }): ChartConfig => {
   const oldChartType = chartConfig.chartType;
   const initialConfig = getInitialConfig({
@@ -715,6 +717,7 @@ export const getChartConfigAdjustedToChartType = ({
     newChartConfig: initialConfig,
     dimensions,
     measures,
+    isAddingNewCube,
   });
 };
 
@@ -727,6 +730,7 @@ const getAdjustedChartConfig = ({
   newChartConfig,
   dimensions,
   measures,
+  isAddingNewCube,
 }: {
   path: string;
   field: Object;
@@ -736,6 +740,7 @@ const getAdjustedChartConfig = ({
   newChartConfig: ChartConfig;
   dimensions: Dimension[];
   measures: Measure[];
+  isAddingNewCube?: boolean;
 }) => {
   // For filters & segments we can't reach a primitive level as we need to
   // pass the whole object. Table fields have an [id: TableColumn] structure,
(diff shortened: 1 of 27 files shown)


app/configurator/components/dataset-control-section.tsx
```diff
@@ -86,19 +86,12 @@ const DatasetRow = ({
...
-  const ref = useRef<HTMLDivElement>(null);
   const locale = useLocale();
   const client = useClient();
   const classes = useStyles();
   const [state, dispatch] = useConfiguratorState(isConfiguring);
   const [loading, setLoading] = useState(false);
 
-  useEffect(() => {
-    if (added && ref.current) {
-      ref.current?.scrollIntoView({ behavior: "smooth" });
```
noahonyejese, line 98: Nice

app/components/metadata-panel.tsx
```diff
@@ -405,7 +405,7 @@ const CubesPanel = ({
 }) => {
   const classes = useOtherStyles();
   const locale = useLocale();
-  const cubes = chartConfig.cubes.map((x) => ({ iri: x.iri }));
+  const cubes = chartConfig.cubes.map((cube) => ({ iri: cube.iri }));
```
noahonyejese, line 408: Thanks - 😅

app/configurator/components/add-dataset-dialog.tsx
```diff
@@ -686,6 +692,8 @@ export const DatasetDialog = ({
     typeof searchDimensionOptions | undefined
   >(undefined);
 
+  console.log("selectedSearchDimensions", selectedSearchDimensions);
```
noahonyejese: 👀

app/configurator/components/chart-options-selector.tsx
```diff
@@ -1591,15 +1705,16 @@ const ChartFieldColorComponent = (props: ChartFieldColorComponentProps) => {
...
-          options={measuresOptions}
-          isOptional={optional}
+          options={options}
+          optionGroups={optionGroups}
+          optional={optional}
         />
         {enableUseAbbreviations && (
           <Box sx={{ mt: 1 }}>
             <ChartFieldAbbreviations
               field={field}
```
noahonyejese, line 1715: This File feels a super large, can we split things or is it better to keep everything here in this file?

app/charts/index.ts
```diff
@@ -1444,15 +1450,32 @@ const chartConfigsAdjusters: ChartConfigsAdjusters = {
...
+              (d) => d.id === rightMeasureId
+            );
+
+            // We're trying to find a measure that comes from a different cube,
+            // to better showcase the fact that we have two cubes now.
+            return (numericalMeasures.find(
+              (d) =>
+                d.cubeIri !== rightMeasure?.cubeIri &&
+                d.unit !== rightMeasure?.unit
+            ) ??
```
noahonyejese, line 1472: Maybe restructuring this into variables before returning would help, and maybe even prevent the need for comments? - What do you think?

e2e/selectors.ts
```diff
@@ -111,15 +111,9 @@ export const createSelectors = ({ screen, page, within }: Ctx) => {
...
-          .locator(`[data-chart-loaded="true"]`)
-          .waitFor({ timeout: 40 * 1000, ...options });
-        const hasMap = (await page.locator("[data-map-loaded]").count()) > 0;
-        if (hasMap) {
-          await page.locator('[data-map-loaded="true"]');
-          await sleep(500); // Waits for tile to fade in
-        }
+      loaded: async () => {
+        await page.waitForLoadState("networkidle");
+        await sleep(1000);
```
noahonyejese, line 116: ls there a particular reason we are sleeping here for 1s?

noahonyejese (approved):
Nice work, seems like a lot has changed, I have also tested it and it works fine ✅ - there a few nits but other than that LGTM.

noahonyejese:
@bprusinowski 

Noticed that when resizing the page to a certain screen size it does this. This might maybe also be a different issue.

https://github.com/user-attachments/assets/61bfbce4-0c77-4a69-8023-0c1fb16dab15

e2e/selectors.ts
bprusinowski: Yes, there are two things that come to my mind, one is that some charts have enter animation (e.g. bar charts) and another is map that take a bit of time (previously we had 500ms) for the tiles to fade in.

Generally it shouldn't really be needed, but some tests interact with the charts themselves, like tooltip tests, so we need to wait for the elements to be visible 👍 I'll add a comment directly in the code so it's clear :)

noahonyejese:
LGTM 🎉

app/charts/index.ts
bprusinowski: Yes, good point – I tried to make it more clear in https://github.com/visualize-admin/visualization-tool/pull/1907/commits/f7ad94bc610a2de5db6857f494c7ca2e3ed35061 🥹

app/configurator/components/chart-options-selector.tsx
bprusinowski: Yeah, that's true 😅 I think we could split things up, I'll create a new issue for this that we can pick up maybe during Maintenance Quarterly or in case we have some "downtime", as I expect this would take a bit of time to extract all the components into respective files and wouldn't bring any new features or improvements to the end users – but for sure would make our lives easier, so we should tackle it at some point 💯

bprusinowski: See https://github.com/visualize-admin/visualization-tool/issues/1908 :)

bprusinowski:
@noahonyejese thanks for testing and uncovering a bug 🙇‍♂️ I don't think it's related to this PR, looks like a problem with `ResizeObserver` & appearing / disappearing browser scrollbar 🤔 

I tried to reproduce both in Safari and Chrome, but the scrollbar didn't behave like this and I couldn't recreate the bug on my side. Could you open a new issue and describe reproduction steps, so we can investigate further?

PR #1911, opened 2024-11-25
Fix/fix pie chart tooltip alignment

noahonyejese:
**This PR:**
- This PR handles the Pie Chart tooltip issue for smaller screens
- It increases the Pie Chart size on Mobile Screens
- It moves the tooltip on top of the Chart
- It removes the triangle on the tooltip, therefore emphasizing the highlighting
- this PR resolves issue # 111 

### How to test
1. Go to this link 
2. Create a visualization with using a Pie Chart 
3. Publish your visualization 
4. Test the tooltip

---

CC: @KerstinFaye

### DIFF
--- a/app/charts/pie/pie.tsx
+++ b/app/charts/pie/pie.tsx
@@ -9,6 +9,7 @@ import { useInteraction } from "@/charts/shared/use-interaction";
 import { Observation } from "@/domain/data";
 import { useTransitionStore } from "@/stores/transition";
 import useEvent from "@/utils/use-event";
+import { useIsMobile } from "@/utils/use-is-mobile";
 
 export const Pie = () => {
   const { chartData, getPieData, getSegment, colors, bounds, getRenderingKey } =
@@ -19,7 +20,11 @@ export const Pie = () => {
   const [, dispatch] = useInteraction();
   const ref = useRef<SVGGElement>(null);
 
-  const maxSide = Math.min(chartWidth, chartHeight) / 2;
+  const isMobile = useIsMobile();
+
+  const maxSide = isMobile
+    ? Math.min(chartWidth, chartHeight)
+    : Math.min(chartWidth, chartHeight) / 2;
 
   const innerRadius = 0;
   const outerRadius = maxSide;
--- a/app/charts/shared/interaction/tooltip-box.tsx
+++ b/app/charts/shared/interaction/tooltip-box.tsx
@@ -30,6 +30,7 @@ export interface TooltipBoxProps {
   x: number | undefined;
   y: number | undefined;
   placement: TooltipPlacement;
+  withTriangle?: boolean;
   margins: Margins;
   children: ReactNode;
 }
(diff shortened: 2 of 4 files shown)


app/charts/pie/pie.tsx
```diff
@@ -19,7 +20,11 @@ export const Pie = () => {
   const [, dispatch] = useInteraction();
   const ref = useRef<SVGGElement>(null);
 
-  const maxSide = Math.min(chartWidth, chartHeight) / 2;
+  const isMobile = useIsMobile();
+
+  const maxSide = isMobile
+    ? Math.min(chartWidth, chartHeight)
+    : Math.min(chartWidth, chartHeight) / 2;
```
bprusinowski, line 27: @noahonyejese I think this is causing a bug that makes the pie chart grow beyond the available space, see the screenshot (you can reproduce by reducing the window size to a mobile-desktop breakpoint and seeing that it's correct on desktop, but broken just before the desktop breakpoint).

Screenshot 2024-11-25 at 12 23 55

I think it would be safest to revert this change, as handling this correctly would probably mean playing around with chart's aspect ratio, or some other related things higher up in the code hierarchy 👍

app/charts/shared/interaction/tooltip-box.tsx
```diff
@@ -128,8 +129,9 @@ export const TooltipBox = ({
   placement,
   margins,
   children,
+  withTriangle = true,
 }: TooltipBoxProps) => {
-  const triangle = mkTriangle(placement);
+  const triangle = withTriangle ? mkTriangle(placement) : null;
```
bprusinowski, line 134: 👍

bprusinowski (commented):
Thanks @noahonyejese, I think it will be good without the pie chart size adjustment (or another way of doing it to prevent a cut-off bug) 👍 

Also, I wonder if the tooltip could be "centered" when on mobile? See that now it jumps a bit from left to right, I think it could be a bit nicer to simply keep it in the middle, but maybe @KerstinFaye also has an opinion here :)

https://github.com/user-attachments/assets/814d6786-5305-4a71-89c9-1549be559015

noahonyejese:
@bprusinowski I adjusted the size back to before, however it is really small. Additionally the tooltip is now completely centered along the X-axis.

CC: @KerstinFaye

bprusinowski (approved):
LGTM, thanks @noahonyejese 🙇‍♂️ 

I remember we previously had an aspect ratio of 1 for pie charts, but aligned it to be same as other charts (2 / 5) some time ago as it was too big in the editor mode. Maybe it needs a more sophisticated "fix", but I think we'd need to discuss the overarching solution taking all the places into account (editor, dashboard, mobile, etc) 👍

bprusinowski:
PS. Would be nice to add a CHANGELOG entry 😄

noahonyejese:
@bprusinowski I agree, lets tackle this separately - I believe It's already in the CHANGELOG summed up in one point, or do you think I should mentioned every issue / bug fixed individually. Idk but I think that would be a lot - What do you think?

bprusinowski:
Ah, I didn't see an entry in this PR, didn't know it was already added :) I think it's good like this, thanks!

sosiology:
I checked the tooltip (also on my mobile device) and it LGTM. Thank you @noahonyejese and @bprusinowski

noahonyejese:
This could turn out to be a bit tricky, I will try to tackle this asap. @sosiology

sosiology:
Thanks for the quick feedback @noahonyejese as this is an issue identified by us - this improvement would have lower priority than the color palettes, maybe something to look into while you are waiting for feedback.

PR #1912, opened 2024-11-25
fix: undefined and null values get replaced w/ "-"

### LINKED ISSUE #1888: PROD : Visualize dispay Value = "0"  instead of ‘no value’.
When I export the Visualize csv file for following Visualization https://www.visualize.admin.ch/de/v/-u-BR-bpgybi, 
I notice that Visualize exports a Zero value for the LF1 time reference, whereas there is no value for the LFI1 period.
This is correct in Lindas.

1. Open https://www.visualize.admin.ch/de/v/-u-BR-bpgybi
2. Export csv of the Chart
3. Open SPARQL-Query
4. See error : Value 0 is not the same as "No value". It must be fixed in Visualize (for download chart AND dataset) as soon as possible.

- Visualize environment and version: PROD : v4.9.3 ([4f0db71]
- Browser and version : edge

noahonyejese:
This PR: 
- This PR ensures that downloaded data files format `undefined` and `null` to `"-"` instead of `"0"`
- -
- This PR resolves #1888 

### How to test
1. Go to this [link](https://www.visualize.admin.ch/de/v/-u-BR-bpgybi)
2. Create a visualization w/ this Data => `LFI: Themen nach Baumart (5 Klassen)`
3. Publish Visualization 
4. Click on Dataset in the bottom left corner 
5. Click on Download data 
6. Click CSV 
7. Scroll through the file see how there are a bunch of zeros ❌

_repeat steps 1 - 6 with this [link](https://visualization-tool-git-fix-zero-for-undefined-values-ixt1.vercel.app/)_ 

8. See how some of the zeros, were actually undefined values now they are replaced with `"-" `dashes ✅

---

CC: @sosiology

### DIFF
--- a/app/components/data-download.tsx
+++ b/app/components/data-download.tsx
@@ -35,6 +35,7 @@ import {
 } from "@/domain/data";
 import {
   dateFormatterFromDimension,
+  formatIdentity,
   getFormatFullDateAuto,
   getFormattersForLocale,
 } from "@/formatters";
@@ -105,6 +106,7 @@ const prepareData = ({
     sortingType: "byAuto",
     sortingOrder: "asc",
   };
+
   const sorters = sortedComponents.map<
     [string, ReturnType<typeof makeDimensionValueSorters>]
   >((d) => {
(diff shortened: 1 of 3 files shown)


app/components/data-download.tsx
```diff
@@ -126,23 +128,39 @@ const prepareData = ({
...
-            ...{ [makeColumnLabel(col)]: parser(obs[key] as string) },
-          }
-        : acc;
+      if (!col) return acc;
+
+      const value = formattedObs[key];
+      let parsedValue;
+
+      if (value?.toString() === "–") {
+        parsedValue = "-";
```
bprusinowski, line 152: Nit: Would be nice to extract "-" to a constant in the formatters file, so we don't need to go to `formatIdentity` function definition in another file and see that it returns a hard-coded "-" in order to make this comparison. What do you think?

bprusinowski (approved):
LGTM, thanks @noahonyejese 🙇‍♂️

app/components/data-download.tsx
noahonyejese: sure!

PR #1919, opened 2024-11-27
fix: Component id separator

### LINKED ISSUE #1918: v5.0.1 on INT: Area-Chart: Segmentation doesn't work as expected
Since v5.0.1 was deployed on INT, the area chart no longer works as expected. Although I only selected three out of five categories during segmentation, the area chart shows all five categories.

Image

https://int.visualize.admin.ch/create/_dGpBaksL8sK?dataSource=Int

bprusinowski:
Fixes #1918

This PR changes the usage of `___` component id separator to a much more unique one: `(VISUALIZE.ADMIN_COMPONENT_ID_SEPARATOR)`, as it turns out some cube iris can end with `_`, breaking the separation and joining logic.

I didn't introduce a new chart config migration to replace the old separator with new one, as the application was not deployed to PROD since component ids were introduced. It means that some TEST and INT charts that were created recently might get broken (but as these are testing environments, it should be fine).

## How to test
1. Go to [this link](https://visualization-tool-git-fix-component-id-separator-ixt1.vercel.app/en/browse?previous=%7B%22order%22%3A%22SCORE%22%2C%22search%22%3A%22Moderhinke%22%2C%22includeDrafts%22%3Atrue%7D&dataset=https%3A%2F%2Fhealth.ld.admin.ch%2Ffsvo%2FMH_%2F4&dataSource=Int) and start a new visualization.
2. ✅ See that it's possible to make a chart.
3. Switch to an area chart.
4. Add color segmentation.
5. Make it interactive.
6. Remove one or more filters from the multi-filter panel.
7. ✅ See that the filters were also removed from the chart legend.


sosiology:
@bprusinowski tested on the PR and it looks good to me. Thanks for the quick fix

noahonyejese (changes requested):
However, Vercel Deployment is still failing...

bprusinowski:
@noahonyejese looked like some Sentry error. I re-deployed and seems to work fine 👍

noahonyejese (approved):

PR #1931, opened 2024-12-03
feat: Image download

bprusinowski:


sosiology:
Thank you for this @bprusinowski LGTM! I tested on Chrome and didn't encounter any issues. 
@KerstinFaye please also review.

app/utils/use-screenshot.ts
```diff
@@ -0,0 +1,100 @@
...
+      const a = document.createElement("a");
+      a.href = dataUrl;
+      a.download = `${name}.${type}`;
+      a.click();
+    })
+    .catch((error) => console.error(error))
+    .finally(() => wrapperNode.remove());
+};
+
+export const DISABLE_SCREENSHOT_ATTR_KEY = "data-disable-screenshot";
```
noahonyejese, line 169: 👍

app/charts/map/map.tsx
```diff
@@ -403,6 +403,8 @@ export const MapComponent = () => {
           initialViewState={defaultViewState}
           mapLib={maplibregl}
           mapStyle={mapStyle}
+          // Important so we can take a screenshot of the map
```
noahonyejese, line 407: Do we still need this comment?

app/charts/shared/chart-helpers.tsx
```diff
@@ -287,6 +287,20 @@ export const extractChartConfigComponentIds = ({
...
+  { components }: { components: Component[] }
+) => {
+  const componentIds = extractChartConfigComponentIds({
+    chartConfig,
+    includeFilters: false,
+  });
+
+  return componentIds
+    .map((id) => components.find((component) => component.id === id))
+    .filter(truthy); // exclude potential joinBy components
```
noahonyejese, line 301: Do we still need this comment?

package.json
```diff
@@ -25,9 +25,7 @@
     "lint": "eslint app -c .eslintrc.json",
     "locales:extract": "NODE_ENV=development lingui extract --clean; ./scripts/strip-locale-line-numbers.sh",
     "locales:compile": "lingui compile --verbose",
-    "locales:push": "dotenv -- accent sync --add-translations; yarn locales:extract # putting back file location comments",
-    "locales:pull": "dotenv -- accent export; yarn locales:extract # putting back file location comments",
-    "locales:sync": "yarn locales:pull && yarn locales:push",
+    "locales:sync": "dotenv -- accent sync --add-translations --merge-type=passive --order-by=key; yarn locales:extract # putting back file location comments",
```
noahonyejese, line 28: 🤩

noahonyejese (approved):
LGTM 🎉 Well done @bprusinowski

app/charts/map/map.tsx
bprusinowski: I would say yes, as `preserveDrawingBuffer` is `false` by default and can decrease the performance of the map, the comment should make sure that we don't remove it and unknowingly break screenshotting of the maps :)

package.json
bprusinowski: Should be a bit easier now 🤞

app/charts/shared/chart-helpers.tsx
bprusinowski: Here I would also say yes, as I wouldn't know why we need to use `truthy` for – in this case, we have a connection between using it and `joinBy` dimensions. Maybe it would be better to have something like

```typescript
const excludeJoinByComponent = (c: Component) => truthy(c);
```

but I think a comment should suffice here 👍 It was a bit of a pain to catch all the edge-cases of merging of cubes, I'd rather keep more context in places related to them 🥹

bprusinowski: I think comments that describe a "clear" logic should be generally removed, like this:

```typescript
// Find a component
const component = components.find(c => c.id === id);
```

but when they are related to other, connected parts of the code or generally are not "obvious", it's okay to use them 👍

KerstinFaye:
The whole flow works perfectly. 🎉

PR #1938, opened 2024-12-10
Fix/minor bugs

noahonyejese:
**This PR:** 
- This PR adds a getTextHeight function 
- And slightly improves AxisLabel positioning

cc: @bprusinowski

### DIFF
--- a/app/charts/combo/axis-height-linear-dual.tsx
+++ b/app/charts/combo/axis-height-linear-dual.tsx
@@ -10,7 +10,7 @@ import { useChartState } from "@/charts/shared/chart-state";
 import { useChartTheme } from "@/charts/shared/use-chart-theme";
 import { OpenMetadataPanelWrapper } from "@/components/metadata-panel";
 import { theme } from "@/themes/federal";
-import { getTextWidth } from "@/utils/get-text-width";
+import { getTextHeight, getTextWidth } from "@/utils/get-text-width";
 import { useAxisTitleAdjustments } from "@/utils/use-axis-title-adjustments";
 
 import { TITLE_VPADDING } from "./combo-line-container";
@@ -65,6 +65,9 @@ export const AxisHeightLinearDual = (props: AxisHeightLinearDualProps) => {
     TITLE_HPADDING * (isOverlapping ? Math.floor(overlapAmount) : 2);
 
   const titleWidth = isOverlapping ? axisTitleWidth / overlapAmount : "auto";
+  const axisLabelHeight = getTextHeight(axisTitle, {
+    fontSize: axisLabelFontSize,
+  });
 
   return (
     <>
@@ -73,8 +76,8 @@ export const AxisHeightLinearDual = (props: AxisHeightLinearDualProps) => {
         width={axisTitleWidth + TITLE_HPADDING * 2}
         height={
           (isOverlapping
-            ? axisLabelFontSize * Math.ceil(overlapAmount) + TITLE_VPADDING
-            : axisLabelFontSize + TITLE_VPADDING) *
+            ? axisLabelHeight * Math.ceil(overlapAmount) + TITLE_VPADDING
+            : axisLabelHeight + TITLE_VPADDING) *
             2 +
           TOP_MARGIN
         }
--- a/app/utils/get-text-width.ts
+++ b/app/utils/get-text-width.ts
@@ -14,3 +14,39 @@ export const getTextWidth = (text: string, options: { fontSize: number }) => {
 
   return ctx.measureText(text).width;
 };
+
+export const getTextHeight = (
+  text: string,
+  options: {
+    fontSize: number;
+    width?: number;
+    lineHeight?: number | string;
+  }
+) => {
+  const measureDiv = document.createElement("div");
+
+  measureDiv.style.font = `${options.fontSize}px ${fontFamily}`;
+  measureDiv.style.position = "absolute";
+  measureDiv.style.visibility = "hidden";
+  measureDiv.style.lineHeight = options.lineHeight
+    ? typeof options.lineHeight === "number"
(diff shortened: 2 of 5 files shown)


app/utils/get-text-width.ts
```diff
@@ -14,3 +14,39 @@ export const getTextWidth = (text: string, options: { fontSize: number }) => {
...
+) => {
+  const measureDiv = document.createElement("div");
+
+  measureDiv.style.font = `${options.fontSize}px ${fontFamily}`;
+  measureDiv.style.position = "absolute";
+  measureDiv.style.visibility = "hidden";
+  measureDiv.style.lineHeight = options.lineHeight
+    ? typeof options.lineHeight === "number"
+      ? `${options.lineHeight}px`
+      : options.lineHeight
```
bprusinowski, line 34: Nit: As line-height is [unitless by default](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height), I think maybe it would be good to make the type equal to `string`, so we are forced to define a correct unit and not default to pixels?

Here I think sometimes it's a case to use e.g. 1.5 to indicate 150%, and it would get converted to 1.5px (with string type we'd actually need to write "1.5px"). It's a small thing and maybe too much, but could be worth taking a look at 👍

app/charts/combo/axis-height-linear-dual.tsx
```diff
@@ -65,6 +65,9 @@ export const AxisHeightLinearDual = (props: AxisHeightLinearDualProps) => {
     TITLE_HPADDING * (isOverlapping ? Math.floor(overlapAmount) : 2);
 
   const titleWidth = isOverlapping ? axisTitleWidth / overlapAmount : "auto";
+  const axisLabelHeight = getTextHeight(axisTitle, {
+    fontSize: axisLabelFontSize,
+  });
```
bprusinowski, line 70: @noahonyejese I believe here we need to pass `width: chartWidth / 2`, otherwise the height is computed based on full chart width, which is not correct?

app/charts/combo/axis-height-linear-dual.tsx
```diff
@@ -73,8 +76,8 @@ export const AxisHeightLinearDual = (props: AxisHeightLinearDualProps) => {
         width={axisTitleWidth + TITLE_HPADDING * 2}
         height={
           (isOverlapping
-            ? axisLabelFontSize * Math.ceil(overlapAmount) + TITLE_VPADDING
-            : axisLabelFontSize + TITLE_VPADDING) *
+            ? axisLabelHeight * Math.ceil(overlapAmount) + TITLE_VPADDING
+            : axisLabelHeight + TITLE_VPADDING) *
             2 +
           TOP_MARGIN
```
bprusinowski, line 82: If we add `width` as above (`chartWidth / 2`) to `getTextHeight`, I am pretty sure we don't need this logic :) The goal of `getTextHeight` was to get "correct" height, without a need to multiply or divide things anymore 👍

app/charts/combo/axis-height-linear-dual.tsx
bprusinowski: I think this comment should also be applied to other places where we have two titles that can overlap (combo line column chart) 👀 

Let me know if I missed something!

app/utils/use-axis-title-adjustments.ts
```diff
@@ -33,12 +33,14 @@ export const useAxisTitleAdjustments = ({
   const overlapAmount =
     (axisTitleWidthLeft + axisTitleWidthRight) / containerWidth;
 
+  const axisLabelHeight = getTextHeight(leftAxisTitle, { fontSize });
```
bprusinowski, line 36: If I am not mistaken, here we have a "regular chart", with one title? I think if you passed `width: chartWidth` here, we could remove the `isOverlapping`, `overlapAmount` logic down below 🤔

bprusinowski (commented):
Hey @noahonyejese, thanks for changing this! I think if we tweak the logic to use actual `chartWidth` (or `chartWidth / 2` where applicable), the values should be correct without a need for additional calculations.

If they still are not, we'd need to make sure we pass correct line height, font size, etc to the function – here maybe there would be a way to connect this to MUI typography variants somehow, if the text elements are in fact MUI components 👍

app/utils/use-axis-title-adjustments.ts
noahonyejese: @bprusinowski This is for comboCharts, should I adjust the name to `use-dual-axis-title-adjustments` ?

app/charts/combo/axis-height-linear-dual.tsx
noahonyejese: @bprusinowski I will try to refactor it, I have an Idea...

noahonyejese: image

The problem is stuff like this. This counts as 3 lines I don't why, but (#) doesnt count as a line for some reason. if log the value it says. 3 lines.

bprusinowski: But with this we don't need to know the number of lines, as the text would naturally wrap inside the container so we get correct height back, right?

noahonyejese: the text is within a `foreignObject` which needs the exact height of all lines combined to ensure correct amount of space for the span element.

bprusinowski: Yes, but here I was thinking about getTextHeight method - the flow would be:
- get width of the label (half of chart width, potentially with some padding),
- get text height, passing the width to the function, so that the text can naturally wrap inside the measurement container,
- set foreignObject height using the text height.

I don't see a reason to get number of lines of the text, did I misunderstand something? 🥹

PR #1940, opened 2024-12-16
fix: use standard SPARQL IRI starting with <https://...>

Rdataflow:
`<foo>` leads to unpredictable results
stardog: `<tag:/foo>`
fuseki: `<http://server/unset-base/foo>`
...

therefore let's stick to the SPARQL standard
thus switch to `<https://visualize.admin.ch/foo>`

cc @bprusinowski @kronmar-bafu @zellersabine


Rdataflow:
how to test:
- observe Themes selection panel still appears in cube list view
- observe Themes labels still appers in cube metadata

bprusinowski (approved):
Thanks @Rdataflow for catching this, LGTM!

PR #1943, opened 2025-01-06
fix: Keep color mapping up to date

### LINKED ISSUE #1513: chart shows the additional year, but it doesn't appear in the left-hand filter

bprusinowski:
Fixes #1513

This PR makes sure we keep the `colorMapping` up to date in case new values appear in the cube (only possible when editing a chart that's based on a cube that was updated with a new value in dimension used as a color dimension after the chart was initially made). It also makes sure we keep old colors for existing values when updating the `colorMapping`.

## How to test
1. Go to this link.
2. 123

## How to reproduce
1. Go to this link.
2. 123

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ You can also check the
 
 - Features
   - It's now possible to export charts as images
+- Fixes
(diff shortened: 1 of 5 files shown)


noahonyejese (approved):
🎉

PR #1947, opened 2025-01-07
feat: Added Footer to Profile Page

### LINKED ISSUE #1688: Show Footer in User Profile
I was wondering if it would make sense from a UX perspective to include the footer in the login area. 
I think the footer contains links that could be helpful for users such as "Tutorials" or the "Status" Page link that would benefit the user when they are managing their charts and drafts.

Add the footer to the login area 

Adding a dedicated section in the login with "Resources" or similar to include tutorials and the status page 

**Usecases and impact**
Including this would ensure that the users working with a login has access to these links easily, as I assume they will be less likely to go to the main homepage as the data sets can be accessed directly from the login area via the "Browse all datasets" button in the top right-hand corner

**Additional context**
-

noahonyejese:
**This PR**
- fixes #1688

**Description**
This PR adds the Footer to the Profile Page

### DIFF
--- a/app/login/components/profile-content-tabs.tsx
+++ b/app/login/components/profile-content-tabs.tsx
@@ -1,6 +1,6 @@
 import { t } from "@lingui/macro";
 import { TabContext, TabList, TabPanel } from "@mui/lab";
-import { Box, Tab, Theme, useTheme } from "@mui/material";
+import { Box, Tab, Theme } from "@mui/material";
 import { makeStyles } from "@mui/styles";
 import clsx from "clsx";
 import groupBy from "lodash/groupBy";
@@ -12,6 +12,11 @@ import { useRootStyles } from "@/login/utils";
 import useEvent from "@/utils/use-event";
 
 const useStyles = makeStyles<Theme>((theme) => ({
+  section: {
+    borderBottom: 1,
+    borderColor: "divider",
+    marginTop: theme.spacing(6),
+  },
   tabList: {
     minHeight: "fit-content",
 
@@ -46,7 +51,7 @@ export const ProfileContentTabs = (props: ProfileContentTabsProps) => {
   const handleChange = useEvent((_: React.SyntheticEvent, v: string) => {
     setValue(v);
   });
-  const theme = useTheme();
+
   const rootClasses = useRootStyles();
   const classes = useStyles();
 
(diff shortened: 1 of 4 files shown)


app/login/utils.ts
```diff
@@ -7,20 +7,27 @@ import { HEADER_HEIGHT } from "@/components/header-constants";
...
+    backgroundColor: theme.palette.muted.main,
   },
   sectionContent: {
     width: "100%",
     maxWidth: 1400,
     margin: "0 auto",
   },
   noTooltip: {
-    // Disable native tooltip in Safari
     "&::after": {
```
bprusinowski, line 31: @noahonyejese here I'd also keep the comment to know what problem is being solved with the `noTooltip` class.

app/pages/profile.tsx
```diff
@@ -57,7 +58,10 @@ const ProfilePage = (props: Serialized<PageProps>) => {
     <AppLayout>
       <Box className={rootClasses.root}>
         <ProfileHeader user={user} />
-        <ProfileContentTabs userId={user.id} />
+        <Box sx={{ flexGrow: 1, display: "flex", flexDirection: "column" }}>
+          <ProfileContentTabs userId={user.id} />
+        </Box>
+        <Footer />
```
bprusinowski, line 64: 👍

app/login/components/profile-content-tabs.tsx
```diff
@@ -66,9 +64,15 @@ export const ProfileContentTabs = (props: ProfileContentTabsProps) => {
...
+      <Box className={clsx(rootClasses.section)}>
         <Box className={rootClasses.sectionContent}>
-          <Box sx={{ borderBottom: 1, borderColor: "divider" }}>
+          <Box
+            sx={{
+              borderBottom: 1,
+              borderColor: "divider",
+              marginTop: theme.spacing(6),
+            }}
+          >
```
bprusinowski, line 75: @noahonyejese generally I think we should prefer using `makeStyles` / `useStyles` over `sx` even if only to improve readability of JSX code by having the styling extracted. Here we already are using it, I'd bring the `section` back and even extract the `Box`'s `sx` to e.g. `tabWrapper`, so we have all styles next to each other 👍

bprusinowski (approved):
LGTM, thanks @noahonyejese!

A friendly reminder for the CHANGELOG entry, I've added a PR template in https://github.com/visualize-admin/visualization-tool/pull/1913/files#diff-b2496e80299b8c3150b1944450bd81c622e04e13d15c411d291db0927d75fd6b so we don't forget about it, it also often happens to me 😅

PR #1948, opened 2025-01-07
docs: Improve preview via API documentation

### LINKED ISSUE #1583: Improve calling charts via API documentation
Currently, we link to a TypeScript file with a lot of nested io-ts type definitions – it's quite hard to understand the "resolved" type we actually expect.

We should have a way to either insert resolved TS type directly into the documentation, or think about some other way of making this more accessible.

bprusinowski:
Closes #1583

This PR:
- extracts chart adjusters and utils from `config-types.ts` to separate files, so that we actually only have types and simple type checks in `config-types.ts` (e.g. `isMapChart`),
- adds `generate-json-schema` script that can be run manually via CLI or automatically on build, which generates JSON Schema files for configurator state and chart config types,
- improves preview via API documentation to point to new JSON Schema files, that can be used as a help when wanting to adjust the configs via VS Code autocomplete.

Autocomplete could be tested by opening the [example files](https://github.com/visualize-admin/visualization-tool/tree/docs/better-preview-via-api-docs/app/public/json-schema) in VS Code, once this PR is deployed to PROD (as the schema ids point to PROD instance for simplicity).


noahonyejese (approved):
Nice stuff, thanks @bprusinowski 🎉

Issue #1270, opened 2023-11-15
Switch back to Argos for screenshot testing


bprusinowski:
As we now have a Pro Argos account, we should switch to this service instead of Percy, which has far less monthly screenshots limit.

PR #1949, opened 2025-01-07
feat: Switch to Argos & fix E2E tests

bprusinowski:
Closes https://github.com/visualize-admin/visualization-tool/issues/1270

This PR:
- adds back screenshot testing using Argos (including screenshots of failed tests, so it's much easier to know what and how exactly failed by going to Argos PR build),
- fixes e2e tests,
- fixes "Score" sorting option being not available to select in dataset search results.

I also took another look at e2e tests and managed to fix them (at least they work as of now 😅) – which means from now on we should strive to have "green light" on CI checks before merging. We'll have to see how stable they are in time, as in the past both cubes used to change often and LINDAS was very unstable, which improved over the years 🤞

If they will break often, we might need to re-record HAR traces and enable HAR on CI, but for now I'd check with how well it works by connecting directly to LINDAS.

📸 The Argos results are publicly visible under https://app.argos-ci.com/interactive-things/visualization-tool. Once we merge to `main`, we'll have baseline screenshot to compare to, now we only have orphan builds.


e2e/har-utils.ts
```diff
@@ -81,7 +81,7 @@ export const getEnv = (name: string) => {
  */
 
 export const harReplayGraphqlEndpointQueryParam =
-  process.env.E2E_HAR !== "false"
+  process.env.E2E_HAR !== undefined && process.env.E2E_HAR !== "false"
```
noahonyejese, line 84: I think you can just say: 

```javascript
 process.env.E2E_HAR && process.env.E2E_HAR !== "false"
 ```

noahonyejese (approved):
Nice stuff! 🎉

Issue #898, opened 2022-11-22
Mobile: Browse by dataset screen not rendered correctly


AnninaWalker:
**Is your enhancement related to a problem? Please describe.**
Currently the Search area is not rendered correctly on Mobile Screens. The blue bar is somehow shifted.

**Describe the solution you'd like**
Fix blue bar.

**Additional context**
![IMG_9450](https://user-images.githubusercontent.com/59613691/203323297-a13c43ef-05eb-4953-841b-888ba8599280.PNG)

AnninaWalker:
Still persistent on int, even with the new CI CD Implementation


<img width="1170" height="2532" alt="Image" src="https://github.com/user-attachments/assets/c9ecb11a-4039-459e-9657-580945e22f78" />

AnninaWalker:
@sosiology Yes, I double checked already last friday.. Looks good now!

PR #1951, opened 2025-01-08
fix: Improved readability on mobile

noahonyejese:
**This PR**
- Left alignes the searchbar, which partially resolves_ #898 
- Improves readability for the dataset preview title and button


bprusinowski (approved):
Thanks @noahonyejese 🎨 

I wonder though if the left aligning of the search panel shouldn't only happen on mobile? It looks like the old design was correct when looking at the [Figma file](https://www.figma.com/design/Hn9wvuEYLUmdJtg4KwZYUx/visualize.admin.ch---design?node-id=2235-51034&t=x7bojjQTLLFXx5W6-4), maybe we can only introduce this change on smaller devices?

I know that it's coming in the new CI/CD design, but I am not sure if we'll work on it before the next release 🤔

noahonyejese:
> Thanks @noahonyejese 🎨
> 
> I wonder though if the left aligning of the search panel shouldn't only happen on mobile? It looks like the old design was correct when looking at the [Figma file](https://www.figma.com/design/Hn9wvuEYLUmdJtg4KwZYUx/visualize.admin.ch---design?node-id=2235-51034&t=x7bojjQTLLFXx5W6-4), maybe we can only introduce this change on smaller devices?
> 
> I know that it's coming in the new CI/CD design, but I am not sure if we'll work on it before the next release 🤔

I can also scope it only for mobile for now no problem!

PR #1962, opened 2025-01-09
chore: Add knip, optimize CI, clean up

### LINKED ISSUE #1961: Add knip to CI
In order to be able to continuously clean up the application from unused code and dependencies, we should introduce a [knip](https://knip.dev) GitHub Action that would run as a part of our CI pipeline.

bprusinowski:
Closes #1961

This PR:
- adds [knip](https://knip.dev) to CI, so that we can catch dead code and unused files / dependencies as soon as they are introduced,
- cleans up many packages and dead code that was not used anymore (based on `knip` output),
- adds packages that were installed implicitly to `package.json` files,
- improves performance of basic CI checks (unit tests, linting, knip, type checking) by using [sharding](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow),
- updates some packages, including switching from single `d3` package to individual modules.

I am not 100% happy with how the knip config looks like now, as there was a need to ignore quite a few places in the application, but I think it's a good starting point that could be optimized in the future. But the main goal of catching dead code or unused dependencies should work from now on by a `knip` GitHub Action as part of every PR 👍


noahonyejese (approved):
Thanks @bprusinowski for improving some of the d3 and other imports! 🎉Overall LGTM.

PR #1966, opened 2025-01-10
chore: E2E tests maintenance

### LINKED ISSUE #1950: Review and add missing E2E tests
It's been a while since we did a proper review and updated the tests to e.g. not be based on depreciated cubes. It would be good to also include newer features in these, to make sure we catch regressions as soon as possible.

bprusinowski:
Closes #1950

This PR:
- improves performance of E2E tests by using cached LINDAS endpoints,
- cleans up unused chart config fixtures (and switches from `dev` to `test` naming convention for them),
- enables localized abbreviations test on CI, as it seems to work correctly now,
- adds two new E2E tests for chart duplication and image download.


e2e/chart-actions.spec.ts
```diff
@@ -18,3 +20,23 @@ test("it should be possible to duplicate a chart", async ({
...
+  });
+  await selectors.chart.loaded();
+  const chartMoreButton = await selectors.chart.moreButton();
+  await chartMoreButton.click();
+  const downloadPromise = page.waitForEvent("download");
+  await (await selectors.mui.popover().findByText("Export PNG")).click();
+  const download = await downloadPromise;
+  expect((await promises.stat(await download.path())).size).toBeGreaterThan(
+    // Assuring the file is not empty.
+    4000
```
noahonyejese, line 40: Is this supposed to detect empty files or files under 4000 bytes, the comment could be a bit confusing due to the number and vice versa. 👀

noahonyejese (approved):
LGTM 🎉

e2e/chart-actions.spec.ts
bprusinowski: Good point, I didn't want to do > 0, as I've noticed when developing that in case of some initial bugs with downloading the images, the were empty but still around 1KB, so that would bypass the check. I'll adjust the comment 👍

Issue #1861, opened 2024-11-13
✓ Implement: Line diagram with points


florelina:

KerstinFaye:
Looks good to me, tested on Brave, Safari & Chrome

Issue #1850, opened 2024-11-12
✓ [s6.1] Line diagram with points (BLW)


florelina:
### **Description**
Add the capability for a line chart with data points displayed at each value for more visual clarity.

### **Examples**
![Image](https://github.com/user-attachments/assets/0e9c1515-a9c2-480f-b841-b0a4395e6389)

### **Possible Approaches**
- Add a toggle in chart settings to enable or disable data points.
- ~~Allow customization of data point size and color within the color palette settings.~~

### **Defintion of Done**
- [ ] A new line chart option with data points is available on the dashboard.
- [ ] Data points are displayed at each value along the line, providing clearer data visualization.
- [ ] Users can toggle the display of data points on or off within chart settings.
- [ ] ~~Data point size and color can be customized via settings.~~ This is not a requirement from the client

AnninaWalker:
@florelina Has it been a requirement by the client that the color can be adjusted separately? In my opinion the datapoint should have always the same color as the line. So whatever color the user chooses for the line, this will automatically be applied to the data point.

florelina:
No this is not a request from the client as far as I know. @sosiology please correct me if I'm wrong (see 4th to do point)

sosiology:
@florelina correct. This is not a requirement (that is known at this time)

KerstinFaye:
LGTM, tested on Chrome, Brave and Safari. Works all well. Also Mobile view looks good, the amount of dots visible are adapting.

PR #1967, opened 2025-01-13
Dots on line chart

squiles:
Refers to [#1861](https://github.com/visualize-admin/visualization-tool/issues/1861)

Adds a control on the vertical axis that allows the user to show dots on top of a line chart. Size can be controlled through the control.


app/charts/chart-config-ui-options.ts
```diff
@@ -802,6 +808,7 @@ const chartConfigOptionsUISpec: ChartSpecs = {
         },
         options: {
           showStandardError: {},
+          showConfidenceInterval: {},
```
bprusinowski, line 811: Nice catch 👍

app/charts/line/chart-lines.tsx
```diff
@@ -37,6 +37,7 @@ const ChartLines = memo((props: ChartProps<LineConfig>) => {
         <ChartSvg>
           <AxisHeightLinear /> <AxisTime /> <AxisTimeDomain />
           <Lines />
+          <Points chartConfig={chartConfig} />
```
bprusinowski, line 40: @squiles I wonder if it wouldn't be a bit better to have a conditional here, like:

```tsx
{chartConfig.fields.y.showDots ? <Points dotsSize={chart.fields.y.showDotsSize} /> : null}
```

so that we don't need to compute `Points`-related things (like e.g. `ticksPos` in `useMemo`) when the dots aren't supposed to be shown?

Then you could also remove `undefined` from the switch in `getPointRadius`, as when it's undefined, technically we shouldn't show any dots, right?

app/charts/line/lines.tsx
```diff
@@ -122,3 +124,72 @@ const Line = memo(function Line({
...
+      ).center;
+
+      const i = bisectDate(chartData, date, 1);
+      const y = yScale(getY(chartData[i]) as number);
+
+      const segment = getSegment(chartData[i]);
+      const color = colors(segment);
+
+      return { x, y, color };
+    });
```
bprusinowski, line 169: @squiles I think this logic needs to be adjusted to also support `TemporalEntityDimension`s (see that nothing happens when you enable dots in [this cube](https://visualization-tool-git-feat-line-dots-ixt1.vercel.app/en/create/new?cube=https://agriculture.ld.admin.ch/foag/cube/MilkDairyProducts/Consumption_Price_Month&dataSource=Prod)) 👀 

I wonder if we even need the `bisectDate` thing, wouldn't it be enough to simply loop over `chartData` and draw the dots like that?

app/charts/line/chart-lines.tsx
squiles: yeah, definitely makes sense. I've changed it accordingly 👍

app/charts/line/lines.tsx
squiles: Unfortunately we do have to bisect the date because otherwise if we just loop the `chartData` then every single point along the line will have a dot on it.

app/charts/line/lines.tsx
```diff
@@ -122,3 +124,73 @@ const Line = memo(function Line({
...
+  const ticksPos = useMemo(() => {
+    return xScale.ticks(bounds.chartWidth / 90).map((tick) => {
+      const x = xScale(tick);
+      const date = xScale.invert(x);
+
+      const bisectDate = bisector(
+        (d: Observation, date: Date) => getX(d).getTime() - date.getTime()
+      ).center;
+
+      const i = bisectDate(chartData, date, 0);
```
squiles, line 163: @bprusinowski this was the guy that was tricking the first value of the data set. I changed the last argument from a 1 to 0 and now all data sets that have a dot on the first index of the data seem to work properly. I should have caught this but unfortunately it seems that the 3 or 4 datasets that I used to  test the feature weren't like this.

app/charts/line/lines.tsx
bprusinowski: Hmmm, I thought that this supposed to be the case 🤔 Otherwise we are simply highlighting the X axis ticks on a chart, right? `chartData` should only contain actual data points, so e.g. if we have a monthly data, I'd expect to see 12 dots between the years, basically to see all the points you can hover over, but without hovering (also see the [comment here](https://www.figma.com/design/Hn9wvuEYLUmdJtg4KwZYUx/visualize.admin.ch---design?node-id=6360-116488&m=dev)) 😄

Let me know if I missed something @squiles 🥹

squiles: I thought the goal was to do just that - to highlight the x axis ticks on the line itself with the dots, but maybe i misunderstood it. If you take for example the dataset named: `Market figures milk and dairy products - Monthly consumer prices` you'll see that we have data for not only the years that we show on the x axis but also the months that are in between the years so if we show all the dots we have access to it looks like this: 

Screenshot 2025-01-15 at 10 54 15

Instead of this: 

Screenshot 2025-01-15 at 10 55 21

KerstinFaye: @bprusinowski @squiles
Seeing the example above I guess it would make sense to only highlight the x axis ticks. Otherwise we would simply redraw the complete line with dots. Not sure if it would meet the users need. I checked again the initial example from the ticket and I think this would be aligned with their needs: https://github.com/visualize-admin/visualization-tool/issues/1850

KerstinFaye: It would make sense to ask again if this fulfills BLWs need as soon as they test the implemented version. @sosiology

bprusinowski (approved):
Thanks @squiles, LGTM!

KerstinFaye:
@squiles Looks good. Thanks! And I like the default size option that you implemented.

PR #1989, opened 2025-01-21
Bar chart scroll issue

squiles:
Very hacky way of solving the bar chart scrolling issue where the tooltip jumped around. Now it's always rendered in the right place. Still need to see if we can improve the current implementation to make it simpler.


squiles:
Outdated. Solved by #1994

PR #1994, opened 2025-01-23
fix: Bar chart tooltip scroll

### LINKED ISSUE #1978: Horizontal Barchart: Improve Tooltip Behaviour
Currently the tooltip jumps around outside of the chart "view port" when scrolling, specifically on smaller screens or reduced browser windows. 

https://github.com/user-attachments/assets/fd03aa50-ca28-461d-87b5-a6279acfd122

Possible Resolution: Disable tooltips when the user is scrolling.

bprusinowski:
Closes #1978

This PR changes the behavior of `usePosition` hook to re-calculate tooltip's position on `y` coordinate change. Previously we were returning early when bounding client rect was already set, which led to using outdated coordinates when scrolling.

## How to test

1. Go to [this link](https://visualization-tool-git-fix-tooltip-scroll-ixt1.vercel.app/en/create/new?cube=https://agriculture.ld.admin.ch/foag/cube/MilkDairyProducts/Consumption_Price_Month&dataSource=Prod&flag__enable-experimental-features=true).
2. Switch to a bar chart.
3. ✅ Hover over the chart, scroll and see that the tooltip correctly positions itself after the scrolling event.

## How to reproduce
1. Go to [this link](https://test.visualize.admin.ch/en/create/new?cube=https://agriculture.ld.admin.ch/foag/cube/MilkDairyProducts/Consumption_Price_Month&dataSource=Prod&flag__enable-experimental-features=true).
2. Switch to a bar chart.
3. ❌ Hover over the chart, scroll and see that the tooltip goes outside of chart's area after scrolling event.

---

- [x] Add a CHANGELOG entry
- [x] Add an end-to-end test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ You can also check the
 
 # Unreleased
 
-Nothing yet.
+- Fixes
+  - Bar chart tooltip doesn't go off the screen anymore during scroll
 
 # [5.2.0] - 2025-01-22
 
--- a/app/charts/shared/interaction/tooltip-box.tsx
+++ b/app/charts/shared/interaction/tooltip-box.tsx
@@ -1,6 +1,6 @@
 import { Box } from "@mui/material";
 import throttle from "lodash/throttle";
-import { ReactNode, useCallback, useEffect, useMemo, useState } from "react";
+import { ReactNode, useEffect, useMemo, useRef, useState } from "react";
 import ReactDOM from "react-dom";
 
 import { Margins, useSize } from "@/charts/shared/use-size";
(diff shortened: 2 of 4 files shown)


bprusinowski:
Hey @squiles, I took a look at the bar chart tooltip scroll issue and I think I managed to fix it a bit closer to the actual cause of the bug. One thing that's different compared to https://github.com/visualize-admin/visualization-tool/pull/1989 is that we don't update tooltip's position on scroll, but with this PR the tooltip won't go outside of viewport, and we are not causing the whole chart to re-render on scroll.

Let me know what you think!

noahonyejese (approved):
LGTM, thanks @bprusinowski ✅

squiles (approved):
Awesome! Much more elegant than what I managed to come up with. I think I was going in the right direction in my first attempt after all 😅

Issue #2003, opened 2025-01-24
✓ Horizontal barchart: Interactive Filter does not work


sosiology:
**Description**
The interactive filters on the vertical access do not work for the horizontal bar charts and results in an unexpected error. 

![Image](https://github.com/user-attachments/assets/5331eff0-7ee1-4bff-8350-1f645ec77ce5)

**To Reproduce**
1. Go to this [link](https://int.visualize.admin.ch/en/v/xCe7LWn7VFlg?dataSource=Prod)
2. Copy and edit the visualization
3. Open the vertical axis panel and set the filter to interactive with the toggle. 
4. See Unexpected error 

**Expected Results**
Interacting with the toggle should not result in an error. If the error cannot be avoided due to the year being on the y-axis, then it should either not be shown, or the user should not be able to toggle it and a message should be shown to explain why. (similar to how we handle this for animations) 

![Image](https://github.com/user-attachments/assets/688f9d8e-dfca-49b2-ab6d-bace6604c61f)

**Screenshots**
https://github.com/user-attachments/assets/b5f6dceb-86b6-4227-b0c6-8b08a6a286cb

sosiology:
Tested on TEST v5.2.1 ([3a41c67](https://github.com/visualize-admin/visualization-tool/commit/3a41c67b8379ca9e8c5bb4d377e22340d6700948)) in Chrome with feature flag. Works as expected: no error is thrown when the vertical axis filter is set to "interactive"

PR #2025, opened 2025-01-27
Fix interactive filter bar chart

squiles:
Closes [#2003](https://github.com/visualize-admin/visualization-tool/issues/2003)

This PR fixes a bug where the chart would crash when the interactive filter was selected on the y axis of a bar chart. The reason was that the computations necessary to make the brushes were not expecting a bar chart that has the axis flipped (when comparing to column charts). 

## How to test

1. Go to a dataset
2. Select the Y axis
3. Toggle the interactive switch
4. Notice the chart works as intended

- [x] Add a CHANGELOG entry

### DIFF
--- a/app/charts/shared/brush/index.tsx
+++ b/app/charts/shared/brush/index.tsx
@@ -119,27 +119,19 @@ export const BrushTime = () => {
 
   // Brush dimensions
   const { width, margins, chartHeight } = bounds;
+  const scaleTimeRange =
+    chartType === "bar" ? yScaleTimeRange : xScaleTimeRange;
   const brushLabelsWidth =
-    chartType === "bar"
-      ? getTextWidth(formatDate(yScaleTimeRange.domain()[0]), {
-          fontSize: labelFontSize,
-        }) +
-        getTextWidth(" - ", { fontSize: labelFontSize }) +
-        getTextWidth(formatDate(yScaleTimeRange.domain()[1]), {
-          fontSize: labelFontSize,
-        }) +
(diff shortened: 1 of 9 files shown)


noahonyejese (approved):
LGTM thanks @squiles 👍

app/charts/shared/brush/index.tsx
```diff
@@ -79,33 +83,63 @@ export const BrushTime = () => {
...
+        }) +
+        getTextWidth(" - ", { fontSize: labelFontSize }) +
+        getTextWidth(formatDate(xScaleTimeRange.domain()[1]), {
+          fontSize: labelFontSize,
+        }) +
+        HANDLE_HEIGHT;
   const brushWidth = width - brushLabelsWidth - margins.right;
-  const brushWidthScale = xScaleTimeRange.copy();
+  const brushWidthScale =
+    chartType === "bar" ? yScaleTimeRange.copy() : xScaleTimeRange.copy();
```
bprusinowski, line 142: Nit: @squiles I wonder if we can't extract something like

```typescript
const scaleTimeRange = chartType === "bar" ? yScaleTimeRnage : xScaleTimeRange;
```

and avoid having ternaries and duplicating the logic – I think it's all the same except for scale variable 🤔

bprusinowski (approved):
Thanks @squiles, code LGTM with one suggestion, but it looks like the interactive slider overlaps with X axis title? 🤔 

Screenshot 2025-01-28 at 17 21 42

PR #2030, opened 2025-01-28
feat: Add new GitLab mirroring GitHub Action

### LINKED ISSUE #2076: GitLab mirroring broken
**Describe the bug**
Mirroring of this repository to GitLab is broken since late January. The problem looks like it has to do with how GitLab handles references to HEAD when we attempts to push to the repo. This is really only an alias, but it seems that GitLab doesn't correctly interpret this (anymore?). This behaviour can be reproduced manually with the following syntax:

`git push --tags --force --prune mirror refs/remotes/origin/*:refs/heads/*`

Most available actions use some variant of this logic, but unfortunately, the [refspec](https://git-scm.com/book/en/v2/Git-Internals-The-Refspec) here seems too lenient.

Image

**To Reproduce**

Issue happens on every trigger of the GitHub Action and can be reproduced by initiating the push manually.

**Expected behavior** A clear and concise description of what you expected to
happen.

Mirroring this repository to GitLab via Action runs successfully.

bprusinowski:
Closes #2076 

This PR implements a different GitHub Action for mirroring the repo to GitLab (https://github.com/wearerequired/git-mirror-action)

## How to test

Pushing code to this PR will initiate the Action.

See #2076 for how to reproduce manually.

- [ ] Add a CHANGELOG entry


Rdataflow:
@adintegra have you ever tried Step Order
1. downgrade git
2. checkout 
3. mirror

adintegra:
Thanks @Rdataflow! Yes, tried with various versions of git – this doesn't seem to have been the issue.

The problem looks like it has to do with how GitLab handles references to `HEAD` when we attempts to push to the repo. This is really only an alias, but it seems that GitLab doesn't correctly interpret this. This behaviour can be reproduced manually with the following syntax:

`git push --tags --force --dry-run --prune mirror refs/remotes/origin/*:refs/heads/*`

Most available actions use some variant of this logic, but unfortunately, the [refspec](https://git-scm.com/book/en/v2/Git-Internals-The-Refspec) here is too lenient. The action which works successfully here takes a slightly different approach.

Screenshot 2025-02-04 at 15 02 01

@bprusinowski I can't add you as a reviewer here, but perhaps you could take a look and let me know if you see any issues with this approach.

.github/workflows/gitlab-mirror-test.yml
```diff
@@ -2,6 +2,10 @@ name: GitLab Mirror Test
 
 on: [push, delete]
 
+# Ensures that only one mirror task will run at a time.
```
noahonyejese, line 5: super-nit: I guess the word `concurrency` already implies that it will run one after another - you could potentially remove this comment 🤷🏽‍♂️👀

noahonyejese (approved):
@adintegra LGTM, I think this should work just fine. 🎉 If you want to be extra safe you could use `matrix strategy` here, something like this

 ```yml
   strategy:
      matrix:
        destination:
          - name: GitLab Mirror Test
            repo: "git@gitlab.ldbar.ch:interactivethings/visualize-mirroring-test.git"
          - name: GitLab Mirror
           repo: "git@gitlab.ldbar.ch:interactivethings/visualization-tool.git"
```

and finally add

```yml
destination-repo: ${{ matrix.destination.repo }}
```

.github/workflows/gitlab-mirror.yml
```diff
@@ -2,15 +2,26 @@ name: GitLab Mirror
...
+        uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - uses: pixta-dev/repository-mirroring-action@v1
+      - name: Logging
+        run: |
+          git remote show origin
+          git for-each-ref
+      - name: Mirror to GitLab
+        uses: wearerequired/git-mirror-action@v1
```
bprusinowski, line 22: @adintegra so the fix was to use a different action? It looks like here we are not doing anything different than previously, or was the runner configuration changed?

bprusinowski:
Thanks @adintegra, I'm very glad this works again 🎉 Could we already remove `gitlab-mirror-test` action and GitLab repo / test secrets added to GitHub repo settings as "part" of this PR?

adintegra:
> Could we already remove `gitlab-mirror-test` action and GitLab repo / test secrets added to GitHub repo settings as "part" of this PR?

@bprusinowski thanks for looking at this! Yes, for sure. I had just left it in to be safe but it can be removed now 👍🏼

.github/workflows/gitlab-mirror.yml
adintegra: This action uses slightly [different logic](https://github.com/wearerequired/git-mirror-action/blob/c44c68aebe42de49a80a8dfd557416b6a0cafdad/git-mirror.sh#L20) to determine which refs will be pushed. I think the key is that it doesn't attempt to push `origin/HEAD`.

bprusinowski: Thanks for the context @adintegra 🙇‍♂️

adintegra:
Thanks for reviewing @noahonyejese! 🙇‍♂️

PR #2035, opened 2025-01-29
fix: Disable multiline Input

### LINKED ISSUE #2010: Search bar height increases when hitting enter
**Description**
Hitting enter in the empty search bar above a table chart, results in the height of the field increasing. 

Image

Screen Recording: https://github.com/user-attachments/assets/2dc05bd8-170c-4bff-bbdc-5a93d21e52e0

**Steps to Reproduce**
1. Go To https://int.visualize.admin.ch/v/RQIGaqHronDZ?dataSource=Prod
2. Click the search bar so that the cursor appears there. Hit enter. 
3. See that the height of the search bar increases. 

**Expected Behaviour**
The height of the search bar doesn't change regardless of user input. 

Tested on INT v5.2.0 in Chrome

bprusinowski:
Closes #2010

This PR removes the `multiline` property from Input so that it does not extend when hitting Enter. This change was introduced while working on the dashboard text blocks, but as we switched to using a third-party library for Markdown inputs, this should have been removed.

## How to test

1. Go to [this link](https://visualization-tool-git-fix-multiline-input-ixt1.vercel.app/en/v/Oa9u2jQOjAtr?dataSource=Prod).
2. Enter some text in the search bar above the table and hit Enter.
3. ✅ See that the input field doesn't extend.

## How to reproduce
See the instructions in #2010.

---

- [x] Add a CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ You can also check the
   - PNG image download now correctly retains interactive legend's colors
   - Charts based on a new cube are not hidden anymore in the layouting step
   - Scroll bars are not shown in the chart area when not needed
+  - Input fields do not extend anymore when hitting Enter
 
 # [5.2.0] - 2025-01-22
 
--- a/app/components/form.tsx
+++ b/app/components/form.tsx
@@ -642,7 +642,6 @@ export const Input = ({
         id={name}
         size="small"
         color="secondary"
-        multiline
         name={name}
         value={value}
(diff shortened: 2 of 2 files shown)


bprusinowski:
@noahonyejese @squiles as the fix was trivial, as it essentially reverts one change I did when working on dashboard text blocks and we want to deploy most recent version to INT, I will merge this PR 👍

PR #2043, opened 2025-01-30
Created by visualize link points to Chart

### LINKED ISSUE #2031: Link "created with visualize.admin.ch" should point to the published chart
Currently the link at the bottom of the charts and dashboards "created with visalize.admin.ch" points to the Visualize landing page. In the weekly call on 28.01.2025 we received the request that this link should point to the published chart so users can copy and edit the chart if they would like. 

Example: https://visualize.admin.ch/de/embed/iAmx1iuakRwr?dataSource=Prod
The link created with visualize.admin.ch should lead the users to this page https://visualize.admin.ch/de/v/iAmx1iuakRwr?dataSource=Prod (the chart URL) and not visualize.admin.ch

**Use cases and impact**
Users who see charts embedded in iframes have a possibility to copy and edit the chart without needing to start from scratch and going through the entire chart creation flow. 

**Additional context** 
This was implemented as described in the past (points to the published chart and not the landing page), at the time however there were fewer users embedding charts from Visualize, so it seemed redundant to have a link on the published chart page, that leads to the same page.

noahonyejese:
Closes #2031 

**What's new?**
This PR Now navigates users to the corresponding Chart when clicked on the "Created by visualize.admin.ch" section in the footer, like this users can benefit from quick chart creation & potentially starting out with a more established Chart.

## How to test

1. Go to this Link
2. Create a new Visualisation 
3. Publish Visualisation 
4. Memorize last part of the URL (e.g `Xx3-xKHq9K2G`) -> this is necessary because we only support production charts
5. Click the "Created by visualize.admin.ch"
6. See how it navigates to the same Chart (See url) ✅

---

@bprusinowski please let me know if it makes sense here to make the URL dynamic or if I should keep it only for visualize.admin.ch, Vercel also has a URL called VERCEL_URL we could be using this for cases like these. Let me know what you think. 👀

### DIFF
--- a/app/components/chart-shared.tsx
+++ b/app/components/chart-shared.tsx
@@ -402,7 +402,7 @@ const DownloadPNGImageMenuActionItem = ({
   chartKey: string;
   components: Component[];
 } & Omit<UseScreenshotProps, "type" | "modifyNode" | "pngMetadata">) => {
-  const modifyNode = useModifyNode(configKey);
+  const modifyNode = useModifyNode();
   const metadata = usePNGMetadata({
     configKey,
     chartKey,
@@ -428,7 +428,7 @@ const DownloadPNGImageMenuActionItem = ({
   );
 };
 
-const useModifyNode = (configKey?: string) => {
+const useModifyNode = () => {
   const theme = useTheme();
   const chartWithFiltersClasses = useChartWithFiltersClasses();
 
@@ -455,14 +455,13 @@ const useModifyNode = (configKey?: string) => {
         `.${CHART_FOOTNOTES_CLASS_NAME}`
       );
 
-      if (footnotes && configKey) {
+      if (footnotes) {
         const container = document.createElement("div");
         footnotes.appendChild(container);
         const root = createRoot(container);
         root.render(
           <ThemeProvider theme={theme}>
             <VisualizeLink
-              configKey={configKey}
               createdWith={t({ id: "metadata.link.created.with" })}
             />
           </ThemeProvider>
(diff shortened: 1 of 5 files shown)


sosiology:
Thanks for the enhancement @noahonyejese i believe this will need to be on TEST before it can be tested. I tested in the PR with this chart: https://visualization-tool-git-fix-visualize-link-to-public-charts-ixt1.vercel.app/en/v/cZhzddlopnkb?dataSource=Prod the target link i see is https://visualize.admin.ch/en/v/cZhzddlopnkb (which looks correct - as this suggests it points to a chart and not the landing page) but when I click the link I get a 404 - this page could not be found.

app/components/chart-shared.tsx
```diff
@@ -428,7 +428,7 @@ const DownloadPNGImageMenuActionItem = ({
   );
 };
 
-const useModifyNode = () => {
+const useModifyNode = (configKey?: string) => {
```
bprusinowski, line 431: Super-nit: @noahonyejese as this is used in a static screenshot, I'd say we don't need to pass it down, as this wouldn't change the end result, link is not clickable in PNG 👍 But it's also fine like this, if you think it makes sense feel free to leave as is :)

bprusinowski (approved):
Thanks @noahonyejese, LGTM!

I think it would be good to consider two comments related to `state.key`, but other than that really nice :)

I wouldn't introduce dynamic URLs, it would only make it work for testing environments, which aren't supposed to hold shareable charts anyway.

We'd also potentially introduce another dependency to Vercel, which I don't think we really need 👍

app/components/chart-shared.tsx
noahonyejese: true 😅

KerstinFaye:
Hi @noahonyejese I tried to test as described above. But after clicking on create with Visualize, it leads me to a 404 page.

sosiology:
> Hi @noahonyejese I tried to test as described above. But after clicking on create with Visualize, it leads me to a 404 page.

@KerstinFaye i assume it has to do with it being a PR -> also see comment here: https://github.com/visualize-admin/visualization-tool/pull/2043#issuecomment-2624152980

PR #2050, opened 2025-01-30
fix: NaNs in map tooltip

### LINKED ISSUE #2034: NaN values in map tooltip
When we create a map with a symbol layer with a constant size, we can see `NaN` values in the tooltip – I think here we shouldn't show such label at all.

To reproduce:
1. Go to [this link](https://test.visualize.admin.ch/en/v/nQLq0TL7vkkt?dataSource=Prod).
2. Hover over the dots and see the `NaN` values in the tooltip.

bprusinowski:
Closes #2034

This PR prevents displaying of NaN values in map tooltip.

## How to test

1. Go to this link.
2. Hover over the circles.
3. ✅ See that there are no `NaN` values in the tooltip.

## How to reproduce
See #2034

---

- [ ] Add a CHANGELOG entry
- [ ] Add an end-to-end test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ You can also check the
     published mode
   - Map legend items do not overlap anymore in PNG image downloads
   - PNG image download now correctly retains Y axis label colors in combo charts
+  - NaN values are not displayed in map tooltip anymore
 - Style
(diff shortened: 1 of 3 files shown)


noahonyejese (approved):
LGTM

PR #2051, opened 2025-01-31
feat: Added color change when changing segment dimensions

### LINKED ISSUE #2006: Segmentation Dimension can no longer be changed
**Description**
On INT the color of the pie chart sections cannot be changed.

**How to Reproduce**
1. Go to int.visualize.com
2. Select data set "Bathing water quality"
3. Start visualization and select Pie Chart. 
4. See that all sections of the pie chart appear in blue per default 
5. Open the segmentation panel and see that in the "Filter" Section 819 filters are selected but they are not listed at the bottom of the side bar. 

**Expected** 
The user should be able to edit the colors of the pie chart. Per default not all sections of the pie should be the same color. 

Example Charts: 
INT: https://int.visualize.admin.ch/en/v/desJRM5bDx7c?dataSource=Prod
DEV: https://visualize.admin.ch/en/v/iAmx1iuakRwr?dataSource=Prod

Image

### LINKED ISSUE #2016: color segmentation broken on TEST
**Describe the bug** 
most probably this is already known. reporting just to ensure we keep track of this case...
color segmentation is broken - every segment appears in the same color.

**To Reproduce** Steps to reproduce the behavior:

1. Go to https://test.visualize.admin.ch/create/new?cube=https://energy.ld.admin.ch/sfoe/bfe_ogd18_gebaeudeprogramm_energiewirkung/7&dataSource=Prod
2. Click on *Segmentation*
3. Choose *Measure*
4. See every segment has the same color (blue) - ignoring any choice in the color palette

**Expected behavior**
- Apply default color palette
- Apply chosen color palette

**Additional context**
while trying to reproduce another time we learn
- this is not strictly reproducible, but only happens sometimes - not sure why - (maybe it was some legacy code parts still cached locally?)

noahonyejese:
Closes #2006 and Closes #2016

**This PR:**
This PR ensure that when switching segmentation dimensions the colors also adjust not only for the initial dimension selection.

## How to test

1. Go to this [Link](https://visualization-tool-git-fix-segment-no-color-ixt1.vercel.app/)
2. Create a new Chart with `Market figures milk and dairy products - Monthly consumer prices`
3. Select a Pie Chart f.e 
4. Click on `Segmentations` 
5. Change dimension to something like `Date`
6. See how all the colors are being applied and are changable ✅

## How to Reproduce bug

1. Go to this [Link](https://int.visualize.admin.ch/en/create/lOzHVC9xwBRD?dataSource=Prod)
2. Create a new Chart with `Market figures milk and dairy products - Monthly consumer prices`
3. Select a Pie Chart f.e 
4. Click on `Segmentations` 
5. Change dimension to something like `Date`
6. See how everything remains one color ❌
---

_ps: @bprusinowski I think this will require some refactoring as this is more or less duplicated from
 `color-palettes.tsx` however the other PR already has some util files for colors so I don't want to recreate too much again here. But let me know what you think_ 

---

- [x] Add a CHANGELOG entry

cc: @sosiology , @bprusinowski

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ You can also check the
   - Weekly-based temporal dimensions now work correctly when used in dashboard
     filters
   - Fixed color changes when changing segment dimension
+  - Fixed Filter selection made it so that color are showing again correctly for
+    selected filters
   - Chart area doesn't overflow anymore when changing chart types in editing
     mode in Chrome
   - Tall dashboard layouts now correctly align chart elements between columns in
@@ -22,8 +24,10 @@ You can also check the
   - Map legend items do not overlap anymore in PNG image downloads
   - PNG image download now correctly retains Y axis label colors in combo charts
   - NaN values are not displayed in map tooltip anymore
+  - Fixed color shuffling & resetting to work again with custom color palettes
 - Style
   - Improved vertical spacing between map legend items
+  - Fixed the spacing between navigation and header in the /profile view
 
 # [5.2.1] - 2025-01-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,11 +15,26 @@ You can also check the
   - Weekly-based temporal dimensions now work correctly when used in dashboard
     filters
   - Fixed color changes when changing segment dimension
+  - Chart area doesn't overflow anymore when changing chart types in editing
+    mode in Chrome
+  - Tall dashboard layouts now correctly align chart elements between columns in
+    published mode
+  - Map legend items do not overlap anymore in PNG image downloads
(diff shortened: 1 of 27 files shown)


bprusinowski (approved):
Thanks @noahonyejese, looks good! You're right that it looks quite similar to the changes from another PR, maybe we can merge the other one first, and then try to consolidate the logic between them? 🤔

noahonyejese:
@bprusinowski yeah, probably however I haven't yet fully completed creating all tests 😬

bprusinowski:
Ah yes – then all good, maybe you we could reverse then and merge this one first and try to to do it the other way around? But the other PR is already big, maybe this can also be done after we merge everything 😅

PR #2071, opened 2025-02-04
chore: Increase the number of charts pre-loaded to Varnish cache

bprusinowski:
This PR increases the number of charts pre-loaded to the Varnish cache from 25 to 250, as the caching capabilities improved in the meantime.

---

- [x] Add a CHANGELOG entry

cc @Rdataflow


noahonyejese (approved):
Nice, LGTM 👍

PR #2085, opened 2025-02-05
fix: Most recent filter + interactive time range

bprusinowski:
This PR fixes initialization of published charts that use most recent date filter together with interactive filter, so that the full range is displayed, instead of being "stuck" in the start date.

## How to test

1. Go to this link.
2. Open X axis field.
3. Enable most recent date filter.
4. Make the filter interactive.
5. Publish the chart.
6. ✅ See that's initialized correctly.

## How to reproduce
Follow the above steps on TEST and see that chart is broken in published mode (only a single year is selected in interactive filter).

---

- [x] Add a CHANGELOG entry
- [x] Add an end-to-end test

cc @sosiology

### DIFF
--- a/app/charts/shared/brush/index.tsx
+++ b/app/charts/shared/brush/index.tsx
@@ -74,8 +74,6 @@ export const BrushTime = () => {
       setSelectionExtent(0);
     }
   };
-
-  const { from, to } = timeRange;
   const {
     brushOverlayColor,
     brushSelectionColor,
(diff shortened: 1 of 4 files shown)


app/charts/shared/brush/index.tsx
```diff
@@ -121,6 +119,15 @@ export const BrushTime = () => {
   const { width, margins, chartHeight } = bounds;
   const scaleTimeRange =
     chartType === "bar" ? yScaleTimeRange : xScaleTimeRange;
+
+  let { from, to } = timeRange;
+
+  // FIXME: Should be fixed in useSyncInteractiveFilters where we try to parse
+  // the date that can be a string (VISUALIZE_MOST_RECENT_VALUE).
```
noahonyejese, line 126: Is this something worth opening an issue for? Or how urgent is this? Any specific timeline we want to tackle this? - sorry for all the questions 😅

noahonyejese (approved):
LGTM ✅ - Should be able to merge 🙇🏽‍♂️

app/charts/shared/brush/index.tsx
bprusinowski: Sorry for such late notification @noahonyejese (automatic review request as code owner), I didn't expect a review at such time, but wanted to open a PR so we have it documented (but thank you for the review 🔝). This is a hot-fix just before the PROD release tomorrow in the morning, but I'll open an issue to fix it properly 👍

bprusinowski: https://github.com/visualize-admin/visualization-tool/issues/2086

PR #2089, opened 2025-02-07
refactor: added visible param to legends

### LINKED ISSUE #1874: Implement: Enhanced Legend Titles & Label Options for Dual Y-Axis Charts
First Design Draft:

https://www.figma.com/design/LrxKyljpgvadf7H0jlsqVb/visualize.admin.ch---design_NEW?node-id=8871-42166&t=7kj5a3NtqtcurLRS-4

noahonyejese:
Closes #1874 

**What changes?**
This PR allows users to show or hide the legend titles as needed.

## How to test

1. Go to this [Link](https://visualization-tool-git-fix-show-legend-toggle-ixt1.vercel.app/)
2. Create a Chart using (e.g Einmalvergütung für Photovoltaikanlagen)
3. Add Segmentation (select Kanton)
4. Scroll down, see there is now a new toggle switch called "Show legend titles", which is on by default ✅
5. Toggle the switch, see how the legends are no longer visible ✅

---

_@bprusinowski I still need to add migrations however I wanted you to have a quick look in case, I'm not going the right direction here. Also maybe it would make sense to do some renaming. Additionally this could be even factored out of `interactiveFiltersConfig`, however I also didn't want to blast the scope._

---

- [x] Add a CHANGELOG entry
- [x] Added e2e Tests


app/charts/bar/chart-bar.tsx
```diff
@@ -77,6 +77,7 @@ const ChartBars = memo((props: ChartProps<BarConfig>) => {
               interactive={
                 fields.segment && interactiveFiltersConfig?.legend.active
               }
+              visible={interactiveFiltersConfig?.legend.visible}
```
bprusinowski, line 80: `showTitle`? The legend is always visible in case segmentation is present :)

app/charts/shared/legend-color.tsx
```diff
@@ -182,6 +183,7 @@ export const LegendColor = memo(function LegendColor(props: LegendColorProps) {
     symbol,
     getLegendItemDimension,
     interactive,
+    visible = true,
```
bprusinowski, line 186: @noahonyejese I think I wouldn't use a default `true` here – old charts shouldn't be changed to show legend titles IMO, as in case they are embedded somewhere, that would result in layout shifts and possibly scrollable or cut iframes 🥹

app/charts/shared/legend-color.tsx
```diff
@@ -309,7 +315,7 @@ const LegendColorContent = (props: LegendColorContentProps) => {
     }
   });
 
-  return (
+  return visible ? (
```
bprusinowski, line 318: @noahonyejese I think this variable should only show / hide legend title, not the whole legend – did I get this wrong? 🤔

app/configurator/components/filters.tsx
```diff
@@ -354,44 +355,86 @@ const MultiFilterContent = ({
...
-                <MaybeTooltip
-                  tooltipProps={{ enterDelay: 600 }}
-                  title={
-                    <TooltipTitle
-                      text={
-                        <Trans id="controls.filters.interactive.tooltip">
-                          Allow users to change filters
+            <>
+              <Flex
+                sx={{ height: "33px", alignItems: "center", width: "100%" }}
```
bprusinowski, line 375: @noahonyejese I am not sure about hard-coding height like this, what was the problem without it? 👀

app/configurator/components/filters.tsx
```diff
@@ -354,44 +355,86 @@ const MultiFilterContent = ({
...
+                            <Trans id="controls.filters.show-legend.tooltip">
+                              Allow users to change Legend visibility
+                            </Trans>
+                          }
+                        />
                       }
-                    />
+                    >
+                      <div>
+                        <Trans id="controls.filters.show-legend.toggle">
```
bprusinowski, line 422: Nit: Translation string should probably be named along the lines of `show-legend-title` 🤔

package.json
```diff
@@ -36,6 +36,7 @@
     "postversion": "git push --follow-tags",
     "release:npm": "yarn build:npm && yarn publish app",
     "e2e:dev": "E2E_BASE_URL=http://localhost:3000 yarn playwright test",
+    "e2e:ui": "E2E_BASE_URL=http://localhost:3000 npx playwright test --ui",
```
bprusinowski, line 39: 👍 any reason to use `npx` instead or `yarn` as above?

noahonyejese:
@bprusinowski kinda having some issue's with migrations currently, I feel like it's something simple you would see immediately. Let me know if you find something that could help me out and speed up the process a little bit. 🙇🏽‍♂️

e2e/filters.spec.ts
```diff
@@ -85,4 +84,38 @@ describe("Filters", () => {
...
+  test("Show legend titles, should toggle legend titles visibility", async ({
+    page,
+    actions,
+    selectors,
+  }) => {
+    test.slow();
+
+    await page.goto(
+      "/en/create/new?cube=https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/5&dataSource=Prod"
+    );
```
bprusinowski, line 96: @noahonyejese thanks for adding the tests 💯 I wonder if the test was actually slow or was this copied for the one above? I think I'd try with removing it, I had an impression that tests are working quite okay now without marking them as slow.

Also, the Photovoltaikanlagen cube's version can be bumped up, it has version 10 now 👍

e2e/filters.spec.ts
```diff
@@ -85,4 +84,38 @@ describe("Filters", () => {
...
+
+    await actions.editor.selectActiveField("Segmentation");
+
+    await selectors.edition.drawerLoaded();
+
+    await (await selectors.panels.drawer().within().findByText("None")).click();
+
+    await actions.mui.selectOption("Kanton");
+
+    const legend = page.locator('[data-testId="legend-container"]');
```
bprusinowski, line 107: @noahonyejese same thing as previously, I think it should be only legend title that is shown / hidden, not the whole container 👍

CHANGELOG.md
```diff
@@ -13,6 +13,21 @@ You can also check the
 
 - Features
   - Added option for hiding legend titles using a toggle switch
+Nothing yet.
```
bprusinowski, line 16: 👁️

app/utils/chart-config/versioning.ts
```diff
@@ -1454,6 +1454,44 @@ export const chartConfigMigrations: Migration[] = [
...
+      };
+
+      if (newConfig?.interactiveFiltersConfig?.legend) {
+        newConfig.interactiveFiltersConfig.legend = {
+          ...newConfig.interactiveFiltersConfig.legend,
+          visible: true,
+        };
+      }
+
+      return newConfig;
```
bprusinowski, line 1483: @noahonyejese I think we shouldn't make the title visible by default for old chart, also see one of my previous comments 👍 

Btw. the `showTitle` parameter in chart config could be made optional with `t.partial`, so in fact we wouldn't need to have migrations at all, as old charts would still have it set to undefined, and for new we'd initialize the types correctly. But it's also fine to have it required, but then we should set it to false in migrations 👍

app/utils/chart-config/versioning.ts
```diff
@@ -2125,6 +2163,57 @@ export const configuratorStateMigrations: Migration[] = [
...
+    },
+  },
+  {
+    description: `ALL {
+      interactiveFiltersConfig {
+        legend {
+          + visible
+        }
+      }
+    }`,
```
bprusinowski, line 2176: @noahonyejese I am not sure why we have two migrations added here? When the feature has not landed on PROD yet, I think we should only have one migration to not complicate things 👍

bprusinowski (commented):
Hey @noahonyejese, I think it's a good start! I've checked the code and left some comments, the main things would be:
- we should only show and hide legend title, not the whole legend,
- migrations could be avoided altogether if `showTitle` argument was made partial,
- the new argument should be initialized correctly when we first create a chart config.

Let me know if these things are clear, otherwise let's have a sync on this 😄

noahonyejese:
Ahhhh, okay I see 👍🏽

app/configurator/components/filters.tsx
noahonyejese: Removed it

package.json
noahonyejese: Ah no sorry, It's just my go to method 😅

e2e/filters.spec.ts
noahonyejese: @bprusinowski Yep still from the copying how do I bump that version?

noahonyejese:
@bprusinowski `showTitle` is meant to be true by default right? - Also I'm not sure I added a new action `CHART_SHOW_LEGEND_TITLE_CHANGED` maybe there is already some similar existing action. lmk 👍

e2e/filters.spec.ts
bprusinowski: Just by changing 5 to 10 in the URL, just before the `&` :)

noahonyejese: Ah sure thanks

app/locales/en/messages.po
```diff
@@ -642,8 +642,8 @@ msgid "controls.filters.show-legend.toggle"
 msgstr "Show legend titles"
 
 #: app/configurator/components/filters.tsx
-msgid "controls.filters.show-legend.tooltip"
-msgstr "Allow users to change Legend visibility"
+msgid "controls.filters.show-legend-toggle"
+msgstr "Allow users to change legend visibility"
```
bprusinowski, line 717: @noahonyejese I think still this string should mention legend **title**?

Screenshot 2025-02-11 at 11 35 45

app/charts/shared/legend-color.tsx
```diff
@@ -354,7 +353,8 @@ const LegendColorContent = (props: LegendColorContentProps) => {
                   return (
                     <LegendItem
                       key={`${d}_${i}`}
-                      item={label}
+                      data-testId={`legend-title-${showTitle ? "visible" : "hidden"}`}
+                      item={showTitle ? label : ""}
```
bprusinowski, line 357: @noahonyejese I think we might still have a bit of a misunderstanding about how the feature should behave. Did you have a chance to check the [design file](https://www.figma.com/design/LrxKyljpgvadf7H0jlsqVb/visualize.admin.ch---design_NEW?node-id=8889-44406&m=dev) before implementing? 😄

What we’re aiming for here is the ability to conditionally enable the legend title. This title doesn’t exist in most legend types yet, but the toggle should control showing or hiding  e.g. the `Period` title as described in the design, which represents the dimension label linked to the `Details` panel.

Screenshot 2025-02-11 at 11 40 28

It’s important to keep the legend items visible as they are and not hide their labels. 👍

app/config-types.ts
```diff
@@ -421,7 +430,10 @@ const AreaConfig = t.intersection([
 export type AreaFields = t.TypeOf<typeof AreaFields>;
 export type AreaConfig = t.TypeOf<typeof AreaConfig>;
 
-const ScatterPlotSegmentField = GenericField;
+const ScatterPlotSegmentField = t.intersection([
+  GenericField,
+  t.partial({ showTitle: t.boolean }),
```
bprusinowski, line 436: 👍

app/configurator/configurator-state/reducer.tsx
```diff
@@ -662,6 +663,20 @@ const reducer_: Reducer<ConfiguratorState, ConfiguratorStateAction> = (
     case "CHART_FIELD_CHANGED":
       return handleChartFieldChanged(draft, action);
 
+    case "CHART_SHOW_LEGEND_TITLE_CHANGED":
+      if (isConfiguring(draft)) {
+        const chartConfig = getChartConfig(draft);
+
+        if (isSegmentInConfig(chartConfig) && chartConfig.fields.segment) {
```
bprusinowski, line 663: Do we still need to make `fields.segment` check? 🤔

app/configurator/configurator-state/segment-config-state.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { ChangeEvent } from "react";
+
+import { isSegmentInConfig } from "@/config-types";
+import { getChartConfig } from "@/config-utils";
+import { useConfiguratorState } from "@/src";
+import useEvent from "@/utils/use-event";
+
+import { isConfiguring } from ".";
+
+export const useLegendTitleVisibility = (target: "showTitle") => {
```
bprusinowski, line 11: Nit: I think we can remove `target` and hard-code the "showTitle" in the function body, I don't see the function being used in other contexts, at least at the moment :)

app/configurator/configurator-state/segment-config-state.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const [state, dispatch] = useConfiguratorState(isConfiguring);
+  const chartConfig = getChartConfig(state);
+  const onChange = useEvent((e: ChangeEvent<HTMLInputElement>) => {
+    if (
+      isSegmentInConfig(chartConfig) &&
+      chartConfig.fields.segment?.[target]
```
bprusinowski, line 17: @noahonyejese I believe this makes the toggle non-usable at the moment, I can't toggle it. I think we don't need this condition, as it would only allow users to disable currently active toggle, not the other way around.

bprusinowski (commented):
Thanks @noahonyejese, I've left some comments below 👍 

For the question on if the toggle should be on by default, I think we should check with @KerstinFaye, it was set to "false" in the past, maybe we should leave it like that. In this case, we don't need to change the chart config initialization logic – otherwise, you'd need to look for the places where it's defined, I think we have `getInitialConfig` function, but there might be other places, I'd check configurator state actions to see if there are other ways we initialize them 🔍

app/charts/shared/legend-color.tsx
noahonyejese: Ahh 😂 I thought it's not visible in the design I didn't see the subtle change 😅

app/configurator/configurator-state/segment-config-state.tsx
noahonyejese: yep I removed it but I'm currently having issue's with git

bprusinowski: Let me know if you need support 👍

e2e/filters.spec.ts
```diff
@@ -102,17 +102,16 @@ describe("Filters", () => {
 
     await actions.mui.selectOption("Kanton");
 
-    const legend = page.locator('[data-testId="legend-title-visible"]');
+    const legend = page.locator('[data-testId="colorLegendTitle"]');
```
bprusinowski, line 105: Super-nit: `const legendTitle`? 🥹

app/charts/shared/legend-color.tsx
```diff
@@ -184,22 +188,37 @@ export const LegendColor = memo(function LegendColor(props: LegendColorProps) {
...
+        dimensionsById && (
+          <Typography
+            data-testId="colorLegendTitle"
+            component="div"
+            variant="subtitle1"
+            color={theme.palette.primary.main}
+          >
+            {dimensionsById[chartConfig.fields.segment.componentId]?.label}
+          </Typography>
+        )}
```
bprusinowski, line 211: @noahonyejese I have two comments here:
- I think we should use the same typography styles as for other legend titles, they look consistent in the design (looks like it's a caption style, but I'd double check to make sure we use the same style in e.g. Y axis title in the code),

Screenshot 2025-02-12 at 15 21 41

- we still miss the details panel wrapper that would open the panel when clicking, you can see what happens when you click the Y axis title. I'd check how it's implemented there and use the same approach 👍 

Screenshot 2025-02-12 at 15 23 15

bprusinowski (commented):
Thanks @noahonyejese, it goes into a good direction, but there are still couple of last things to improve 🙏 🔍

bprusinowski (approved):
LGTM, thanks @noahonyejese 🎉

PR #2104, opened 2025-02-14
fix: Show entire palette regardless the palette type

### LINKED ISSUE #2100: Custom color palette does not display correctly when using it in a chart
Custom color palette created in the user profile displays incorrectly when applied to charts.

1. Add a color palette to my user profile
2. Create a chart and try to use the color palette from the user profile
3. See that the color palette selection field only displays the selected color instead of the complete palette
Image

5. Try changing the color of the measure using the pipette tool. Notice that the suggested colors don't match the ones in the added color palette. 
Image

**Expected behavior** 
- The color palette selection field should display the complete palette.
- When changing colors, the suggested options should match those in the selected color palette.

- Visualize environment and version: [e.g. TEST]
- Browser and version [e.g. Chrome]

noahonyejese:
Closes #2100   

**What changes?**
This PR changes selected custom color palettes to always display all colors even if type is `single` and inside the custom color picker the swatches should always dynamically set palette colors.

## How to test

1. Go to this [link](https://visualization-tool-git-fix-y-axis-color-palette-ixt1.vercel.app/)
2. Create a Chart f.e (Einmalvergütung für Photovoltaikanlagen)
3. Click on `Y axis` 
4. Select a custom color palette, see how it displays all colors not just the used one ✅
5. Go back, and Select `Segmentations` (select `Kanton`, from the dropdown)
6. Select any color palette
7. Try to change a color with the custom color picker, See how the palettes colors are used for the swatches ✅

## How to Reproduce

1. Go to this [Link](https://test.visualize.admin.ch/en)
2. Create a Chart f.e (Einmalvergütung für Photovoltaikanlagen)
3. Click on `Y axis` 
4. Select a custom color palette, see how it displays only one color from the palette ❌
5. Go back, and Select `Segmentations` (select `Kanton`, from the dropdown)
6. Select any color palette
7. Try to change a color with the custom color picker, See how the it's always the default color palette for the swatches ❌

---

- [x] Add a CHANGELOG entry
- [x] Add e2e tests

### DIFF
--- a/app/configurator/components/field.tsx
+++ b/app/configurator/components/field.tsx
@@ -40,7 +40,11 @@ import {
 import { OpenMetadataPanelWrapper } from "@/components/metadata-panel";
 import SelectTree from "@/components/select-tree";
 import useDisclosure from "@/components/use-disclosure";
-import { ChartConfig, isColorInConfig } from "@/config-types";
+import {
+  ChartConfig,
+  CustomPaletteType,
+  isColorInConfig,
+} from "@/config-types";
 import { getChartConfig, useChartConfigFilters } from "@/config-utils";
 import {
   ControlTab,
@@ -95,11 +99,14 @@ import { useTimeFormatLocale } from "@/formatters";
 import { TimeUnit } from "@/graphql/query-hooks";
 import { Locale } from "@/locales/locales";
 import { useLocale } from "@/locales/use-locale";
+import { useUser } from "@/login/utils";
 import { getPalette } from "@/palettes";
 import { assert } from "@/utils/assert";
+import { getCustomColorPalettes } from "@/utils/chart-config/api";
 import { hierarchyToOptions } from "@/utils/hierarchy";
 import { makeDimensionValueSorters } from "@/utils/sorting-values";
 import useEvent from "@/utils/use-event";
+import { useFetchData } from "@/utils/use-fetch-data";
 
 const ColorPickerMenu = dynamic(
   () =>
@@ -746,7 +753,10 @@ export const TextBlockInputField = ({ locale }: { locale: Locale }) => {
   );
 };
 
-const useMultiFilterColorPicker = (value: string) => {
+const useMultiFilterColorPicker = (
+  value: string,
+  customColorPalettes?: CustomPaletteType[]
+) => {
   const [state, dispatch] = useConfiguratorState(isConfiguring);
   const chartConfig = getChartConfig(state);
   const filters = useChartConfigFilters(chartConfig);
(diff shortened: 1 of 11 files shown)


app/configurator/components/field.tsx
```diff
@@ -818,8 +825,20 @@ export const MultiFilterFieldColorPicker = ({
...
-    useMultiFilterColorPicker(value);
+  const user = useUser();
+
+  const { data: customColorPalettes } = useFetchData({
+    queryKey: ["colorPalettes", user?.id],
+    queryFn: getCustomColorPalettes,
+    options: {
+      enable: !!user?.id,
+    },
+  });
```
bprusinowski, line 836: @noahonyejese I wonder if it wouldn't make sense to extract this to a `useUserPalettes` hook? I think it's also used in different places in the application right now, and probably would make sense even to make sure we use the same `queryKey` across them 🤔

app/palettes.ts
```diff
@@ -67,10 +67,12 @@ export const getPalette = ({
   paletteId,
   colorField,
   colors,
+  fallbackPalette,
 }: {
   paletteId?: string;
   colorField?: ColorField;
   colors?: string[];
+  fallbackPalette?: string[];
```
bprusinowski, line 75: @noahonyejese it doesn't seem to be used anywhere, it this is the case I think we should remove it 👍

bprusinowski (commented):
@noahonyejese thanks for implementing the change, looks good in segmentation; however when opening the color picker when using a custom palette in other places (e.g. Y axis color), it doesn't have the same behavior, see the below screenshot.

Screenshot 2025-02-17 at 09 14 35

I think the behavior should be the same across the application – and this could be a good candidate for an e2e test 😄

app/palettes.ts
noahonyejese: @bprusinowski it's used in `fields line 798`

bprusinowski: Ah yes, missed that!

noahonyejese:
@bprusinowski thanks I will tackle this and also add e2e tests

noahonyejese:
@bprusinowski this should now be fixed and work for y axis and segmentation (single and multi color values).

noahonyejese:
@bprusinowski about the e2e tests what were you thinking, something like checking the colors of the select palette inside the select and comparing them to the swatches in the custom color picker?

bprusinowski:
Thanks @noahonyejese, looks good in Y color picker now 👍 

For the tests, I think we could go to an editor, change to segmentation, change palette from a default one and then open color picker and check if the first color is equal to what it should be based on new palette, and do the same for e.g. Y and segmentation fields.

I also noticed that we can't change palette type in map charts in symbol layer when we have colors based on categorical dimension – maybe you also can check this (or we can open a new issue)? See https://visualization-tool-git-fix-y-axis-color-palette-ixt1.vercel.app/en/v/dorJTo4pT6Ie?dataSource=Prod, copy and edit and see that the color palette select is missing 🥹

noahonyejese:
> Thanks @noahonyejese, looks good in Y color picker now 👍
> 
> For the tests, I think we could go to an editor, change to segmentation, change palette from a default one and then open color picker and check if the first color is equal to what it should be based on new palette, and do the same for e.g. Y and segmentation fields.
> 
> I also noticed that we can't change palette type in map charts in symbol layer when we have colors based on categorical dimension – maybe you also can check this (or we can open a new issue)? See [visualization-tool-git-fix-y-axis-color-palette-ixt1.vercel.app/en/v/dorJTo4pT6Ie?dataSource=Prod](https://visualization-tool-git-fix-y-axis-color-palette-ixt1.vercel.app/en/v/dorJTo4pT6Ie?dataSource=Prod), copy and edit and see that the color palette select is missing 🥹

@bprusinowski seems like the map issue is only happening if the symbols and area are the same dimension I think.

bprusinowski:
@noahonyejese it's also not there when Areas are not enabled, just FYI:

Screenshot 2025-02-17 at 10 17 59

noahonyejese:
> @noahonyejese it's also not there when Areas are not enabled, just FYI:
> 
> Screenshot 2025-02-17 at 10 17 59

alright thanks, I think this should be a new issue tho

bprusinowski (approved):
👌

PR #2121, opened 2025-02-19
feat: Custom map layers

### LINKED ISSUE #2068: Map Background with Raster Data (+ Vector Data)
### Description
This feature is relevant for several projects, including the "hydromaps.ch" project by the Department of Hydrology (see two PDFs below).

The initial implementation should follow a "light" version, as outlined in 2024-11-04_Rasterdaten.docx (Note: The "light version" does not refer to a specific variant from the EBP report!).

This light version can be further expanded, e.g., #262 (closed).

**Definition of Done (DoD):**

- [ ] Users can replace the default background map with a custom WMTS or WMS service using the EPSG:3857 projection (e.g., [WMTS](https://wmts.geo.admin.ch/EPSG/3857/1.0.0/WMTSCapabilities.xml?lang=en) or [WMS](https://wms.geo.admin.ch/?REQUEST=GetCapabilities&SERVICE=WMS&VERSION=1.3.0&lang=fr)).
- [ ] The default WMTS service selection is:
https://wmts.geo.admin.ch/EPSG/3857/1.0.0/WMTSCapabilities.xml?lang=.
- [ ] Users can select a layer from a WMTS service and set it as the map background.
- [ ] For time-enabled WMTS layers (e.g., SWISSIMAGE Time Travel), the map time filter is automatically synchronized at all times with the other controls and data (chart data, Visualize filters, time slider).
- [ ] Users can manually disable automatic time synchronization for WMTS layers.

**Source**
https://gitlab.ldbar.ch/bafu/visualize/-/issues/705

### LINKED ISSUE #2120: Initial Exploration on Feasibility, Constraints and Risks
Initial evaluation of the feature request. Timeboxed to .5 a day. 

References: 
Gitlab Issue -> https://gitlab.ldbar.ch/bafu/visualize/-/issues/705
Feature Page -> https://www.notion.so/interactivethings/Raster-Data-19e30da5430280e2994bc99b6919e4b8?pvs=4

bprusinowski:
Closes #2068
Closes #2120

This PR:
- adds support for adding custom map layers,
- fixes issues with `useFetchData` that were resulting in outdated data and infinite re-renders,
- make time slider work again,
- fixes invalidation of chart configs after removing one from the user profile.

## How to test

1. Go to this link.
2. Switch to a map chart.
3. Open `Base Layers` field.
4. ✅ See that Custom Layers section is visible. Click on the autocomplete inpue and see that a list of possible WMTS layers is displayed.
5. ✅ Add `112 Alarm centers` layer and see that it displayed on a map, along with a legend with: an info button to provide more context and current time period in parentheses.
6. Search for "plants" and add `Water availability for plants` layer.
7. Use drag and drop feature to put the new layer behind `112 Alarm centers` layer.
8. ✅ Toggle on the "Behind area layer" switch in `Water availability for plants` layer and see that it was moved behind the area layer.
9. Switch the toggle off, so it appears on top of the area layer again.
10. Notice that the legend title for that layer has (2018) suffix.
11. Toggle "Sync with temporal filters" in `Water availability for plants` layer.
12. ✅ See that the map changed, along with the legend suffix which is now equal to (2014).
13. ✅ Close the sub panel and change single year filter value from 2014 to e.g. 2015 and see that the map reacts.
14. Open an animation field and select `Jahr der Vergütung`.
15. ✅ Play around with the time slider and see that the map and legend react accordingly.
16. Remove animation field (set component to none).
17. Add a second chart.
18. Proceed to layout options.
19. Toggle `Jahr der Vergütung` dashboard filter.
20. ✅ Interact with the filter and see that the map and its legend react.

---

- [x] Add a CHANGELOG entry
- [x] Make it possible to put custom layers in front or behind the data layers
- [x] Check temporal filtering of layers

**Next PR**
- [ ] Check WMS data source
- [ ] Use OpenLayers to parse the XML files


app/charts/map/wmts-utils.ts
```diff
@@ -0,0 +1,81 @@
...
+  });
+};
+
+export const isValidWMTSLayerUrl = (url: string) => {
+  return url.includes("wmts.geo.admin.ch");
+};
+
+export const getWMTSLayerData = (url: string) => {
+  return url
+    .replace("{Time}", "current")
```
Rdataflow, line 77: @bprusinowski FYI this value can differ from `current`
you may switch this to the parsed value of the WMTSCapabilities.xml
`<Layer><Dimension> <ows:Identifier>time</ows:Identifier><Default>current</Default>`
i.e. for Layer "Wasserverfügbarkeit für Pflanzen" default is a number i.e. `2018`

nb: did you see there is wmts parsing and map embedding available in openlayers framework which can be included in deck.gl ? (see my comments in https://gitlab.ldbar.ch/bafu/visualize/-/issues/705#:~:text=Kontext)

app/charts/map/wmts-utils.ts
```diff
@@ -0,0 +1,81 @@
...
+        return res.text().then((text) => {
+          return parser.parse(text).Capabilities.Contents.Layer;
+        });
+      });
+    },
+  });
+};
+
+export const isValidWMTSLayerUrl = (url: string) => {
+  return url.includes("wmts.geo.admin.ch");
```
Rdataflow, line 72: FYI there exist many more WMTS / WMS sources i.e. some of them are made available through [a prefilled list of values on map.geo.admin.ch](
https://github.com/geoadmin/web-mapviewer/blob/develop/packages/mapviewer/src/modules/menu/components/advancedTools/ImportCatalogue/external-providers.json)
alternatively a visitor of https://map.geo.admin.ch may type any other on his own
you may try yourself ;-)
image

app/charts/map/wmts-utils.ts
bprusinowski: @Rdataflow thank you, yes, that's right – I am still exploring with this PR and will definitely change that, also to try to combine temporal filtering coming from Visualize with the compatible layers 😄 

I've managed to get parsed information already, but I'll also take a look at openlayers to see if we gain something by switching to them (e.g. it would be nice to have "correct" types :))

Rdataflow: @bprusinowski 💯 I see you are well on track to 🥇

app/charts/map/wmts-utils.ts
sosiology: Thank you for your input, @Rdataflow. 
We are still in the exploration phase of this feature, and determining which approach works best for Visualize. As part of the exploration phase, we will internally evaluate the pros and cons of allowing users to add custom map sources and possible ways to implement this effectively. We'll reach out if we have any questions or once we have concrete proposals to share.

cc: @bprusinowski @adintegra

app/charts/map/wmts-utils.ts
Rdataflow: @bprusinowski to keep in mind before you reinvent the wheel 😉 
- there might also be different `Dimension`s than `Time`/year i.e. day, but also `elevation` or whatever dimension occurs in a dataset. (this is where openlayers can help to not reinvent the wheel)
- there are also WMTS using key-value-pair i.e. `?key1=value1&key2=value2` params which is already available in openlayers
- also WMS and the whole aspects of the OGC Standards of the geo stack (reprojection, etc.) are readily available in OL

app/configurator/config-form.tsx
```diff
@@ -132,6 +132,20 @@ export const useChartFieldField = ({
...
+          locale,
+          field,
+          componentId: dimensionId,
+          selectedValues: leaves,
+        },
+      });
+
+      if (!get(chartConfig, `fields["${field}"].paletteId`)) {
+        return;
+      }
```
noahonyejese, line 147: I changed a lot here in the persisting PR, also note that `fields["${field}"].paletteId` will not work in some cases (maps & tables) since the have `symbolLayer` or `areaLayer` as field but still need `color` before accessing `paletteId` if you want to exclude maps and tables just use `isColorInConfig(chartConfig)`

noahonyejese (approved):
@bprusinowski Overall this PR looks pretty good, I think if e2e aren't priority right now I don't see a 100% need that you implement any. It would be nice to have but yeah time is pushing.

- About this `WMTS` stuff if this is the only url that will be supported ignore all my comments about this, otherwise I would prefer we follow a more generic approach to accommodate other sources in the future 

Other than that great stuff, feels like this added a whole new dimension to visualize! 🎉🧀

app/configurator/components/custom-layers-selector.tsx
```diff
@@ -0,0 +1,89 @@
...
+          parseAttributeValue: true,
+        });
+
+        return res.text().then((text) => {
+          return parser.parse(text);
+        });
+      });
+    },
+  });
+  const loading = !data && !error;
```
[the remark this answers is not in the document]
bprusinowski: Looks like this was changed in the process 👍

app/utils/chart-config/versioning.ts
```diff
@@ -2185,6 +2212,45 @@ export const configuratorStateMigrations: Migration[] = [
...
+    },
+  },
+  {
+    description: "ALL (bump ChartConfig version)",
+    from: "4.3.0",
+    to: "4.4.0",
+    up: async (config) => {
+      const newConfig = { ...config, version: "4.4.0" };
+
+      const chartConfigs: any[] = [];
```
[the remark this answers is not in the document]
bprusinowski: Technically not :) I removed this from all versioning entries 👍

app/configurator/components/chart-configurator.tsx
```diff
@@ -881,7 +881,7 @@ const ChartFields = (props: ChartFieldsProps) => {
               key={field}
               value={field}
               icon="baseLayer"
-              label={<Trans id="chart.map.layers.base">Map Display</Trans>}
+              label={<Trans id="chart.map.layers.base">Base Layers</Trans>}
```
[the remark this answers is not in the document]
bprusinowski: Thanks @noahonyejese, I see 👍 Generally we use both patterns in the application and I'd say there's not a practical difference in their usage if types allow `ReactNode` – as I just changed the translation here, I'll leave if like this if it's fine, as we'd need to review everything if we really wanted to get rid of that :)

app/charts/map/wmts-utils.ts
```diff
@@ -0,0 +1,81 @@
+import { XMLParser } from "fast-xml-parser";
+
+import { useLocale } from "@/locales/use-locale";
+import { useFetchData } from "@/utils/use-fetch-data";
+
+type WMTSData = {
```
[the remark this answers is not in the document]
bprusinowski: @noahonyejese generally we need to support WMTS and WMS layers with this feature, at first I focused on WMTS and indicated this in variable names – I think we will need to leave this separation in most places, as the data we fetch doesn't have the same structure and would need to be parsed differently :/

I think the `WMTSData` type will need to be separate from `WMSData`; for other things like e.g. legends, I think it should be possible to align them (if I don't miss anything, just started to work on this 😄). I started to tackle this in https://github.com/visualize-admin/visualization-tool/pull/2141 👍

app/configurator/config-form.tsx
bprusinowski: Thanks @noahonyejese, I needed to tackle this already, as I discovered that my configurator state were removed after adding animation field and refreshing the page 😅 

Here the problem was about the thing that without this check, we were dispatching `CHART_PALETTE_CHANGED` with the current `field` without checking if the field should actually include color. This led to having an animation field with paletteId, which was leading to broken decoding of configurator state :(

Generally this logic is only fired when we change "main" component id of a given field – I checked and maps still work as previously and we don't add color-related properties now when it doesn't make sense. Hope it makes sense 👍

bprusinowski:
Thanks for review @noahonyejese, in fact there's one new end-to-end test to make sure the slider works 🥹 I think for the rest we'd need to come back to them in the future 👍

noahonyejese:
I saw I think that's fine 👍🏽

app/charts/map/map.tsx
```diff
@@ -475,3 +509,15 @@ const ControlButton = ({
     </Button>
   );
 };
+
+const isValidWMTSLayerUrl = (url: string) => {
+  return url.includes("wmts.geo.admin.ch");
+};
```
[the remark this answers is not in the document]
bprusinowski: Yes, that's a good idea 💯 I'll tackle this in https://github.com/visualize-admin/visualization-tool/pull/2141 where WMS data sources will be added, it should come naturally then 😄

PR #2132, opened 2025-02-21
Persisting colors across chart type changes and field changes

### LINKED ISSUE #2097: ✓ Color Palette Settings should persist when user changes chart type
### Description 
Currently, if the user defines a color in the chart segmentation settings (manually or by choosing a color palette), and then changes the chart type the colors are reset. 

https://github.com/user-attachments/assets/ac9416e3-9444-445e-9d37-1a4b5dbe5a00

### Describe the solution you'd like
The Colors should not reset when the chart type is changed but should persist as the user edits the chart.

noahonyejese:
Closes #2097 

**What changes?**
This PR allows colors to persist over chart type changes.

## How to test

1. Go to this [link](https://visualization-tool-git-feat-color-persistance-chart-change-ixt1.vercel.app/) 
2. Create a visualization 
3. Choose Segmentation 
4. Select any color palette (custom or default)
5. Change chart type
6. See how colors persist ✅

## How it's currently 
1. Go to this [link](https://test.visualize.admin.ch) 
2. Create a visualization 
3. Choose Segmentation 
4. Select any color palette (custom or default)
5. Change chart type
6. See how colors revert to default palette ❌

---

- [x] Add a CHANGELOG entry


app/config-adjusters.ts
```diff
@@ -132,6 +136,7 @@ type AreaAdjusters = BaseAdjusters<AreaConfig> & {
   fields: {
     x: { componentId: FieldAdjuster<AreaConfig, string> };
     y: { componentId: FieldAdjuster<AreaConfig, string> };
+    color: FieldAdjuster<ColumnConfig, ColorField>;
```
bprusinowski, line 139: @noahonyejese it looks like the type of chart config is wrong in a few places (`ColumnConfig` is used in area, line, bar and column adjusters) 👓

app/charts/index.ts
```diff
@@ -1062,25 +1058,19 @@ const chartConfigsAdjusters: ChartConfigsAdjusters = {
...
             draft.fields.segment = newSegment;
-            draft.fields.color = newColor;
           }
         });
       },
+      color: ({ oldValue, newChartConfig }) => {
+        return produce(newChartConfig, (draft) => {
+          draft.fields.color = oldValue as ColumnFields["color"];
+        });
+      },
```
bprusinowski, line 1073: @noahonyejese I think the type should be correct without casting, do you know what was the type of `oldValue` if you didn't cast it? If it's `ColorField | undefined`, as we don't have color fields in tables and maps, I think it would be safer to have a check if `oldValue` is defined and return early, not to cast like this and potentially assign `undefined` to a color field.

Let me know if I missed something 👍

app/charts/index.ts
```diff
@@ -1894,6 +1829,11 @@ const chartConfigsAdjusters: ChartConfigsAdjusters = {
           };
         });
       },
+      color: ({ oldValue, newChartConfig }) => {
+        return produce(newChartConfig, (draft) => {
+          draft.fields.color = oldValue as MeasuresColorField;
+        });
+      },
```
bprusinowski, line 1836: @noahonyejese I am not sure it's that simple unfortunately 😭 I opened a Photovoltaikanlagen cube and switched to a combo chart and got an application error when using the PR link.

Screenshot 2025-02-21 at 09 39 04

I think here we need more logic, for example – if `oldValue` was of `MeasuresColorField`, then we can assign it, if the `componentIris` of the current `y` field match these in `oldValue`, right? We'd probably need to have different check for every combo chart, as they have different properties for storing component iris.

app/charts/index.ts
```diff
@@ -1013,36 +1020,25 @@ const chartConfigsAdjusters: ChartConfigsAdjusters = {
...
         if (oldChartConfig.chartType === "table") {
-          const maybeSegmentAndColorFields =
-            convertTableFieldsToSegmentAndColorFields({
-              fields: oldValue as TableFields,
-              dimensions,
-              measures,
-            });
+          const maybeSegmentField = convertTableFieldsToSegmentFields({
+            fields: oldValue as TableFields,
+          });
```
bprusinowski, line 1034: @noahonyejese I am not exactly sure where the code is wrong, but you can see that the colors are set wrongly when going to Photovoltaikanlagen cube, adding segmentation by Kanton, switching to a table chart and back to a column chart – we get all blue bars then 👀 

Screenshot 2025-02-21 at 09 47 35

app/charts/index.ts
```diff
@@ -1172,25 +1151,19 @@ const chartConfigsAdjusters: ChartConfigsAdjusters = {
...
             draft.fields.segment = newSegment;
-            draft.fields.color = newColor;
           }
         });
       },
+      color: ({ oldValue, newChartConfig }) => {
+        return produce(newChartConfig, (draft) => {
+          draft.fields.color = oldValue as BarFields["color"];
+        });
+      },
```
bprusinowski, line 1166: @noahonyejese another thing I've noticed is when you go to a column chart in Photovoltaikanlagen cube, switch X axis to Kanton, add segmentation by year, switch to a line chart and then back to a column chart – the color is black then 🤔 

Screenshot 2025-02-21 at 09 50 44

app/charts/index.ts
bprusinowski: Also, for every other non-combo chart, we most probably can't just carry over `oldValue` as cast its type, as it can happen that we switch from a combo chart to a regular chart – then we also need to probably extract some information from that `MeasuresColorField` and convert it into a single color field 👍

bprusinowski (changes requested):
Hey @noahonyejese, thanks for implementing this 🙇‍♂️ 

It goes into a good direction, but there are some tweaks need before we go further (e.g. we have an application crash when switching to a combo chart). Let me know if some of my comments are not clear 👍

app/charts/index.ts
noahonyejese: @bprusinowski It's not `ColorField | undefined` it's in fact just `ColorField` however `ColorField` has 3 types, `single`, `segment` and `measure` however not all Chart types need all 3, but it would probably make sense to migrate for most charts from `measure` type to something else.

app/charts/index.ts
noahonyejese: Yeah true I just understood the whole picture thanks for the insights

app/charts/index.ts
noahonyejese: 😵‍💫

app/charts/index.ts
bprusinowski: I see, yeah, then we need to handle each potential type in a different way here 👍

app/charts/index.ts
bprusinowski: @noahonyejese I checked the same flow as above, and now it looks like the segmentation is not removed when switching to a column chart, and the chart gets broken 🥲 

Screenshot 2025-02-24 at 09 03 03

bprusinowski: (or it's just the colors that are wrong)

app/charts/index.ts
bprusinowski: @noahonyejese this is also still the same unfortunately :(

bprusinowski:
Closing in favor of https://github.com/visualize-admin/visualization-tool/pull/2274

Issue #1810, opened 2024-10-31
[6.2] Improve Opendata.swiss appearance


sosiology:
### Description
Opendata.swiss has improved the display of the harvested cubes from LINDAS:

![Image](https://github.com/user-attachments/assets/75770a86-b5bb-44f3-a24d-9b12b3941ebc)

They embed the resource, which appears in the CKAN endpoint under the name "visualize.admin.ch," as an iframe—this is just a working theory, as I’m currently in contact with Opendata.swiss to confirm if that’s really the case. The link used for this resource is the one for the cube’s preview. Unfortunately, this now means that the Visualize application could potentially be accessed via any dataset, which could confuse users—especially since the preview page also contains metadata displayed further down.

My suggestion would be for us to offer something along these lines:
![Image](https://github.com/user-attachments/assets/c63ede6e-0473-4195-98bf-15f4a960891a)

A simplified version of this table, without the title, a reduced header and footer, and a limited selection of data—since it's a preview, displaying between 20 and 100 entries would probably suffice. This would be part of the Visualize API. This API call could then be used as a substitute for the link in the CKAN endpoint.

### DOD

- [ ] Conceptualize an alternative approach on how to best integrate Visualize into Open Data Swiss

sosiology:
@KerstinFaye @bprusinowski based on the discussion with Marco and BFS on 25.11.2024 here is the proposed scope of this task: 

- Develop a custom iframe for open data swiss. The iframe should include:
     - a table view of the dataset (analog the current preview) 
     - The back to the datasets button should be removed 
     - The metadata sidebar should be removed
     - The header should be removed, including the language selection
     - The download data option below the table should be removed. 
     - A call to action button at the top of the table: "Visualierung auf Visualize mit diesen Datensatz erstellen"
     - The button should be prominent and inform the user that they will be directed to a new page 
     - Clicking the button should open a new tab, the target is the preview of the data set in visualize

- The iframe should be dynamic in size, we are clarifying if it is possible to add a script tag see: https://gitlab.ldbar.ch/bafu/visualize/-/issues/709#note_17874

bprusinowski:
@sosiology assuming the design will be mostly like dataset preview we have now (and that we don't add new UI embedding options, but e.g. simply a new URL with cube iri, like `/preview/cube-iri`), with some removed elements, it's not a big task – I'd say it shouldn't take more than 8-12h of development, but let's see how @KerstinFaye would like to approach this 👍

[...]

PR #2138, opened 2025-02-24
Create optimal Iframe view for ODS

### LINKED ISSUE #2059: Implement new Iframe for Opendata.swiss
Develop the new iframe for Opendata.swiss based on the design

noahonyejese:
Closes #2059

**What changes?**
This PR adds an improved view of Dataset Preview for the use of ODS.

## How to test

1. Go to this [link](https://visualization-tool-git-feat-opendata-swiss-iframe-ixt1.vercel.app/) 
2. Create a Visualization w/ a Dataset
3. Go until the preview page 
4. add `&odsiframe=true` to the end of the `url` 
5. See how view is stripped of all mentioned elements under #1810

## How it was before

1. Go to this [link](https://test.visualize.admin.ch)
2. 2. Create a Visualization w/ a Dataset
3. Go until the preview page 
4. add `&odsiframe=true` to the end of the `url` 
5. See how view has all mentioned elements under #1810

---

- [x] Add a CHANGELOG entry

---

_ps: @bprusinowski just tried to implement it the best of my understanding, so there might be a couple comments here_ 

---

cc: @KerstinFaye let me know about the translations and the button in general

### DIFF
--- a/app/browser/dataset-preview.tsx
+++ b/app/browser/dataset-preview.tsx
@@ -1,3 +1,5 @@
+import { ParsedUrlQuery } from "querystring";
+
 import { Trans } from "@lingui/macro";
 import { Box, Button, Paper, Theme, Typography } from "@mui/material";
 import { makeStyles } from "@mui/styles";
@@ -7,6 +9,7 @@ import { useRouter } from "next/router";
 import React, { useCallback, useEffect } from "react";
 
 import { DataSetPreviewTable } from "@/browse/datatable";
+import { CHART_RESIZE_EVENT_TYPE } from "@/charts/shared/use-size";
 import { useFootnotesStyles } from "@/components/chart-footnotes";
 import { DataDownloadMenu } from "@/components/data-download";
 import Flex from "@/components/flex";
@@ -19,8 +22,11 @@ import {
 } from "@/graphql/query-hooks";
 import { DataCubePublicationStatus } from "@/graphql/resolver-types";
 import { useLocale } from "@/locales/use-locale";
-import { CHART_RESIZE_EVENT_TYPE } from "@/charts/shared/use-size";
 import { useResizeObserver } from "@/utils/use-resize-observer";
+
+export const isOdsIframe = (query: ParsedUrlQuery) => {
+  return query["odsiframe"] === "true";
+};
 
 const useStyles = makeStyles<Theme, { descriptionPresent: boolean }>(
   (theme) => ({
@@ -149,8 +155,6 @@ export const DataSetPreview = ({
     const { dataCubeMetadata } = metadata;
     const { dataCubePreview } = previewData;
 
-    const isOdsIframe = router.query["odsiframe"] === "true";
-
     return (
       <Flex ref={ref} className={classes.root}>
         {dataCubeMetadata.publicationStatus ===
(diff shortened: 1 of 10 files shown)


app/browser/dataset-preview.tsx
```diff
@@ -127,6 +131,14 @@ export const DataSetPreview = ({
     window.scrollTo({ top: 0 });
   }, []);
 
+  const handleHeightChange = useCallback(
+    ({ height }: { width: number; height: number }) => {
+      window.parent.postMessage({ type: CHART_RESIZE_EVENT_TYPE, height }, "*");
```
bprusinowski, line 136: 👍

app/browser/select-dataset-step.tsx
```diff
@@ -195,7 +198,13 @@ const SelectDatasetStepContent = ({
     leading: true,
   });
   const router = useRouter();
-  const classes = useStyles({ datasetPresent: !!dataset, variant });
+  const isOdsIframe = router.query["odsiframe"] === "true";
```
bprusinowski, line 201: @noahonyejese could be good to extract to an util function, as it's now used across two different files 👍

bprusinowski (commented):
Thanks @noahonyejese 💯 Generally looks good with two comments:
- It looks like we have a brief flash of the Header when reloading the page. I guess it's like this as the router didn't  have a chance to mount at first, and the query param is undefined for a while. It looks like you have an access to `AppLayout` component in `/browse` page file, here I'd try to get the query param in `getServerSideProps`, add a parameter "showHeader" in `AppLayout` and remove the condition from the Header itself to make sure it's only rendered when needed.
- I tested the iframe in iframe tester (https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_Iframe) along with a resize script, and it works mostly fine, but there is still some vertical scroll which should not be there:
Screenshot 2025-02-24 at 09 23 24

Could you take a look at these? Let me know if I can support 👍

bprusinowski:
@noahonyejese I also forgot to mention that clicking the new chart button should open in a new tab – otherwise we start making a chart in the iframe, which probably shouldn't be the case 👍

app/browser/dataset-preview.tsx
```diff
@@ -166,9 +180,15 @@ export const DataSetPreview = ({
               className={classes.createChartButton}
               component="a"
             >
```
wereHamster, line 182: @noahonyejese you probably want `target="_top"` here, otherwise only the page within the iframe will navigate.  same in the other `Create visualization on visualize from this dataset` link below…

app/pages/browse/index.tsx
```diff
@@ -15,11 +17,20 @@ export type BrowseParams = {
...
+  odsiframe?: string;
+};
+
+export const getServerSideProps: GetServerSideProps = async ({ query }) => {
+  return {
+    props: {
+      odsiframe: query.odsiframe === "true" ? true : false,
+    },
+  };
 };
```
bprusinowski, line 29: 💯

bprusinowski (approved):
LGTM, thanks @noahonyejese 🎉

bprusinowski:
@noahonyejese sorry, in fact there's still the new tab thing that was not tackled, right? I tried to click the button in iframe tester and it opened inside iframe 👀

app/browser/dataset-preview.tsx
```diff
@@ -197,9 +202,10 @@ export const DataSetPreview = ({
               }&dataSource=${sourceToLabel(dataSource)}`}
               passHref
               legacyBehavior
+              target={isOdsIframe(router.query) ? "_top" : undefined}
```
bprusinowski, line 195: @noahonyejese I think here it needs to be `_blank`, not `_top` in order to open in a new tab 🤔

PR #2142, opened 2025-02-25
feat: Added LinkExternal Icon

noahonyejese:
Added External Link Icon to the ODS button in the iframe

### DIFF
--- a/app/browser/dataset-preview.tsx
+++ b/app/browser/dataset-preview.tsx
@@ -20,6 +20,7 @@ import {
   useDataCubePreviewQuery,
 } from "@/graphql/query-hooks";
 import { DataCubePublicationStatus } from "@/graphql/resolver-types";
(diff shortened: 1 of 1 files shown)


app/browser/dataset-preview.tsx
```diff
@@ -181,7 +183,7 @@ export const DataSetPreview = ({
                 </Trans>
               ) : (
                 <Trans id="browse.dataset.create-visualization-visualize">
-                  Create visualization on visualize from this dataset
+                  Create with visualize.admin
```
bprusinowski, line 186: @noahonyejese I think you also need to change it in the .po file, this won't really make a difference if it was already defined in .po file 👍

bprusinowski (approved):

app/browser/dataset-preview.tsx
noahonyejese: It's already like this in the po just changed here to align @bprusinowski

PR #2153, opened 2025-02-26
fix: schemeCategory10 -> category10

noahonyejese:
Just noticed that I accidentally called the schema., `schemeCategory10` instead of just `category10` 😬


bprusinowski (approved):
👍

Issue #2057, opened 2025-01-31
Show all Values: Implementation


sosiology:
Implementation of the Annotations as defined in the design concept

sosiology:
Tested Show values for Lines, Bars, Areas, Columns and Pies ✅ 
Tested Segmentation for Bars and Columns (Stacked)

Feedback should be provided in new issues (FYI @KerstinFaye )

PR #2167, opened 2025-03-04
feat: Show all values (area and line charts)

bprusinowski:
Contributes towards #2057

This PR adds support for showing value labels in area and line charts (non-segmented).

## How to test

1. Go to [this link](https://visualization-tool-git-feat-show-all-values-lines-ixt1.vercel.app/en/create/new?cube=https://environment.ld.admin.ch/foen/ubd01041prod/4&dataSource=Prod).
2. Open `Vertical Axis` and turn `Always show values` on.
3. Switch to a line chart.
4. ✅ See that the `Always show values` option was persisted and that the value labels are shown.
5. ✅ See that there are no value labels overlaps, as we prevent such situations by hiding overlapping value labels.
6. ✅ Change browser width and see how the value labels adapt to available screen size.
7. ✅ Switch to an area chart are repeat the above steps to see that the behaviors are the same.
8. Go to [this link](https://visualization-tool-git-feat-show-all-values-lines-ixt1.vercel.app/en/v/jvQsngSeoG-e?dataSource=Prod).
9. ✅ Take a closer look at first and last visible value labels and see that they are shifted to not overlap the Y axis ticks (first) and not go outside of the chart (last).

---

- [x] Add a CHANGELOG entry


app/charts/line/show-values-utils.ts
```diff
@@ -0,0 +1,101 @@
...
+    dimensions: Dimension[];
+    measures: Measure[];
+  }
+): ShowLineValueLabelsVariables => {
+  const { showValues = false } = y;
+  const yMeasure = measures.find((d) => d.id === y.componentId);
+
+  if (!yMeasure) {
+    throw Error(
+      `No dimension <${y.componentId}> in cube! (useShowLineValueLabelsVariables)`
```
noahonyejese: Not sure here but if there is no dimension will this not be undefined?

app/charts/line/lines.tsx
```diff
@@ -88,14 +88,16 @@ export const ErrorWhiskers = () => {
   return <g ref={ref} />;
 };
 
-export const Lines = () => {
+export const Lines = ({
+  dotSize,
+}: {
+  dotSize: LineConfig["fields"]["y"]["showDotsSize"];
```
noahonyejese: super-nit; could start of from LineFields

app/charts/area/areas.tsx
```diff
@@ -37,21 +44,47 @@ export const Areas = () => {
     });
   }, [series, areaGenerator, colors]);
 
+  const valueLabelsData = useRenderTemporalValueLabelsData();
+
   useEffect(() => {
-    if (ref.current) {
-      renderContainer(ref.current, {
+    const g = ref.current;
```
noahonyejese, line 50: Maybe it's just me but just having g feels a bit odd 🤷🏽‍♂️

noahonyejese (approved):
@bprusinowski Overall looks good to me super quick, good idea splitting the chart types into their own PRs 🎉

app/charts/area/areas.tsx
bprusinowski: I see, thanks for raising this @noahonyejese. I generally agree that it can be a bit weird, but it's also a quite common convention when writing D3.js code, see that we also have it in other places since ~2 years 😅 To be consistent and not refactor everything now, I'll keep it, but we can revisit in the future if it proves to be unclear 🤞 

Screenshot 2025-03-05 at 11 20 42

PR #2175, opened 2025-03-12
feat: Add InlineLoading

### LINKED ISSUE #2164: Map Layers: Optimize Loading Indicator for Legend
Sometimes the map legend from the external source takes a while to load, we would probably want to allow a bit more space between the data set information and the loading screen, or maybe consider using another type of loading screen (e.g. just the spinner without the text)? That being said, I was using INT-Uncached when i first tested and could only reproduce the issue with that endpoint.

Image

- Option 1: Add spacing between the loading indicator and filter information (keep spinner and text as is)
- Option 2: Only show spinner element, and not text
- Option 3: Only show spinner, with text "loading data" and not "While we are continuously optimizing performance, processing large data sets may current require additional time)

    - Example: Loading Data Spinner
    Image 
    - Example: Loading Data Spinner with extended text
Image

- None

**Use cases and impact** 
Having a loading spinner provides users with feedback that the elements on the page are still being processed and not fully rendered yet. Showing only the spinner, and removing the current  text would not make the user think that this has to do with performance. 

**Additional context** Add any other context or screenshots about the feature
request here. If the enhancement concerns the editor, please provide a dataset,
class of datasets. If the enhancement concerns publication, please provide an
example of where the chart is published.

bprusinowski:
Closes #2164

This PR improves the styles of custom map legends loading indicator.

## How to test

1. Go to [this link](https://visualization-tool-git-feat-optimize-custom-map-leg-2393c5-ixt1.vercel.app/en/v/dMuqmcsH31D8?dataSource=Prod).
2. ✅ See that the loading indicator doesn't take the full width, doesn't switch to "loading this dataset may require additional time" after 7.5s and that it's positioned further away from the footnotes.

---

- [x] Add a CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ You can also check the
 
 - Features
   - It's now possible to show segment value labels in column and bar charts
+- Styles
+  - Optimized the custom map legends loading indicator appearance
 
 # [5.4.0] - 2025-03-11
 
--- a/app/charts/map/map-custom-layers-legend.tsx
+++ b/app/charts/map/map-custom-layers-legend.tsx
@@ -7,7 +7,7 @@ import {
   ParsedWMTSLayer,
   useWMTSLayers,
 } from "@/charts/map/wmts-utils";
-import { Error, Loading } from "@/components/hint";
+import { Error, InlineLoading } from "@/components/hint";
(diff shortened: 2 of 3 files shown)


noahonyejese (approved):
👍

PR #2176, opened 2025-03-12
fix: Field naming

### LINKED ISSUE #2173: Rename "Vertical Axis" for Pie Charts in the sidebar to "Measures"
To maintain consistent terminology and avoid user confusion, we should rename "Vertical Axis" to "Measures" in the sidebar's measure selection area. 

Image

bprusinowski:
Closes #2173

This PR makes sure we use correct field labels in the encoding options panel.

## How to test

1. Go to this link.
2. Switch to a pie chart.
3. Open a `Measure` field.
4. See that the section title is correctly set to `Measure` instead of `Vertical Axis`.

---

- [x] Add a CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ You can also check the
 - Features
   - Added an option to show all values in area, bar, column, line and pie charts
   - X axis titles are now always displayed
+- Fixes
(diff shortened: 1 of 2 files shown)


noahonyejese (approved):
👍

Issue #2064, opened 2025-01-31
Parameterizable URL: Implementation


sosiology:
Implementation of a Parameterizable URL to allow for all filter options to be configurable via URL parameters so that visualizations with specific filters can be embedded.

sosiology:
thanks for preparing @bprusinowski. I tested it with the instructions in the PR and the chart loaded in the iframe preview. As the enhancement is already merged to test, i also tested it by changing the first part of the URL from `https://visualization-tool-git-feat-parametrisable-url-ixt1.vercel.app` to `http://test.visualize.admin.ch`and it worked. After that I am not sure what else to test / do but I assume our expert users will...

bprusinowski:
Thanks @sosiology, generally you could pass any configurator state to the URL as described in the docs (https://test.visualize.admin.ch/docs/charts/preview-via-api) and it should work :) I've just seen that we have a typo in debug flag name in the docs, I'll try to fix it today.

PR #2195, opened 2025-03-18
Support longer HTTP headers

adintegra:
Potentially supports #2064 

This PR includes a Node server configuration parameter for supporting long HTTP headers up to 64KB in length (default is 8K).

## How to test

Currently, there is no possibility of testing the effectiveness of this parameter in our deployment environments, as it is dependent on other configuration changes (load balancer, WAF).

---

- [x] Add a CHANGELOG entry

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,3 +3,5 @@
 *Dockerfile*
 # Ignore node_modules to prevent issues with wrong binaries (e.g. sentry cli)
 node_modules
+.DS_Store
+*.log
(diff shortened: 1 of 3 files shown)


noahonyejese (approved):
LGTM ✅

bprusinowski (approved):
Thanks @adintegra, LGTM – btw, I believe current default since Node 13 is 16kB (https://nodejs.org/docs/v22.14.0/api/cli.html#--max-http-header-sizesize) 🤔

adintegra:
Ah, thanks @bprusinowski 👍🏼 They must have changed the default at some point in that case (maybe motivated by [this discussion](https://github.com/nodejs/node/issues/24692))

PR #2197, opened 2025-03-19
fix: Consolidate multiple Node options

adintegra:
Potentially supports #2064 

Fix for previous PR (#2195) as the GC heap value was being overwritten with the option for long header support.

## How to test

Currently, there is no possibility of testing the effectiveness of this parameter in our deployment environments, as it is dependent on other configuration changes (load balancer, WAF).


bprusinowski (approved):
👍

PR #2198, opened 2025-03-20
feat: Interactive filters toggle changes

### LINKED ISSUE #2192: Improved Filter Visibility: Remove Hide Filters when default is set to "open"
### Description 
When the default is set to show interactive filters as open, the "hide filter" option to hide them again is not needed.

Image

### LINKED ISSUE #2193: Improved Filter Visibility: Change default to "Open"
### Description 
In the current implementation (INT v5.4.1) the user has a checkbox to "Show filter area open per default" the logic should be reversed so that per default the interactive filters are always shown open **unless** the user checks the box, in which case they will be shown closed. 

Already published charts should not be automatically be affected by this change. 

Image

bprusinowski:
Closes #2192
Closes #2193

This PR sets the state of interactive filters to open by default. It also removes the function to collapse filters when this option is active.

## How to test

1. Go to [this link](https://visualization-tool-git-feat-interactive-filters-tog-5be719-ixt1.vercel.app/en/create/new?cube=https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/10&dataSource=Prod).
2. Enable `Kanton` interactive filter.
3. ✅ See that it's shown open and that `Hide filters` toggle is missing.

@KerstinFaye @sosiology I know the client has already decided about removing `Hide filters` toggle, if I understood correctly, to save some space – I just wanted to point that removing it doesn't save any space, as it's rendered in the same row as `Details` panel button (and not having a way to hide a long list of filters could be frustrating).

See opened mobile views in both cases:

Open panel with hide toggle removed             |  Open panel with hide toggle shown
:-------------------------:|:-------------------------:
Screenshot 2025-03-20 at 12 32 16  |  Screenshot 2025-03-20 at 12 32 19

---

- [x] Add a CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,12 @@ You can also check the
 
 # Unreleased
 
-- Extended maximum URL length in Node to 64KB
+- Features
+  - Interactive filters area is now open by default
+  - When interactive filters area is initialized open, it's no longer possible
+    to collapse it
+- Maintenance
+  - Extended maximum URL length in Node to 64KB
 
 # [5.4.1] - 2025-03-18
 
--- a/app/charts/index.ts
+++ b/app/charts/index.ts
@@ -2,8 +2,8 @@ import { t } from "@lingui/macro";
 import { ascending, descending, group, rollup, rollups } from "d3-array";
 import produce from "immer";
 import get from "lodash/get";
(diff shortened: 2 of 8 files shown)


noahonyejese (approved):
LGTM, nice! 👍

PR #2200, opened 2025-03-20
perf: Multistage Docker build

adintegra:
Closes #

This PR...

## How to test

1. Go to...

---

- [ ] Add a CHANGELOG entry

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,59 +1,109 @@
-FROM node:20
+# ----------- Base -----------
+FROM node:20-slim AS base
+RUN apt-get update -y && apt-get install -y openssl && apt-get install ca-certificates -y
 
-RUN mkdir -p /usr/src/app
+# ----------- Deps -----------
+# Install deps and build the app
+FROM base AS deps
 WORKDIR /usr/src/app
 
 # build with
-# docker build . \
-#   --build-arg PREVENT_SEARCH_BOTS=<true/false> \
-#   --build-arg COMMIT=$(git rev-parse HEAD) \
-#   --build-arg VECTOR_TILE_URL=<url of the vector service> \
-#   --build-arg MAPTILER_STYLE_KEY=<maptiler style key> \
+# docker build --no-cache . \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,59 +1,109 @@
...
+# This should eventually be fixed in Next.js, but for the time being adjusting
+# the extension seems to be the easiest path forward (thanks @wereHamster!)
+RUN mv ./app/server.js ./app/server.cjs
+
+# Let's not run as root
+USER node
+
+EXPOSE 3000
+
+# Instead of running npm start; handle signals (SIGINT/SIGTERM) properly
```
wereHamster, line 108: Newer versions of Next.js require ENV `HOSTNAME=0.0.0.0`. Might be worth adding it today to not get any surprises when you upgrade Next.js in the future.

See official Docker example: https://github.com/vercel/next.js/blob/62ac6d040183d48a5d83d2b197826c0c471caa62/examples/with-docker/Dockerfile#L65

Dockerfile
```diff
@@ -1,59 +1,109 @@
...
+COPY --from=deps --chown=node:node /usr/src/app/app/.next/static ./app/.next/static
+COPY --from=deps --chown=node:node /usr/src/app/app/public ./app/public
+
+# The file that Next.js generates is CommonJS, but the frontend folder has a
+# package.json with type:module, so node expects ESM when files have a .js
+# extension.
+#
+# This should eventually be fixed in Next.js, but for the time being adjusting
+# the extension seems to be the easiest path forward (thanks @wereHamster!)
+RUN mv ./app/server.js ./app/server.cjs
```
wereHamster, line 101: This project does not use type:module, so renaming the file to add the .cjs extension is not necessary. And I believe the latest Next.js works without this workaround (if you ever decide to switch to type:module).

wereHamster (commented):
I have not tested locally since I don't have Docker installed. But the change looks good.

PR #2227, opened 2025-04-02
feat: CI / CD pt. 4

bprusinowski:
This PR continues the CI / CD migration by:
- completing the migration of the dataset preview page,
- fixing button text overflow issues.

---

### DIFF
--- a/app/browser/select-dataset-step.tsx
+++ b/app/browser/select-dataset-step.tsx
@@ -8,7 +8,12 @@ import uniqBy from "lodash/uniqBy";
 import Head from "next/head";
 import NextLink from "next/link";
 import { Router, useRouter } from "next/router";
-import React, { ComponentProps, useCallback, useMemo } from "react";
+import React, {
+  ComponentProps,
+  type MouseEvent,
+  useCallback,
+  useMemo,
+} from "react";
 import { useDebounce } from "use-debounce";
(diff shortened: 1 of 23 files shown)


app/browser/select-dataset-step.tsx
```diff
@@ -389,7 +389,7 @@ const SelectDatasetStepContent = ({
                 initial={{ opacity: 0, height: 0, overflow: "hidden" }}
                 animate={{ opacity: 1, height: "auto" }}
                 exit={{ opacity: 0, height: 0 }}
-                transition={{ duration: DURATION }}
+                transition={{ duration: DURATION * 2 }}
```
noahonyejese: Is duration used elsewhere aswell, otherwise we could just change the duration value ig.

noahonyejese (approved):
LGTM - nice work 🎉

app/browser/select-dataset-step.tsx
bprusinowski: Good point, in fact it was multiplied to match the enter / exit total duration of other elements on a page, we only need this behavior here. I added a comment in https://github.com/visualize-admin/visualization-tool/pull/2227/commits/929e1bafceb6199a55b129ff2b6e8b986a5e6085 👍

PR #2228, opened 2025-04-02
feat: CI / CD pt. 5

bprusinowski:
This PR continues the migration to the new CI / CD by:
- adjusting the editor sidebar styles,
- adjusting loading and banner styles,
- adding additional MUI component overwrites.

---

- [ ] ~~I added a CHANGELOG entry~~


app/themes/components.tsx
```diff
@@ -154,16 +168,10 @@ export const components: Components = {
...
         margin: 0,
-        padding: 12,
-        paddingLeft: 16,
-        transition: "background-color 0.2s ease",
-
-        "&:not(.Mui-disabled):hover": {
-          backgroundColor: palette.cobalt[50],
-        },
       },
+      // @ts-ignore
```
ptbrowne, line 173: Do you know why we need ts-ignore here when we did not need it before ?

app/components/form.tsx
```diff
@@ -78,6 +77,39 @@ import { valueComparator } from "@/utils/sorting-values";
...
+  sm: {
+    ...theme.typography.caption,
+  },
+  md: {
+    ...theme.typography.body3,
+  },
+  lg: {
+    ...theme.typography.h5,
+  },
+}));
```
ptbrowne, line 90: Wouldn't it be easier to change the typography variant below ? 

```
const sizeToVariant = {  sm: 'caption', .... }

.... below ...

<MUIFormControlLabel
      {...rest}
      componentsProps={{
        typography: {
          variant: sizeToVariant[size],
          as='label'
        },
      }}
    />
```

app/configurator/components/chart-configurator.tsx
```diff
@@ -146,30 +154,33 @@ export const DataFilterSelectGeneric = ({
...
+            {`${index + 1}. ${dimension.label}`}
+          </Typography>
         </OpenMetadataPanelWrapper>
-        <InteractiveDataFilterToggle id={dimension.id} />
+        <Flex
+          sx={{
+            alignSelf: "flex-end",
+            alignItems: "center",
+            gap: 1,
+            marginRight: sideControls && dimension.isKeyDimension ? 7 : 0,
```
ptbrowne, line 172: I wonder if it would not be cleaner to manage the margin right via the parent using flex layout + gap.

ptbrowne (commented):
LGTM, some questions but nothing blocking. I did not yet test thorougly the UI, I will do that when the PR is ready.

app/themes/components.tsx
bprusinowski: Good catch, it was not needed anymore here, I'll remove it in next commits – there's a type error when spreading non-native typography variant styles in style overrides, I think before there was a `...typography.body3` here (or I had it locally for a while, before extracting it to `useStyles`) 🤔

app/components/form.tsx
bprusinowski: Yes, that's indeed easier, thanks!

app/components/data-source-menu.tsx
```diff
@@ -1,9 +1,10 @@
 import { Trans } from "@lingui/macro";
-import { MenuItem, Select, Typography } from "@mui/material";
+import { Typography } from "@mui/material";
 import { useRouter } from "next/router";
 import { useMemo } from "react";
 
 import Flex from "@/components/flex";
+import { Select } from "@/components/form";
```
ptbrowne, line 7: Nit: I wonder know if at some point it makes sense to use eslint rules to forbid use of MUI select inside app.

app/components/select-tree.tsx
```diff
@@ -448,13 +445,16 @@ const SelectTree = ({
...
+      if (e.key === "Enter" || e.key == " ") {
         handleOpen();
-        ev.preventDefault();
+        e.preventDefault();
       }
-    });
+    }
+  );
+
+  console.log(selectSizeToTypography[size]);
```
ptbrowne: 👁️

e2e/tooltip.spec.ts
```diff
@@ -21,7 +21,7 @@ test("should have correct tooltip content", async ({
   await selectors.edition.drawerLoaded();
 
   const filterLocator = await within(
-    selectors.edition.controlSectionBySubtitle("Filters")
+    selectors.edition.controlSectionByTitle("Filters")
```
ptbrowne, line 24: Question: Do you find this helpers useful ? I recently migrated to not try to have DRY code in my E2E tests in favor of generating them directly with the record feature from Playwright. Actually I think here it makes sense not to have to re-record all those tests.

app/configurator/interactive-filters/interactive-filters-configurator.tsx
```diff
@@ -97,36 +95,3 @@ export const InteractiveFiltersConfigurator = ({
...
-          }
-        >
-          <Typography variant="body2">
-            <Trans id="controls.filters.interactive.toggle">Interactive</Trans>
-          </Typography>
-        </MaybeTooltip>
-      }
-    />
-  );
-};
```
ptbrowne, line 139: 🎉

app/themes/index.ts
```diff
@@ -15,6 +15,21 @@ declare module "@mui/material" {
...
+    error: false;
+    warning: false;
+    info: false;
+    success: false;
+
+    red: true;
+    orange: true;
+    yellow: true;
+    blue: true;
+    green: true;
```
ptbrowne, line 28: Why do we have both yellow and orange alerts ?

app/themes/components.tsx
```diff
@@ -12,6 +12,18 @@ import { shadows } from "@/themes/shadows";
...
+      root: ({ ownerState }) => {
+        const color = ownerState.color ?? "blue";
+
+        return {
+          backgroundColor: palette[color][50],
+          color: palette[color].main,
+        };
+      },
+    },
+  },
```
ptbrowne, line 67: Nit: I wonder about this change, we move from a semantic Alert to a UI alert, what triggered this change ?

app/components/hint.tsx
```diff
@@ -198,152 +181,163 @@ export const LoadingOverlay = () => {
...
   return Component;
 };
 
-export const HintRed = mkHint("error", "HintRed");
-export const HintYellow = mkHint("warning", "HintYellow");
-export const HintBlue = mkHint("info", "HintBlue");
+export const HintRed = mkHint("red", "HintRed");
+export const HintGreen = mkHint("green", "HintGreen");
+export const HintBlue = mkHint("blue", "HintBlue");
+export const HintOrange = mkHint("orange", "HintOrange");
```
ptbrowne, line 353: Nit: Could be good to have comments here on when to use which. Comments that would appear when hovering the component.

app/components/hint.tsx
```diff
@@ -338,6 +338,6 @@ const mkHint = (
 };
 
 export const HintRed = mkHint("red", "HintRed");
-export const HintGreen = mkHint("green", "HintGreen");
+const HintGreen = mkHint("green", "HintGreen");
```
ptbrowne, line 351: I see this is not exported anymore, don't we use a Hint green when the chart has been published ?

ptbrowne (approved):
LGTM :rocket:

I do not know if it is expected but when goign to Storybok, I have some errors on the following pages (I would guess the errors were already there before the CI migration)

*   [https://visualization-tool-git-feat-ci-cd-5-ixt1.vercel.app/storybook/index.html?path=/docs/components-selection-tabs--docs](https://visualization-tool-git-feat-ci-cd-5-ixt1.vercel.app/storybook/index.html?path=/docs/components-selection-tabs--docs)
*   [https://visualization-tool-git-feat-ci-cd-5-ixt1.vercel.app/storybook/index.html?path=/docs/components-button--docs](https://visualization-tool-git-feat-ci-cd-5-ixt1.vercel.app/storybook/index.html?path=/docs/components-button--docs)  
     
*   [https://visualization-tool-git-feat-ci-cd-5-ixt1.vercel.app/storybook/index.html?path=/docs/components-tooltip--docs](https://visualization-tool-git-feat-ci-cd-5-ixt1.vercel.app/storybook/index.html?path=/docs/components-tooltip--docs)
*   [https://visualization-tool-git-feat-ci-cd-5-ixt1.vercel.app/storybook/index.html?path=/story/components-reactgrid--example](https://visualization-tool-git-feat-ci-cd-5-ixt1.vercel.app/storybook/index.html?path=/story/components-reactgrid--example)
*   [https://visualization-tool-git-feat-ci-cd-5-ixt1.vercel.app/storybook/index.html?path=/story/charts-charts--columns](https://visualization-tool-git-feat-ci-cd-5-ixt1.vercel.app/storybook/index.html?path=/story/charts-charts--columns)
*   https://visualization-tool-git-feat-ci-cd-5-ixt1.vercel.app/storybook/index.html?path=/docs/organisms-header--docs

Some questions on the following components:

*   [https://visualization-tool-git-feat-ci-cd-5-ixt1.vercel.app/storybook/index.html?path=/story/components-controls--control-section](https://visualization-tool-git-feat-ci-cd-5-ixt1.vercel.app/storybook/index.html?path=/story/components-controls--control-section)  inputs are empty ?  
     
*   [https://visualization-tool-git-feat-ci-cd-5-ixt1.vercel.app/storybook/index.html?path=/story/components-alerts--chart-unexpected-error](https://visualization-tool-git-feat-ci-cd-5-ixt1.vercel.app/storybook/index.html?path=/story/components-alerts--chart-unexpected-error) : I would think the icon could be vertically aligned with the title (perhaps through subgrid)
*   [https://visualization-tool-git-feat-ci-cd-5-ixt1.vercel.app/storybook/index.html?path=/story/design-system-theme--typography-story](https://visualization-tool-git-feat-ci-cd-5-ixt1.vercel.app/storybook/index.html?path=/story/design-system-theme--typography-story) is repeated here https://visualization-tool-git-feat-ci-cd-5-ixt1.vercel.app/storybook/index.html?path=/docs/components-typography--docs

app/themes/index.ts
bprusinowski: I think I initially made a mistake and thought we'd actually have two, but I see it's already corrected in later commits 👍

app/components/data-source-menu.tsx
bprusinowski: I think it could make sense, but currently we have `Option` type which is of a string type and we have at least one custom select for color picker, to render color swatches instead of regular labels – but if we manage to consolidate them, then it would be nice 👍 I think in fact it could already work for less custom components like e.g. checkboxes or switches 🤔

e2e/tooltip.spec.ts
bprusinowski: Yes, I find them useful! Here I think the header level changed to always be the same so we didn't need to have this distinction anymore – but I'd also avoid re-doing the tests, especially that it still sometimes happen that the cube stops working and we actually need to make a change 🥹

app/themes/components.tsx
bprusinowski: It was not really defined what is Visualize's "primary" or "secondary" colors – you can see that generally we have some definitions below here in Figma: https://www.figma.com/design/bMl4bRDClW4oNXSf4DUrS2/BUND---Global-Library?node-id=3621-19382&m=dev, but they are not really aligned with what we use in MUI + many times they were not linked to tokens (see e.g. here for text: https://www.figma.com/design/LrxKyljpgvadf7H0jlsqVb/visualize.admin.ch---design_NEW?node-id=9020-53504&m=dev), so it was tricky to navigate this correctly and jump between functional definitions and actual "hard-coded" color values.

That's why I decided to make it "easier" and actually use the color names instead of functional color definitions across the app, I recently tried [Mantine](https://mantine.dev/core/button/) where it's a default approach and felt it works quite well.

Do you think we should change back to semantic values here? It might align better with MUI's approach and would be a quick change for alerts, but overall (primary, secondary) with all the correct namings we need in MUI (`main`, `dark`, `contrastText`, etc, instead of e.g. `⁨⁩functional-input-on-input`) I think we need to have a broader alignment first 😱

app/components/hint.tsx
bprusinowski: Yep, the thing is that it's also defined in this file and knip was complaining about the export 😅 

```ts
export const PublishSuccess = () => {
  return (
    <HintGreen>
      <Trans id="hint.publication.success">
        Your visualization is now published. You can share and embed it using
        the URL or the options below.
      </Trans>
    </HintGreen>
  );
};
```

bprusinowski:
Thanks for reviewing @ptbrowne 🙇‍♂️

Yes, I was not relying on Storybook for quite some time, so it might be the case that it's broken, I'd assume at least partially it was from before the migration. I think it could be a good idea to take a look at this with a designer when we resume the project and see it we should switch the mode and develop more components there, in the past months it was not really being used 👀

app/themes/components.tsx
ptbrowne: > Do you think we should change back to semantic values here?

No, thanks for the heads up. I think that if this is not really aligned from the design side, then we have to make our life easier :)

PR #2230, opened 2025-04-04
feat: CI / CD pt. 6

### LINKED ISSUE #2004: Tooltip in segmented charts is very laggy for some cubes
Currently, the tooltip in charts with segmentation applied can get very laggy. After a brief investigation, it looks like the reason for that is we have a `useItemStyles` that takes three arguments that makes it very costly to re-render and re-create the class names dynamically. We should improve this by potentially separating the `LegendItem` component into a few specialized ones, instead of one that takes a `symbol` and `usage` arguments (tested in a quick way locally and the performance was greatly improved).

To reproduce, open [this link](https://test.visualize.admin.ch/en/v/Vu1PkjXY0sqJ?dataSource=Prod) in Safari and try to hover over the bars.

bprusinowski:
Closes #2004

This PR continues the migration to new CI / CD by:
- introducing closable section titles,
- correctly positioning editor drawer, which is affected by dynamic navigation size in different viewports,
- consolidating and improving legend items rendering (ditched complex `useStyles` + `::before` logic in favor of using SVG icons, which improves the performance of tooltip rendering),
- updating styles of color palette picker and multi filter panel,
- fixing an unreported issue with filter panel color swatches not having colors (go to [this INT link](https://int.visualize.admin.ch/en/create/new?cube=https://environment.ld.admin.ch/foen/nfi/nfi_T-changes/cube/2024-1&dataSource=Prod), add segmentation by `Region`, click `Edit filters` and un-toggle some hierarchies to see all white swatches),
- introducing overall improvements and adjustments to the editor components.

---

- [ ] ~~I added a CHANGELOG entry~~


app/charts/shared/legend-color.tsx
```diff
@@ -384,30 +328,29 @@ const LegendColorContent = ({
...
+                {colorValues.map((value) => {
+                  const label = getLabel(value);
                   const active = !activeInteractiveFilters.has(label);
 
                   return (
                     <LegendItem
-                      key={`${d}_${i}`}
-                      item={label}
-                      color={getColor(d)}
+                      key={value}
```
ptbrowne, line 337: I wonder why previously we had also i ? Was it because some different values could have the same name ? (canton Zürich and City Zürich?) Perhaps i is necessary ?

e2e/actions.ts
```diff
@@ -123,8 +123,7 @@ export const createActions = ({
     },
   },
   drawer: {
-    close: async () =>
-      await screen.getByRole("button", { name: "Back to main" }).click(),
+    close: async () => await screen.getByTestId("section-close").click(),
```
ptbrowne, line 126: I believe it would be btter to add aria-label="Back to main" on the Box (which should be a Button I guess) so that you can still reach it via `getByRole('button', { name: 'Back to main' })`

If not possible to make it a button, I still believe having an aria-label could be possible, and use getByLabel.

ptbrowne (approved):
LGTM 👍 , sorry it took me a long time.

Well done, I see the green check now 🎉

app/charts/shared/legend-color.tsx
bprusinowski: I checked and in fact you're right, what tricked me here is that we have `value` and `label` that sometimes are both labels, I expected the value to be an iri. I will correct this, thanks 👍

app/charts/shared/legend-color.tsx
```diff
@@ -336,16 +336,18 @@ const LegendColorContent = ({
 
   const soleItemChecked = activeInteractiveFilters.size === numberOfOptions - 1;
 
-  const handleToggle: CheckboxProps["onChange"] = useEvent((ev) => {
-    const disabled = soleItemChecked && !ev.target.checked;
-    const item = ev.target.value;
+  const handleToggle: ComponentProps<typeof Checkbox>["onChange"] = useEvent(
```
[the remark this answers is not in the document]
bprusinowski: The thing here is that what we pass to form `Checkbox` is not the same thing as in default `CheckboxProps` (`onChange` type comes from `FieldProps`, not `CheckboxProps`) – but yeah, it would make sense to do `export type CheckboxProps = ComponentProps<typeof Checkbox>` 👍

app/login/components/color-palettes/color-palette-types.tsx
```diff
@@ -122,7 +122,7 @@ const SequentialColorPaletteCreator = (props: ColorPaletteCreatorProps) => {
         />
       </Box>
       <Box sx={{ fontSize: "1rem", pb: 2, maxWidth: 304 }}>
-        <Label htmlFor="custom-color-palette-base" sx={{ mb: 1 }}>
+        <Label htmlFor="custom-color-palette-base">
```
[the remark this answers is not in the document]
bprusinowski: I don't exactly remember, but it was to align the styles with new CI / CD, so I'd hope the answer is that this is how it should be now 😅 I still see a margin between input elements and labels, so I think it was actually moved to a parent container 👍

e2e/actions.ts
bprusinowski: I agree (just to note, this is how we treated collapsible section "buttons" before, as `div` elements with an `onClick` evert handler, so at least it's aligned 😅) – will check if it doesn't break things to actually use a button here 👍

app/configurator/components/dataset-control-section.tsx
```diff
@@ -41,34 +39,32 @@ import SvgIcTrash from "@/icons/components/IcTrash";
...
+      backgroundColor: "#A6DFE7",
       lineHeight: 1.5,
       padding: "0.75rem",
       whiteSpace: "nowrap",
       maxWidth: "none",
-      // color: theme.palette.success.main,
     },
     [`& .${tooltipClasses.arrow}`]: {
-      color: theme.palette.success.light,
+      color: "#A6DFE7",
```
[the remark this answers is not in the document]
bprusinowski: Thanks for catching this, in fact it was completely wrong, not sure why I changed it 😱 It's actually still a `success` palette 👍

PR #2237, opened 2025-04-11
feat: Future limits

### LINKED ISSUE #2225: Target Values: Show future values on x-axis for Line Charts
Pending feedback if the workaround by adding 0 values to the data is viable

Target values in the future should be shown in the charts. Example: If my data covers the range between 2000-2010, it should be possible to show a target in 2030 on the chart in the form of a line.

bprusinowski:
Closes #2225

This PR:
- adds support for displaying future limits for line charts,
- fixes displaying of dots and tooltips when doesn't make sense (when there was an error in parsing a value or it's undefined),
- aligns styles related with limits.

The high-level way the future limits are displayed is that we:
1. Get related limit dimension values from all measures.
2. Select values that apply to a given dimension and fetch metadata for them (label, position).
3. Return these on top of `values`, in `relatedLimitValues`.
4. Use these `relatedLimitValues` (filtered by currently toggled limits) in relevant places to extend the data domain.

## How to test

1. Go to [this link](https://visualization-tool-git-feat-future-limits-ixt1.vercel.app/en/create/new?cube=https://environment.ld.admin.ch/foen/test_target/cube/target_future/1&dataSource=Test).
2. Switch to a line chart.
3. Go to `Vertical Axis`.
4. Change measure to `CO2 Emissions`.
5. ✅ Enable the limits and see that the X axis range is extended to show future limits, even though we don't have such observations in the data.
6. Go to `Horizontal Axis`.
7. Make the axis interactive.
8. ✅ Interact with the brush and see that it works.

---

- [ ] ~~I wrote tests for the changes (if applicable)~~ – I think an E2E test would be a good candidate here, but as the only cube that supports the feature is a draft cube on TEST, we might add them in the future
- [ ] ~~I wrote configurator and chart config migrations (if applicable)~~ – as the feature is not on PROD yet, I didn't write migrations


bprusinowski:
@ptbrowne FYI the two E2E tests that failed are not related to changes from this PR (multiple cubes found errors).

app/graphql/resolvers/rdf.ts
```diff
@@ -301,7 +301,8 @@ export const dataCubeComponents: NonNullable<
           position: r.position,
         }));
       const values = uniqBy(
-        [...relatedValues, ...rawValues],
+        // rawValues first, because we prefer these
+        [...rawValues, ...relatedValues],
```
ptbrowne: I do not remember what raw and related are 😅 , and why we would prefer those ? if you have this in mind could you write a small comment with a bit more explanation ?

app/charts/line/lines.tsx
```diff
@@ -179,17 +180,23 @@ const Points = ({ dotSize }: { dotSize: LineFields["y"]["showDotsSize"] }) => {
...
-        fill,
-      };
-    });
+    return chartData
+      .map((d) => {
+        const x = xScale(getX(d));
+        const y = yScale(getY(d) as number);
+        const fill = colors(getSegment(d));
+
+        if (y === undefined) {
```
ptbrowne, line 189: Question: I wonder why x and fill cannot be undefined also ? WOuld it make sense to also add them to this condition ?

app/charts/shared/chart-state.ts
```diff
@@ -797,20 +798,25 @@ type ValuePredicate = (v: any) => boolean;
...
     getAxisValueAsDate?: (d: Observation) => Date;
     getSegmentAbbreviationOrLabel?: (d: Observation) => string;
     getTimeRangeDate?: (d: Observation) => Date;
   }
-): Omit<ChartStateData, "allData"> => {
+): ChartStateData => {
   const { interactiveFiltersConfig } = chartConfig;
+  const allData = useMemo(() => {
+    return sortData ? sortData(observations) : observations;
+  }, [observations, sortData]);
```
ptbrowne: Nice change :)

app/charts/shared/chart-state.ts
```diff
@@ -810,22 +811,47 @@ export const useChartData = (
...
   const { interactiveFiltersConfig } = chartConfig;
   const allData = useMemo(() => {
-    return sortData ? sortData(observations) : observations;
-  }, [observations, sortData]);
+    let allData = observations;
+
+    if (axisDimensionId && limits) {
+      const dimensionValuesObservations = limits
+        .flatMap((limit) => limit.related)
+        .filter((limit) => limit.dimensionId === axisDimensionId)
```
ptbrowne, line 837: Nit: wonder if we should keep a "limit" object indexed by dimension id.

ptbrowne (approved):
LGTM, I think the only point of attention is the querying of the dimension limits, to see if it's possible to make a little less requests /

app/graphql/resolvers/rdf.ts
bprusinowski: This logic was later changed, at first I thought we can simply include these limit values in `values`, but this turned out to not be the best idea :) I could have cleaned it up beforehand, sorry for that 🫣

app/charts/shared/chart-state.ts
bprusinowski: Hmmm, in this place I am not 100% sure – generally we need to have easy access to all limits, not scoped to `dimensionId`, e.g. in color legend, to list them all. As it's usually a small number (in the cubes I worked with it was max 4 limits overall), maybe it would be fine to leave as is? 🤔 Otherwise it would either introduce a bit more complexity in other places to flatten the indexed object to an array, or require keeping both indexed object and an array 🥹

app/rdf/limits.ts
```diff
@@ -26,58 +26,93 @@ type LimitRange = BaseLimit & {
...
+            };
+          })
+          .filter(truthy);
+
+        const related = await Promise.all(
+          baseRelated.map(async (r) => {
+            const query = `PREFIX schema: <http://schema.org/>
 
-      const value = a.out(ns.schema.value).value;
+SELECT ?label ?position WHERE {
```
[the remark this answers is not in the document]
bprusinowski: Yes, that's a really good point, thanks! Will optimize this 👍

app/charts/shared/chart-state.ts
ptbrowne: Fine by me :)

app/rdf/queries.ts
```diff
@@ -97,20 +97,23 @@ export const getCubeDimensions = async ({
...
+            units: dimensionUnitIndex,
+            limits: await getDimensionLimits(dim, {
+              locale,
+              unversionedCubeIri,
+              sparqlClient,
+            }),
+          });
+        })
+      )
+    ).sort((a, b) => ascending(a.data.order, b.data.order));
```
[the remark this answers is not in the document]
bprusinowski: Hmmm, I wonder if it would be a good idea in terms of maintainability, in case we rename a variable we return from a query, we'd need to re-create a snapshot, right? This would be on top of the type errors we'd get automatically anyway. Do I miss something here?

I think in this particular feature, I'd prefer to have an E2E test to check if the domain of the chart is extended when we have future limits – I could implement it, but these cubes are still a WIP on TEST LINDAS, so I am pretty sure quite soon we'd get errors related to this 🥲

ptbrowne (approved):
⚡

PR #2263, opened 2025-04-17
feat: Custom Y axis domain

### LINKED ISSUE #1871: Implement: Y-axis scale adjustment
See Design Documented here: 
- https://github.com/visualize-admin/visualization-tool/issues/1870

bprusinowski:
Closes #1871

This PR:
- adds a way to set custom X or Y scale domains for bar, column, line and area charts (behind a `custom-scale-domain` feature flag),
- fixes a bug with property overwrites when using `FIELD_VALUE_NONE` in `CHART_FIELD_UPDATED` reducer action,
- improves styles of the options panel selector.

## How to test

1. Go to [this link](https://visualization-tool-git-feat-custom-y-axis-domain-ixt1.vercel.app/en/create/new?cube=https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/10&dataSource=Prod&flag__custom-scale-domain=true).
2. Open Vertical Axis.
3. ✅ See that there's an `Adjust scale domain` checkbox.
4. ✅ See that it's possible to change the domain by entering a valid number (min must be smaller than max) and clicking Enter or removing the focus from the input (in case of an invalid number, we get an error message).
5. ✅ Switch to a line chart and see that the custom domain was persisted.
6. ✅ Check bar, column, line and area charts with segmentation and see that they also work.
7. Enable 100% mode on a stacked chart and see that it ignores custom domain (it's also not possible to edit it in such state).

---


app/configurator/components/chart-options-selector.tsx
```diff
@@ -494,6 +490,15 @@ const EncodingOptionsPanel = (props: EncodingOptionsPanelProps) => {
                 }
               />
             ) : null}
+            {encoding.options?.adjustScaleDomain && fieldComponent ? (
+              <ChartScaleDomain
+                chartConfig={chartConfig}
+                field={field}
+                defaultDomain={encoding.options.adjustScaleDomain.getDefaultDomain(
+                  { chartConfig, observations }
+                )}
```
ptbrowne, line 493: Should the default domain be memoed ? I presume we need to go over all observations to compute the extent ?

ptbrowne: It could possibily be done in the ChartScaleDomain component.

ptbrowne (commented):
Code looks good. I just tried the feature and it works.

But it seems it comes with some weird behaviors: 

Here I am not sure if we should clip the bars that are outside the range and add an alert on the chart showing that not all data is shown ?
Nit : It seems like the year labels should be at the same z-index as the axis.

image

When I put a negative value inside the min limit, I get vertical grid lines ? 

image

bprusinowski:
Thanks for reviewing @ptbrowne 🙇‍♂️ Yeah, we didn't discuss any specific behaviors on how to handle these weird cases. I think it would probably be safest to hide the feature behind a flag and re-visit this once we are back on a project in a full team, with a designer.

I remember at some point we felt we have "too many warnings" and maybe there would be another way to indicate this fact 🤔

I will tackle the nits and negative scale issue 👍

bprusinowski:
@ptbrowne it looks like the grid lines are actually there for a reason (as the bars are not aligned to the 0 anymore normally, it's easier to connect them to X domain labels), see below:

Screenshot 2025-04-29 at 16 40 30

I've improved the labels positioning though and fixed the overlapping issue 👍

app/charts/shared/axis-width-time.tsx
```diff
@@ -54,6 +54,8 @@ export const AxisTime = () => {
     if (ref.current) {
       const axis = axisBottom(xScale)
         .ticks(ticks)
+        .tickSizeInner(hasNegativeValues ? -chartHeight : 6)
+        .tickPadding(hasNegativeValues ? 6 : 3)
```
ptbrowne, line 58: 💯

ptbrowne (approved):
Perfect 👍

PR #2285, opened 2025-04-29
Tilematrix parsing + provider information

ptbrowne:
- **feat: Add structured info for providers**

We report information on providers in a special storybook page.

- **feat: Allow to parse the tilematrix set**

This way, we would be able to gray out layers if they do not use
the ESPG:3857 projection. I have to yet see how geoadmin does that,
I wonder if it only projects coordinates at query time or if
there is a reprojection of the tiles images also (https://opendem.info/openlayers4/v4.1.0/doc/tutorials/raster-reprojection.html).

---

- **feat: Upgrade storybook, fast refresh works**

There are still some pages which do not work in Storybook but with
fast refresh, it's much better. Only an update of Storybook +
removing of the entry in app/package.json was necessary.


app/charts/map/wms-wmts-providers-extra.json
```diff
@@ -1,13 +1,93 @@
 {
-  "https://bio.discomap.eea.europa.eu/arcgis/services/Ecosystem/Ecosystems/MapServer/WMSServer": {
-    "hidden": true,
-    "note": "Not working, also not working on geoadmin"
+  "https://geodienste.ch/db/av_situationsplan/deu": {
+    "canListLayers": true,
+    "canDisplayLayers": true,
+    "canDisplayOnMapGeoAdmin": true,
+    "workingLayers": ["Berzirkgrenzen", "Situationsplan", "Landeskarte"],
+    "note": ""
```
bprusinowski, line 7: 👍

app/charts/map/wmts-utils.ts
```diff
@@ -127,41 +196,84 @@ export const parseWMTSContent = (content: string, endpoint: string) => {
...
+    parsed.Capabilities.Contents.TileMatrixSet
+  );
+  const tileMatrixById = tileMatrixSets
+    ? Object.fromEntries(
+        tileMatrixSets.map((tileMatrix) => [tileMatrix.id, tileMatrix])
+      )
+    : {};
+  console.log({
+    tileMatrixById: Object.values(tileMatrixById).map((x) => x.supportedCRS),
+  });
```
bprusinowski, line 209: 🔎

bprusinowski (approved):
Thanks @ptbrowne, LGTM! Also nice that the hot reload for Storybook works better now 🎉 

I have a few comments, some not related to this PR, but I've just noticed them, feel free to open new issues if it makes sense 👍 

- [x] The map attribution is only shown when we enable a custom layer, this is already the case on TEST / INT, but attribution is always there on PROD. I think we should only "enhance" it with new information when custom layer is added?
- [x] The legend loads correctly in dropdown popover, but not in chart (see screenshot below) 👀 
- [x] It looks that some layers that were previously working (e.g. 112 Alarm centers from wmts.geo.admin.ch) are still not working with this PR – maybe we'd need to take a look at this afterwards, just a note 👍 

Screenshot 2025-04-29 at 09 12 22

ptbrowne:
Thanks for the checks Bartosz, it's better now ;)

PR #2303, opened 2025-05-12
feat: Add units to limit legend items

### LINKED ISSUE #2301: Target Values + Limits: Add Unit to the Legend for target values
### Description
Add units to the legend items for target values. 

Example: https://int.visualize.admin.ch/de/v/Tv-3-kVhMEQ8
See how in the legend we have a range `17,79072-18,2589` but no unit `Mt` after the values. 

Important to Note: Adding the unit will result it in showing up in multiple areas on the chart. If multiple targets are selected it will be shown for each target.  The unit will appear: 
- Axis label
- For each Legend Item

### Additional Context 
- Change Request from Marco here: https://gitlab.ldbar.ch/bafu/visualize/-/issues/541#note_18688

bprusinowski:
Closes #2301

This PR adds units to limit legend items if they are available.

## How to test

1. Go to [this link](https://visualization-tool-git-feat-add-units-to-limit-lege-7d57f3-ixt1.vercel.app/en/v/Bkq9xccvieTR?dataSource=Int).
2. ✅ See that the legend limit contains a unit.

---

### DIFF
--- a/app/docs/charts.stories.tsx
+++ b/app/docs/charts.stories.tsx
@@ -91,7 +91,11 @@ const ColumnsStory = {
         <InteractiveFiltersChartProvider chartConfigKey={chartConfig.key}>
           <Observer>
             <ColumnChart
-              limits={{ axisDimension: undefined, limits: [] }}
+              limits={{
+                axisDimension: undefined,
+                limitMeasure: undefined,
+                limits: [],
+              }}
               observations={columnObservations}
               measures={columnMeasures}
               measuresById={keyBy(columnMeasures, (d: Measure) => d.id)}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ You can also check the
 
 # Unreleased
 
+- Features
+  - Limit values displayed in the legend now include the unit if it's available
 - Fixes
   - Default min custom scale domain value now defaults to 0 if no negative
     values are present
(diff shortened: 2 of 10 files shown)


app/docs/charts.stories.tsx
```diff
@@ -91,7 +91,11 @@ const ColumnsStory = {
         <InteractiveFiltersChartProvider chartConfigKey={chartConfig.key}>
           <Observer>
             <ColumnChart
-              limits={{ axisDimension: undefined, limits: [] }}
+              limits={{
+                axisDimension: undefined,
+                limitMeasure: undefined,
+                limits: [],
+              }}
```
ptbrowne, line 98: Seems like limits is optional, so we could just remove it ?

ptbrowne (approved):
LGTM :rocket:

app/docs/charts.stories.tsx
bprusinowski: Ah, they are required in the chart types 🥹 

code

Screenshot 2025-05-13 at 10 35 54

ptbrowne: OK, they could be made optional on the chartType right ? It's not a big problem, but it's a bit annoying :sweat\_smile: No problem if you have other things to do.

bprusinowski: It sounds like we still would need to work on the limits to support more types in the near future, I'll make a note to myself 👍

PR #2305, opened 2025-05-13
feat: Conversion units

### LINKED ISSUE #1875: Conversion units (BLW)
### Description
Add a feature to convert units in charts, such as from kg to tons, to facilitate diverse data representation.

### Examples
Charts where data is represented in different units, making it easier for audiences to understand values.

### Possible Approaches
Integrate a unit conversion toggle or dropdown in chart settings.
Automatically adjust axis labels and data values based on the selected units.

### Definition of Done
- [ ] Users can toggle between unit options, such as kg to tons, directly within chart settings.
- [ ] Unit selection automatically updates axis labels, legends, and data values to reflect the chosen unit.
- [ ] All conversions are validated to ensure accuracy and consistency in displayed values.
- [ ] Conversion options support multiple measurement types (e.g., weight, distance) where applicable.
- [ ] Unit conversion feature supports all languages and can be seamlessly applied to charts in German, French, Italian, and English.

bprusinowski:
Closes #1875

This PR adds a way to set custom measure units that override ones coming from the data as well as some general improvements to the application:
- compatible with all charts types except tables,
- settings persisting across chart type changes,
- reflected in chart axes, tooltips, data preview, limits and legends,
- **not reflected** in metadata panel and data download.

This feature is hidden behind a feature flag (`flag__convert-units=true`).

### Comments

🧨 With this PR it's becoming clear that we need to have a more centralized place to control various things related to overriding things, in this case, dimension units and calculations.

- Maybe it would make sense to add something like `getLabel`, `getValue` methods directly to `components` that would ensure we have a single place to set such overrides? Now we define `getValue` getters in every chart, separately – why not consolidate this?
- Currently we need to set these in many different places which is not scalable. This makes it challenging to remember about all the places that should be affected by such changes, like e.g. limits, legends or chart-type persisting logic.
- I would propose we explore options to optimize this (and not only this, all other custom set options, uncertainty definitions, chart type switches, and so on) in the near future.

Also see [this Notion document](https://www.notion.so/interactivethings/Technical-roadmap-25857051934f4c828d893bf1546d0451?pvs=4#1f430da543028052b62fe346285ccd7d).

## How to test

1. Go to [this link](https://visualization-tool-git-feat-conversion-units-ixt1.vercel.app/en/v/Is2dslMYM8Uf?dataSource=Prod&flag__convert-units=true).
2. ✅ See that the unit is overwritten (`Rp./WH` instead of `Rp./kWH`, with a multiplier of 0.001).
3. Copy and edit the visualization.
4. Scroll the left panel to find `Unit Conversion` section at the bottom.
5. ✅ See that it's possible to change the multiplication factor and labels and that chart (axis label, scaling) and tooltip adjust to the changes.
6. ✅ Switch to a column chart and see that your changes to units were persisted.
7. ✅ Add segmentation and turn on the 100% mode to see that it doesn't overflow the chart.
8. ✅ Open Table view and see that the conversion is also applied to relevant columns.
9. Go to [this link](https://visualization-tool-git-feat-conversion-units-ixt1.vercel.app/en/create/new?cube=https://communication.ld.admin.ch/ofcom/test_target/cube/target_ofcom/1&dataSource=Int&flag__convert-units=true).
10. Open Vertical axis and add a target value.
11. ✅ Enable unit conversion and see that the limit stays "in place", being adjusted to the conversion unit.
12. ✅ See that the custom unit label is also reflected in the legend.

---

- [x] I added configurator state migrations


app/config-types.ts
```diff
@@ -352,6 +354,30 @@ export type UnitConversionFieldExtension = t.TypeOf<
...
+
+      if (component && unitConversion) {
+        acc[componentId] = unitConversion;
+      }
+
+      return acc;
+    },
+    {} as Record<string, UnitConversionFieldExtension["unitConversion"]>
+  );
+};
```
ptbrowne, line 413: Could use a test ?

app/browse/data-table-preview.tsx
```diff
@@ -111,20 +120,19 @@ export const DataTablePreview = ({
...
                       component="td"
-                      sx={{
-                        textAlign: isNumericalMeasure(component)
-                          ? "right"
-                          : "left",
-                      }}
+                      sx={{ textAlign: numerical ? "right" : "left" }}
                     >
-                      {format(obs[component.id])}
+                      {format(numerical && cu && v ? +v * cu.factor : v)}
```
ptbrowne, line 135: I wonder why the conversion is done so late, at display time. Another approach would be to change the data themselves. This way we would not have to change the formatting everywhere. It could be seen as a job for the formatter here. The formatter would be created with conversionUnits in mind and would the conversion. Otherwise we have formatting inside the formatter and then here.

app/charts/combo/combo-line-single-state-props.ts
```diff
@@ -47,14 +47,17 @@ export const useComboLineSingleStateVariables = (
...
+
   const numericalYVariables: NumericalYComboLineSingleVariables = {
     y: {
       lines: chartConfig.fields.y.componentIds.map((id) => ({
         dimension: measuresById[id],
         id,
         label: measuresById[id].label,
         color: fields.color.colorMapping[id],
-        getY: (d) => (d[id] !== null ? Number(d[id]) : null),
+        getY: (d) =>
```
ptbrowne, line 59: I think we should use a function convertUnit(Number(d[id], y.unitConversion) here.

1. The state would not have to know about how to multiply and about this factot property
2. We would reduce duplication of this knowledge everywhere
3. We could have in the future unit conversion which is not linear (this happens for various unit pairs)

app/charts/combo/combo-line-single-state-props.ts
```diff
@@ -60,7 +60,9 @@ export const useComboLineSingleStateVariables = (
           d[id] !== null ? Number(d[id]) * unitConversionFactor : null,
         getMinY: (data) => {
           const minY =
-            min(data, (d) => (d[id] !== null ? Number(d[id]) : null)) ?? 0;
+            min(data, (d) =>
+              d[id] !== null ? Number(d[id]) * unitConversionFactor : null
+            ) ?? 0;
```
ptbrowne, line 65: Use a function convertUnit here

app/charts/index.ts
```diff
@@ -1960,6 +1960,22 @@ const chartConfigsAdjusters: ChartConfigsAdjusters = {
...
+            // Only keep the unit conversion if new measures have
+            // the same unit as the old one.
+            if (
+              oldValue &&
+              draft.fields.y.componentIds.includes(oldValue.componentId)
+            ) {
+              return produce(newChartConfig, (draft) => {
+                draft.fields.y.unitConversion = oldValue;
+              });
+            }
```
ptbrowne, line 1993: In the case of a scatterplot, we would need to handle X ?

app/charts/index.ts
```diff
@@ -2180,35 +2188,53 @@ const chartConfigsAdjusters: ChartConfigsAdjusters = {
...
           }
-        } else if (isBarConfig(oldChartConfig)) {
-          leftMeasure = getMeasure(oldChartConfig.fields.x.componentId);
-          columnUnitConversion = oldChartConfig.fields.x.unitConversion;
+          case "comboLineColumn":
+          case "table":
+            break;
+          default:
+            const _exhaustiveCheck: never = oldChartConfig;
+            return _exhaustiveCheck;
```
ptbrowne, line 2268: I am scanning here but I am quite impressed by the extent you had to check here.

I wonder now if the conversion of unit should not happen at the data layer = converting the observations directly and keeping the unitConversion config at chart level, indexed by component, instead of it being an UI thing that is bound to fields (which needs to be carried over when converting the chart). I think the approach would be simpler but maybe it has downfalls I do not see.

app/charts/map/map-state.tsx
```diff
@@ -414,14 +417,16 @@ const useNumericalColors = (
...
+      ? (colorSpec.unitConversion?.factor ?? 1)
+      : 1;
   const getValue = useCallback(
     (d: Observation) =>
-      componentId
-        ? d[componentId] !== null
-          ? Number(d[componentId])
-          : null
+      d[componentId] !== null
+        ? Number(d[componentId]) * conversionFactor
```
ptbrowne, line 427: convertUnit function could be used here so that we do not have to do the default to 1 of factor + knowledge of multiply.

ptbrowne (commented):
I reviewed and scan.

1. I am quite impressed by the amount of code that needed to be changed to implement this feature, I am quite impressed since I see this was started 2 days ago !
2. Technically, I think it would be worth to think a bit if we could simplify the approach by applying the conversion to the data instead of doing it late in the UI. I think it would vastly simplify since we would hold the conversion config per component IRI at root chart config (or cube config I guess), and we would not have to deal with "converting them" across chart type changes. Maybe it's not possible for other problems I do not foresee, but since this changes so much code and adds so much potential for errors, I think it's worthwhile thinking if this could be done. Even if the UI is done at "field" level, I think it could change something at the config - cube level.
3. A lot of times, the knowledge about the multiplier & conversion factor is spread out at UI level, I think it would be worthwhile to extract so that we can change more easily the logic in the future, for example if we implement a logFactor logic or something like this (conversion of decibels to energie for example ? [chatgpt gave me other examples of non linear relationships between units](https://chatgpt.com/share/6825cd2c-0a68-8003-a879-95df1a5f261f))
4. I think from a product perspectives that allowing to customise the labels so much is a bit risky. I could put "mumbo jumbo" in the unit labels with no problem. Historically, the visualize approach was to lock down those possibilities. I see that given the infinite unit possibility, it would be hard to have labels for all of them, but I could see it as Lindas job to provide the units with translations. I would guess that perhaps the database was not ready enough to implement this usecase and this is why we had to relax the constraint on our side, am I correct ?

bprusinowski:
Thanks for reviewing @ptbrowne, all very good points! After starting implementing combo and map charts, I also started to realize that it probably would be better to extract this up. I will take a look at this – hopefully it won't be a big refactor, as the UI would still stay the same.

For the point 4 about restricting this behavior, recently the client's vision for Visualize changed and front-end calculations and other things that were previously "forbidden" are now allowed.

For this particular feature, we didn't go with LINDAS-based implementation because there was a need for flexibility in defining the labels and scaling factors and the BLW data that this feature is initially made for doesn't have units as column metadata, but rather we have a "Unit" column, with e.g. "kg" as values, which would make it super tricky to implement correctly (as normally, we'd take a unit of a given measure and base conversion on it).

Hope it makes more sense now 😅

ptbrowne:
> hopefully it won't be a big refactor, as the UI would still stay the same.

Just to be sure we are on the same page: it still would be a big refactor as all the code related to chart adjustments would not be needed anymore, and also all the code in individual chart states to do the conversion would not be needed anymore, right ?

sosiology:
Thank you for preparing this @bprusinowski and documenting the challenges and updating the technical roadmap. Also thanks to @ptbrowne for the thorough review. 

I focused my testing on the BLW cubes with the following results: 

- Changing Amount in KG to Tons: https://visualization-tool-git-feat-conversion-units-ixt1.vercel.app/en/v/Pseye9_RWlEG
- Changing CHF to Rp.: https://visualization-tool-git-feat-conversion-units-ixt1.vercel.app/en/v/SxPuxJ00Nqqz?dataSource=Int

adintegra:
Thank you for this @bprusinowski and @ptbrowne  🎉 Tested as well. LGTM

bprusinowski:
@ptbrowne I simplified the approach by:
- creating `conversionUnitsByComponentId` at the root level of chart config, instead of on a field level,
- adding a way to transform results of components and observations queries on the client side, in useQuery hooks,
- transforming the component units and values and observations values in case we need to.

I also made it mandatory to pass chart config to every observations and components hook, which gives us access to its properties in case we need to add another similar feature (and preventing us from forgetting to pass it).

I think it's much better now, let me know what you think :)

app/graphql/hooks.ts
```diff
@@ -384,6 +437,47 @@ export const useDataCubesObservationsQuery = makeUseQuery<
...
+                }
+              }
+            );
+
+            return newObservation;
+          }),
+        },
+      },
+    };
+  },
```
ptbrowne: I wonder how is the data cached and when does the transform happens and what makes it trigger again ? Let's sync IRL ? In react-query, the conversion unit would be part of the uqeyr key but since here we use our own useQuery, I wonder how it works ?

app/graphql/hooks.ts
```diff
@@ -330,6 +330,24 @@ export const useDataCubesComponentsQuery = makeUseQuery<
...
+                  case "single":
+                    return {
+                      ...limit,
+                      value: limit.value * conversionUnit.multiplier,
+                    };
+                  case "range":
+                    return {
+                      ...limit,
+                      from: limit.from * conversionUnit.multiplier,
+                      to: limit.to * conversionUnit.multiplier,
```
ptbrowne: I'd use a helper here.

app/graphql/hooks.ts
```diff
@@ -300,80 +300,96 @@ export const useDataCubesComponentsQuery = makeUseQuery<
...
-        dataCubesComponents: {
-          ...data.data.dataCubesComponents,
-          measures: data.data.dataCubesComponents.measures.map((measure) => {
-            const conversionUnit = conversionUnitsByComponentId[measure.id];
-
-            if (!conversionUnit) {
-              return measure;
-            }
+/**
+ * Transforms the data from the data cubes components query.
```
ptbrowne, line 307: Maybe adds that it takes care of the convrersion unit in particular.

ptbrowne (commented):
LGTM 💯  the code looks much simpler 🎉  I just wonder about when the transform function is re-run ? Is it everytime the chartConfig changes ?

ptbrowne (approved):
Cool that we only transform when conversionUnits has changed 👍 

Concerning the extractoin of the util, I had meant that the multiplying itself would be extracted, not the overall function.

```
const convertValue = (x: Number, conversion: ConversionUnit) => {
  // then in the future, if we have other way to convert, for example, log multiply
  // or something, then only this place change
  return x * conversion.multiplier
}
```

It's not very important for this PR as this can be done later, I just wanted to clarify what I wanted to say ;)

Issue #2293, opened 2025-04-30
map: Parsing of CRS i.e. in WMTS Layers


Rdataflow:
**Describe the bug**
WMTS Capabilities Response having `ows:SupportedCRS` doesnt' reflect WMTS Implementation SPEC normative examples [1] 

**To Reproduce** 
1. include any WMTS layer from [2]
2. See no layer appears and console.log messages on CRS not supported

**Actual behavior**
false CRS interpretation, reporting CRS as not supported, where it actually would be supported

**Expected behavior**
proper parsing of urn notation i.e. also consider `urn:ogc:def:crs:EPSG::3857` or `urn:ogc:def:crs:EPSG:6.6:3857` or whichever {version} `urn:ogc:def:crs:EPSG:{version}:3857` i.e. using a pattern of 
```javascript
  (crs) => crs.match(/EPSG:(.*:)?(3857)$/)
```

**Environment (please complete the following information):**
- Visualize environment and version: INT 5.7.5

**Additional context**
[1] [OGC WMTS Simple Profile](https://docs.ogc.org/is/13-082r2/13-082r2.html#28)
[2] https://geoserver.swissdatacube.org/geoserver/gwc/service/wmts?REQUEST=GetCapabilities
[3] see visualize parsing WRT https://github.com/visualize-admin/visualization-tool/blob/00d3bc70e33fcf00df12fc125977ee16d81b7c7d/app/charts/map/wms-wmts-endpoint-utils.ts#L167-L169
[4] vs battle proofed OL implementation example https://github.com/openlayers/openlayers/blob/65602233a67f2fe9ee6d08626be3491a2e2c9170/src/ol/proj/projections.js#L17-L28

**nb**
did you ever spot `CRS:84` - if so - where? (AFAICT this would differ from `EPSG:3857` and have coordinates much like CRS `EPSG:4326` but with inverse axis order...) - therefore, it is safe to skip `CRS:84`...

@ptbrowne cc @adintegra @nrohrbach

Rdataflow:
@bprusinowski pinging you as the author as well 😉 

BTW: there is no real purpose in listing unsupported layers - if you agree, can we filter those? (at least while there is no support for it?)

Rdataflow:
[...]

PR #2306, opened 2025-05-13
WMTS fixes

ptbrowne:
### legend
- **refactor: Extract fetch legend image**
- **feat: Cache individual legend images**
- **feat: Constrain legend image size**

### crs parsing + getTile

We now correctly parse CRS even in "urn" format.
The getTile operation is also supported for layers that do not expose
directly a resourceURL but rely on the GetTile operation published
on the root of the document.

Closes https://github.com/visualize-admin/visualization-tool/issues/2293

- [x] Parse CRS correctly
- [x] Support GetTile for geoserver.swissdatacube to work
- [x] Remove unsupported providers

## How to test

1. Go to this [link](https://visualization-tool-git-fix-wmts-ixt1.vercel.app/en/create/new?cube=https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/10&dataSource=Prod).
2. Create a map chart, go to Base Layer
3. Click Add Layer, try providers
  * https://wms.geo.admin.ch/
  * https://geoserver.swissdatacube.org/geoserver/gwc/service/wmts?REQUEST=GetCapabilities

---


app/charts/map/wmts-utils.ts
```diff
@@ -188,6 +210,10 @@ const mapArrayOrUnique = <T, I>(arr: T | T[], cb: (item: T) => I): I[] => {
   return [cb(arr)];
 };
 
+const formatGetTileUrl = (url: string) => {
+  return `${url.endsWith("?") ? url : `${url}?`}Service=WMTS&Request=GetTile&Transparent=true&Version=1.0.0&Format=image/png&tileMatrixSet={TileMatrixSet}&tileMatrix={TileMatrix}&tileRow={TileRow}&tileCol={TileCol}&layer={Layer}`;
```
Rdataflow, line 214: @ptbrowne nit: Format could differ I.e. JPEG etc.

ptbrowne: Thanks @Rdataflow I was aware that this could be the case, but I am wondering if it's necessary to support this since mostly we would want ability to overlay images and then we need to have the transparency. Do you know usecases where the WMTS server only outputs jpeg files ? From what I have seen on the geoserver.swissdatacubes I was able to ask for both jpeg and png.

Rdataflow: @ptbrowne while you may prefer png in cases both are available. OTOH many WMTS only offer one single format, i.e. https://wmts.geo.admin.ch/EPSG/3857/1.0.0/WMTSCapabilities.xml offers jpeg for many layers (even if that one is not KVP, it still shows the clear trend: jpeg occurs frequently). It's a well defined param in the spec.
see also [WMTS GetTile request.xsd](https://schemas.opengis.net/wmts/1.0/wmtsGetTile_request.xsd)
image
and [OpenGIS Web Map Tile Service Implementation Standard](https://portal.ogc.org/files/?artifact_id=35326) p34
image
HTH for OCG conformance 👍

ptbrowne: I think we can tackle that in another issue, I will not have time to tackle it in this issue.

Rdataflow (commented):
@ptbrowne the canton SG responded upon our feedback and enabled CORS for *.admin.ch
test: https://test.visualize.admin.ch/v/jOU4D48I6HoX?dataSource=Prod 👍 😄  

nb: OTOH they cannot enable EPSG:3857 for WMTS - but at least for SG WMS it works now

app/charts/map/initial-wms-wmts-providers.json
```diff
@@ -0,0 +1,349 @@
+[
```
Rdataflow, line 1: @ptbrowne after SG reconfigured their services I now updated the WMTS-WMS-Monitor to test `Origin: https://visualize.admin.ch`
you may simply reuse the resulting list of endpoints passing CORS and EPSG:3857 check... 
https://rdataflow.github.io/WMTS-WMS-monitor/result.json 😄

Rdataflow (commented):
@ptbrowne 
you may reuse results from https://rdataflow.github.io/WMTS-WMS-monitor/result.json

app/charts/map/wms-wmts-endpoint-utils.ts
```diff
@@ -158,7 +158,7 @@ const indexByKey = ({
  * @see https://github.com/visgl/deck.gl/discussions/6885#discussioncomment-2703052
  */
 export const isCRSSupported = (crs: string | undefined) => {
-  return crs && crs.includes("EPSG:3857");
+  return crs && (crs.includes("EPSG:3857") || crs.includes("EPSG:4326"));
```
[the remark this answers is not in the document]
ptbrowne: I did a bit a investigation seeing that apparently EPSG:4326 was working for the [https://bio.discomap.eea.europa.eu](https://bio.discomap.eea.europa.eu) endpoint. This is because the server is WMS and there is handling of EPSG:4326 for the DeckGL WMSLayer (https://github.com/visgl/deck.gl/blob/9.1-release/modules/geo-layers/src/wms-layer/wms-layer.ts).

The endpoint your refer to is WMTS, this is why it's not wroking.

Since the task is about EPSG:3857, I am removing the 4326 from supported endpoints, it would be tackled in another task if needed.

ptbrowne:
@bprusinowski I moved the "encoding" up in the UI, I think it will be less confusing for the user, I can easily removed the commit if needed.

app/charts/map/initial-wms-wmts-providers.json
ptbrowne: Thanks, I now this JSON in the script updating our providers.

app/charts/map/wmts-utils.ts
ptbrowne: I created the issue here: #2312

app/charts/map/map-custom-layers-legend.tsx
```diff
@@ -115,6 +120,21 @@ const fetchLegendImage = async (
...
+const fetchLegendImageFromCache = async (
+  remoteLayer: RemoteWMSLayer | RemoteWMTSLayer
+) => {
+  if (!remoteLayer?.legendUrl) {
+    return undefined;
+  }
+  const cached = cachedLegendImages[remoteLayer.legendUrl];
+  if (cached) {
+    return cached;
+  }
```
bprusinowski, line 163: 👍

scripts/wmts-wms-providers/main.ts
```diff
@@ -0,0 +1,93 @@
...
+    __dirname,
+    "../../",
+    "./app/charts/map/initial-wms-wmts-providers.json"
+  );
+
+  const providersPath = path.join(
+    __dirname,
+    "../../",
+    "./app/charts/map/wms-wmts-providers.json"
+  ); // Path to your JSON file
```
bprusinowski, line 50: 🤖 🧹 :)

app/charts/chart-config-ui-options.ts
```diff
@@ -1265,6 +1269,13 @@ const chartConfigOptionsUISpec: ChartSpecs = {
           },
         },
       },
+      {
+        idAttributes: ["componentId"],
```
bprusinowski, line 1273: @ptbrowne I think we can use an empty array here and also change `baseLayer` to not have this – as far as I am aware, we do not have a `componentId` property in base layer and custom layers, I think it was wrong before 😱

bprusinowski (approved):
Nice, thanks @ptbrowne 👏 I think it's a good idea to separate base and custom layers 👍 

There's only one functional thing I noticed when testing – if I go to wms.geo.admin.ch endpoint and add e.g. `112 Alarm centers` layer and then switch to wmts.geo.admin.ch, the `112 Alarm centers` layer is selected, even though it's a new layer that comes from a different endpoint, so I would expect to be able to select it. Is this behavior intended? 🤔

ptbrowne:
Thanks @bprusinowski I was able to fix the issue.

Rdataflow:
@bprusinowski I just got notified from Europe Environment Agency, the joined EPSG:3857 as well 🎉 
can you refetch the providers list?

bprusinowski:
Looking forward! FYI, the update should be included in 5.8.2 (https://github.com/visualize-admin/visualization-tool/commit/f7ef93f944a111b756ac10abd5507d3ce945a3af), I didn't open a new PR for it 👍

PR #2336, opened 2025-05-21
feat: Time range limits

### LINKED ISSUE #2311: Target Values: Possibility to add ranges on x-axis
### Description 
In visualize the targets as shown below.

Image

Currently, we do not support ranges on the x-axis as a data set was missing. We now have a test data set here: https://s.zazuko.com/24VcfQW

In a first step, we should have a look at the data set and evaluate the effort +/- to add this change. 

**Additional context** 
Gitlab comments: https://gitlab.ldbar.ch/bafu/visualize/-/issues/541#note_18744

bprusinowski:
Closes #2311

This PR:
- fixes the tandem selection / un-selection issue for some edge cases,
- adds support for symbol limits to bar charts,
- adds support for time range limits.

## How to test

⚠️ The limits rendering and fetching has been re-worked and it would be good to also test with other limit datasets, to make sure nothing broke.

**Tandem toggle / un-toggle issue**

Not possible to test, as the issue was discovered when enabling the time range limits for a cube that without them, had only one limit (not possible to provide "broken case" link).

**Symbol limits for bar charts**

Not possible to test with the current cubes – but once a compatible one is added, they should work.

**Time range limits**

1. Go to [this link](https://visualization-tool-git-feat-range-limits-ixt1.vercel.app/en/create/new?cube=https://environment.ld.admin.ch/foen/test_target/cube/target_future_ghg/1&dataSource=Int).
2. Switch to a line chart.
3. Open `Vertical Axis` field.
4. Enable the first limit (time range limit).
5. ✅ See that it was rendered and correctly spans the 2021-2030 period.
6. ✅ See that you can make the line dashed and change its color.

---


sosiology (commented):
Thanks for preparing this @bprusinowski 

- Time range was rendered and correctly spans the 2021-2030 period. ✅
- Was able to change the symbols, line style and color 😄 (see example below)
- Did some quick tests with the data set `: Soil pollution by heavy metals (Cube for limits)`  (also with y-axis scaling and conversion units). So far I haven't notice any obvious bugs. 
- We might want to consider not showing the target limit menu and toggles when a chart type is selected that is not supported but that is highly dependent on the technical feasibility and complexity and would be an enhancement imo. 





Example Chart: https://visualization-tool-git-feat-range-limits-ixt1.vercel.app/en/create/l9gXuyAJrNKn?dataSource=Int
image

PR #2344, opened 2025-05-26
maintenance: Update Node.js to v22

### LINKED ISSUE #1958: Update Node.js to v22
Thanks to https://github.com/visualize-admin/visualization-tool/pull/1953, we are now using Node.js 20 in the application (bumped up from 18).

To not immediately jump a few major versions ahead, we should test and make sure the application works correctly after the update to 20; if this is the case, we should consider bumping the Node.js version again, to the next major version supported until April 2027.

bprusinowski:
Closes #1958

---

### DIFF
--- a/.github/workflows/base-checks.yml
+++ b/.github/workflows/base-checks.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v4
         with:
-          node-version: 20.x
+          node-version: 22.x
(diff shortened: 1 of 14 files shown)


bprusinowski:
cc @adintegra – I ran the application locally + built a Docker image and didn't observe anything being broken. I also took a look at Node.js 22 changelog, and I think it should be safe to deploy. Let me know what you think :)

PR #2359, opened 2025-05-28
fix: Drop global unique name constrain, scope to user

### LINKED ISSUE #2358: Color Palette Names are unique across the application
## Description 
If a color palette name already exists (even if it's from another user), the palette doesn't get saved (e.g., "BAKOM 1234").

Reported by Nick in Gitlab here: https://gitlab.ldbar.ch/bafu/visualize/-/issues/377#note_18850

**To Reproduce** Steps to reproduce the behavior:

1. Go to test.visualize.admin.ch
2. Login
3. Go to Color Palette
4. Create a color palette with the name `Emma's Color Palette` 
5. See that color palette cannot be saved as the name is already used by another user (me)

**Expected behavior** 
Color palette names should be unique to a user not across the application (unless of course in future this could allow color palettes to be be shared across applications). 

- Visualize environment and version: [v5.8.1 (01495c7)](https://github.com/visualize-admin/visualization-tool/commit/01495c77dabb9f4e1019813a2cb9e9cf3b4d6d57)
- Browser and version [e.g. Chrome 132]

bprusinowski:
Closes #2358

This PR makes sure we only enforce uniqueness of color palettes on a user level, not globally.

---

- [ ] I made a self-review of my own code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ You can also check the
   - Invalid URL error message is now translated
   - Improved UX of interacting with conversion units multiplier input
 - Fixes
+  - Custom color palette names are not required to be globally unique anymore,
+    but unique per user
   - Fixed an infinite render loop when adding a custom map layer
 - Styles
   - Fixed some smaller UI inconsistencies, mostly in table charts
--- a/app/prisma/schema.prisma
+++ b/app/prisma/schema.prisma
@@ -69,7 +69,7 @@ enum PALETTE_TYPE {
 
 model Palette {
   paletteId  String       @id @default(uuid())
-  name       String       @unique @db.VarChar(100)
+  name       String       @db.VarChar(100)
   type       PALETTE_TYPE  
   colors     String[]
   created_at DateTime     @default(now()) @db.Timestamp(6)
@@ -79,4 +79,5 @@ model Palette {
   user_id    Int?
 
   @@map("palettes")
+  @@unique([name, user_id])
 }


app/server/user-controller.ts
```diff
@@ -29,16 +36,39 @@ export const UserController = controller({
     const session = await getServerSession(req, res, nextAuthOptions);
     const userId = session?.user?.id;
 
-    return await getPalettesForUser(userId);
+    if (!userId) {
```
noahonyejese, line 39: nit: Could use assertion here

noahonyejese (approved):
LGTM overall, there still seems to be an issue on Vercel deployment but I'll approve for now. Can we now delete a user account or was this already the case in the past? 😃

bprusinowski:
Thanks @noahonyejese 🙇‍♂️ Yeah, in order to merge this PR, we'd actually first need to tackle https://github.com/visualize-admin/visualization-tool/issues/2360 to improve the way we handle migrations (and to fix Vercel deployment).

We can't remove accounts yet, but just to be prepared for the future 😄

adintegra:
Fundamentally, these changes look good to me. However, the client has recently decided that unique palette names (globally) is useful to them. It may therefore make sense to drop the uniqueness constraint changes here but keep the other enhancements, i.e.

> - user needs to be logged in to create / update / remove / get a color palette (was already a constraint in the UI, now also added to the backend side),
> - we remove user palettes on User deletion (cascading delete).

For added context, there are ongoing conversations outside this thread regarding the overall streamlining of the DB design, enhancing it to accommodate palette sharing and potentially moving to the proper use of prisma migrations (i.e. moving from `prisma db push` to `prisma migrate dev` + `prisma migrate deploy`) – see #2360.

PR #2399, opened 2025-07-23
fix: Always show chart tab scroll buttons when needed

### LINKED ISSUE #1776: Tab Label Improvements
Arrow Missing on Left Hand Side
1. In published mode, the left arrow seems to be missing, and the tab is not aligned with the rest of the frame. I assume that the arrow should only appear once the user scrolls, but the indent before the first tab looks a bit strange to me. see: https://test.visualize.admin.ch/en/v/tFKBIr7nwSRG?dataSource=Prod and screen shot below
Image

2. Add Character Limit to Tabs to avoid really long labels. 
I noticed there is no character limit on the tab label...should we consider introducing a character limit for the tab labels to avoid extremely long labels as can be seen here: https://test.visualize.admin.ch/en/v/tFKBIr7nwSRG?dataSource=Prod. If the label is very long, it may be hard for the user to discover additional charts in a chart published in tab layout. This would be a new issue 💡 

_Originally posted by @sosiology in https://github.com/visualize-admin/visualization-tool/issues/1714#issuecomment-2379587575_

bprusinowski:
Closes #1776

This PR makes sure we always show scroll arrows when needed to avoid empty spaces.

## How to test

1. Go to [this link](https://visualization-tool-git-feat-always-show-scroll-butt-ad5f82-ixt1.vercel.app/en/v/WUo-ZwN8m6WW?dataSource=Prod).
2. ✅ See that the left arrow is visible without having to scroll.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ You can also check the
     filters section
   - Dataset browse view is now correctly displayed on mobile devices
   - Database-related actions are now hidden in preview mode (copy URL, share)
+- Styles
+  - Scroll buttons in the chart selection tabs are now always shown when needed
 - Security
   - Added additional protection against data source URL injection
   - Removed feature flag for custom GraphQL endpoint
--- a/app/components/chart-selection-tabs.tsx
+++ b/app/components/chart-selection-tabs.tsx
@@ -456,13 +456,11 @@ const TabsInner = ({
                     style={{ cursor: disabled ? "auto" : "pointer" }}
                     variant="text"
                   >
-                    {disabled ? null : (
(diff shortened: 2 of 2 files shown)


sosiology:
@bprusinowski this LGTM thanks! I think we should create a new issue for the 2nd point (the long tab titles). This is not something i think we need to solve now, but maybe in future across the whole application (text blocks, titles, descriptions), seeing as we now also introduced a character limit for the annotations. 
example chart: https://visualization-tool-git-feat-always-show-scroll-butt-ad5f82-ixt1.vercel.app/en/v/cOblylIPvXmw?dataSource=Prod

PR #2486, opened 2025-11-04
feat: Links in table charts

bprusinowski:
Closes #

This PR introduces links to table charts that can take a base URL and given dimension to dynamically construct a per-row, per-column link.

## How to test

1. Go to [this link](https://visualization-tool-git-feat-links-in-tables-ixt1.vercel.app/en/create/new?cube=https://energy.ld.admin.ch/sfoe/bfe_ogd84_einmalverguetung_fuer_photovoltaikanlagen/10&dataSource=Prod).
2. Switch to a table chart.
3. ✅ See that "Links" section is available at the bottom of the left panel.
4. ✅ Enable links, set base url and select a component and see that the links appear in the selected column.

---

- [x] I wrote configurator and chart config migrations (if applicable)

### DIFF
--- a/app/charts/table/linked-cell-wrapper.tsx
+++ b/app/charts/table/linked-cell-wrapper.tsx
@@ -15,7 +15,9 @@ const useStyles = makeStyles((theme: Theme) => ({
     alignItems: "center",
     gap: theme.spacing(1),
     color: "inherit",
+    fontWeight: "inherit",
     textDecoration: "none",
+
     "&:hover": {
       textDecoration: "underline",
     },


adintegra:
Nice, @bprusinowski 🎉 Maybe the field labelling is still a bit jargony for non-technical users. Could we consider something along the lines of "Link column" for Target dimension, "Link argument" for Source dimension? Also, I still like the links at the end of the rows (the initial behaviour) and think this could still be valuable for some users. We might want to consider keeping this functionality as an additional option in the dropdown. What do you think?

cc @sosiology

PR #2540, opened 2026-03-24
Set Maptiler key at runtime instead of build time

ludovicm67:
Currently, the Maptiler key was injected at build time.
In order to set a custom key per environment, we would need to be able to inject the key at runtime instead.
This PR allows defining the key at runtime.

---


hupf (approved):

ludovicm67:
I just rebased on `main`, so that the branch is up-to-date.

@hupf could you check if everything still looks good for you?
If so, we can merge it.

I'll already configure `TEST` with the value for `MAPTILER_API_KEY`.

PR #2541, opened 2026-03-24
Enable Sentry back

ludovicm67:
This enables Sentry back.

Work in progress: the best would be to upgrade Next.js first, then continue the work here.
As else we would need to do the work twice.

---

- [ ] I made a self-review of my own code


ludovicm67:
Closing this in favor of #2550.

PR #2545, opened 2026-04-13
chore: use agents.md instead of claude.md

mburri:
The file claude.md contains instructions specifically for the AI assistant Claude, but contributors might use other tools. 

Additionally, new [studies](https://arxiv.org/abs/2602.11988) have shown that extensive context files may actually lower the code quality produced by coding AIs while increasing the token/ credit usage.

This PR contains:
* [x] renamed CLAUDE.md to AGENTS.md
* [x] replace content with suggestions from: https://github.com/drona23/claude-token-efficient
* [x] updated changelog


hupf (changes requested):
I agree that the AGENTS file should not be too big to not clutter the context. But there is some interesting information in the previous CLAUDE.md that should be part of the README at least.

hupf (approved):

PR #2547, opened 2026-04-14
chore: upgrade dependencies with critical and high vulnerabilities

mburri:
This PR contains a first batch of dependency lifecycles with a focus on critical vulnerablities.

## Upgrades include:
* nextjs -> 15
* typescript -> 5.7
* babel-runtime -> 7.26
    * this one is used for lingui, see also  #2548 
* faster-xml-parser
    * upgrade of this lib forced changes to some test files
* madge -> 8
     * madge is used to detect circular dependencies
     
## Other changes:
* the version of `rollup` had some vulnerabilites - instead of upgrading it I decided to rewrite the task where rollup was used to use `esbuild`. This decision was made, because the existing process to build the embed.js file was also dependent on the babel.config.js file. If we want to switch to swc and turbopack (default with next 16) we will have to get rid of this anyway.
* the `npm publish` task used the `@preconstruct/cli` dependency and also used the babel config file. `@preconstruct/cli` had lots of high and critical issues. Since there seems to be little value to publish the app as an npm package, this task was removed without a replacement.

## Fixing critical vulnerabilites using package resolutions

After the first few upgrades it became apparent that the critical vulnerabilites could not be addressed in this way, because most of them were included in the project as transitive dependencies of multiple packages. It would have just taken too long for this first iteration. With assistance of a coding AI, it was decided to enforce patched versions of the affected packages by the `resolutions` property in `package.json`:

```
"resolutions": {
    "@babel/core": "^7.26.0",
    "@babel/parser": "^7.26.0",
    "@babel/traverse": "^7.23.2",
    "minimist": "^1.2.6",
    "pbkdf2": "^3.1.3",
    "sha.js": "^2.4.12",
    "cipher-base": "^1.0.5",
    "handlebars": "^4.7.9",
    "elliptic": "^6.6.1",
    "form-data": "^4.0.4",
    "fast-xml-parser": "^5.5.12"
  },
  ```
  
  This fixed all critical vulnerabilites in one go while not breaking performance or e2e tests.


ludovicm67 (approved):
Looks good for me.

We just need to make sure to not forget to remove the "resolution" part in a later step.

mburri:
Does not work anymore after rebase on main 
🕵

PR #2550, opened 2026-04-21
feat: enable Sentry back

ludovicm67:
This enables Sentry back, using the v7 of `@sentry/nextjs`, like it was done previously on 71c128595db921d7eeae015b06a823a637b3ddd7.
Many of the changes were ported back from that commit as well.

I added a way to configure CSP on the fly in case it doesn't point to sentry.io ; this could be done using `NEXT_PUBLIC_SENTRY_CSP`.

#2541 was an attempt using the v10 of the module, but our Next.js version is too old.
I'm closing #2541 in favor of this PR, as the upgrades would be done with the rest in future PRs.

---


hupf (approved):
LGTM 👍

PR #2557, opened 2026-05-07
fix: use photovoltaikanlagen/10 instead of photovoltaikanlagen/9 and …

mburri:
…regenerate grapqhl performance test workflows

Performance tests failed on main because a change that was already applied to the performance test workflow for PRs was not applied.

This PR changes this


mburri:
I bypassed the rules (merged without review) because we will only see if its ok when the pipeline runs on main

PR #2560, opened 2026-05-12
Add i18n table links

hupf:
Closes these issues:
https://gitlab.lindas.admin.ch/bafu/bafu-visualize/visualization-tool/-/issues/106
https://gitlab.lindas.admin.ch/bafu/bafu-visualize/visualize/-/issues/765

This PR extends the links attributes of table diagrams such that a language-specific base URL can be defined (currently it is only possible to define a single base URL across all languages). As implemented, there is no fallback mechanism between languages; if a language has no base URL defined, the column won't be linked. Previous chart configs with a base URL defined, will be migrated, such that this URL will be used for each language.

## How to test

- Prior to this change, create a chart diagram for the "Federal Councillors, Party, Canton, Terms" cube, enable the "Enable links" option and define a "Base URL" then choose the "Name" column as source and target column. The name column in the table should now have a link.
- Then with this change, open the previously created chart. The name column should still have the link (also if you switch the language).
- Additionally, with the change, create a new chart and define per language base URLs. The name column should have the language-specific link, depending on the currently active language.

---

- [x] I wrote configurator and chart config migrations (if applicable)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ You can also check the
   - Add SPARQL endpoints in the OpenTelemetry traces
   - Add Sentry integration back
   - Action tiles replaced with a link to the service desk
+  - Allow the definition of language-specific links in table diagrams
 - Refactoring
   - Incorporate the Swiss-Federal-CI library into this repository.
 - Fixes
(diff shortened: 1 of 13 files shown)


mburri (approved):
🎉

PR #2562, opened 2026-05-14
build(deps): update node-fetch

tpluscode:
This updates and removes some packages to get rid of vulnerable versions of node-fetch

`node-fetch` was not a direct dependency but pulled by a number of packages. Most notably, I updated the once I maintain myself: `@tpluscode/sparql-builder`, `@zazuko/cube-hierarchy-query`, `rdf-cube-view-query`, `sparql-http-client` but more troublesome were the graphql packages `@graphql-tools/*` and `@graphql-codegen/*`. Especially the latter changed the generated types somewhat and I had to apply numerous updates to imports and other usages.

I also got rid of `isomorphic-fetch` and `isomorphic-unfetch` which appeared unnecessary since fetch is native in recent version of node.

---


tpluscode:
@mburri The CI still fails but locally the build script works 🤷

mburri (approved):
I trggered the build by adding to the changelog  - and then deployment on vercel was successfull. That's something that I observed multiple times now after dependency changes: the first deployment will fail - I don't now why.

Afterwards one of the e2e tests failed - and again, rerunning the e2e tests fixed it.

One thing maybe bothers me: the PR title suggests that you updated node-fetch - but I don't see this from any changes

tpluscode:
Good to merge @mburri?
FYI, I added some more detail to the changelog entry

PR #2565, opened 2026-05-18
Improve dimension values query

BenjaminHofstetter:
This PR makes the Dimension Values query faster. It’s now more flexible, starting with subqueries based on how many and what kind of filters you use. We’ve added a new abstract filter class and two more specific Single Filter and Multi filter classes. For each filters there a 4 kind of filters possible (literal, shared, versioned literal, versioned shared).

## How to test

1. Test with dataset "Tierversuche: Tierversuchsstatistik" 

- [ x] I added a CHANGELOG entry
- [ x] I made a self-review of my own code


app/rdf/query/dimension-values/readme.md
mburri, line 1: Do we need this file?

app/rdf/query/dimension-values/model/cube-dimension-filter.ts
```diff
@@ -0,0 +1,163 @@
...
+    toSparqlFilter(observationVar: string, valueVar: string): string {
+        const isLiteral = this.isLiteralFilter();
+        const isVersioned = this.isVersionedDimension();
+
+
+        if (isLiteral && isVersioned) {
+            return `
+            ?${valueVar} schema:sameAs ?${valueVar}_unversioned  .
+            ?${observationVar} <${this.dimension.path?.value}> ?${valueVar} .
+            FILTER( STR(?${valueVar}_unversioned ) = "${this._value}" )`;
```
mburri, line 67: I think this should be formatted differently. 
Please run prettier to format changed files.
Have you setup prettier in your editor?

We will setup a git commit hook and a pipeline validation check to ensure consistent formatting on the whole project asap.

mburri (changes requested):
Seems to have fixed the problem.

Please fix the failing e2e and unit test, the knip task and it would be nice if you could reformat the changed files

app/rdf/query/dimension-values/readme.md
BenjaminHofstetter: removed

app/rdf/query/dimension-values/model/cube-dimension-filter.ts
BenjaminHofstetter: done

mburri (approved):
looks good now, fix the e2e test and then merge it!

PR #2567, opened 2026-05-19
build(deps): upgrade nth-check

tpluscode:
I updated `@svgr/cli` and `@storybook/builder-webpack5` in order to remove vulnerable versions of `nth-check`

It appears to have bumped some babel too. Let's see...

---

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -195,7 +195,7 @@
     "@mdx-js/loader": "^1.6.22",
     "@playwright-testing-library/test": "^4.5.0",
     "@playwright/test": "^1.49.1",
-    "@svgr/cli": "^5.5.0",
+    "@svgr/cli": "^7",
     "@types/autosuggest-highlight": "^3.2.0",
     "@types/clownface": "^2.0.7",
     "@types/cors": "^2.8.8",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ You can also check the
     findings of studies about the impact of context files for coding AIs
   - Remove LINDASOld endpoint configurations
   - Remove dependencies which used old versions of `node-fetch`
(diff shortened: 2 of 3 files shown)


tpluscode:
Again, locally builds and runs just fine

app/package.json
```diff
@@ -195,7 +195,7 @@
     "@mdx-js/loader": "^1.6.22",
     "@playwright-testing-library/test": "^4.5.0",
     "@playwright/test": "^1.49.1",
-    "@svgr/cli": "^5.5.0",
+    "@svgr/cli": "^7",
```
hupf, line 198: Just a general note, not something to be done here: I think in the future we should pin all versions to fixed numbers, this would be safer from a security perspective.

hupf (approved):
I've also regenerated the icons with the new rsvg version with `yarn icons` and it looks alright 👍

tpluscode:
> I've also regenerated the icons with the new rsvg version with `yarn icons` and it looks alright 👍

Thanks, I missed that

app/package.json
tpluscode: You mean like this?

```diff
-    "@svgr/cli": "^7",
+    "@svgr/cli": "7.0.0",
```

I don't see how pinning is safer if we cannot update? Are you worried about hacked package versions with malicious code?

hupf: → https://docs.renovatebot.com/dependency-pinning/

hupf: But you're right, basically with the lockfile there aren't any "automatic updates" when doing a `yarn install`

tpluscode: We are not using renovate anyway, right? I guess dependabot also handles pinned versions when doing security updates. Im fine if you will prefer pinning as a rule

PR #2568, opened 2026-05-19
build(deps): updated prismjs

tpluscode:
I'm actually surprised this did not come as an automatic dependabot PR. Does it even deserve a changelog entry?

---


hupf (approved):
Can you try to push it again such that the Vercel deployment & E2E tests run through?

tpluscode:
Uh, why is that one test failing 🙄

PR #2571, opened 2026-05-19
build(deps): updated storybook packages

tpluscode:
Soo, I started off by checking what uses the vulnerable `ip` package and it was only used by Storybook.

I updated the whole thing and `ip` is gone and so are many more duplicate dependencies. This may knock off some additional vulnerabilities too 🤞 

---


mburri:
@tpluscode 
Shouldn't there be a change in package.json that reflects the storybook upgrade?

mburri (approved):

tpluscode:
> @tpluscode Shouldn't there be a change in package.json that reflects the storybook upgrade?

Maybe not strictly necessary but I can add it.

tpluscode:
Ah, actually quite required, because IXT had previously pinned their storybook. So indeed, I missed the fact that I have not really updated all that I could have

PR #2574, opened 2026-05-20
build(deps): updated playwright

tpluscode:
---


hupf (changes requested):
The E2E tests are failing. There seems to be a breaking change...

hupf (approved):

PR #2593, opened 2026-07-01
fix: csp headers for index pages

mburri:
No CSP Header was set for index pages, e.g. https://visualize.admin.ch/en 

Pages with more than one path segment included the correct headers: https://visualize.admin.ch/en/imprint

The cause was that the matcher in middleware.ts was is recompiled for i18n routing during the build. The newly generated matcher did only match routes with at least two path segments, e.g. `/en/imprint`, but not `/en` itself.

It is not known to me whether upgrading next.js would fix this too.

---

- [ x ] I added a CHANGELOG entry
- [ x ] I made a self-review of my own code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ You can also check the
 
 ## Unreleased
 
+- Fixes
+  - Correctly set CSP headers for index pages
+
 ## 6.5.0 – 2026-06-01
 
 - Features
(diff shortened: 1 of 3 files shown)


hupf (approved):
Looks good. As the Next.js + i18n combo is causing this behavior, it may be useful to check the header in an E2E test, to catch when the behavior changes again...

PR #2599, opened 2026-08-03
feat: embededded browse page to create new visualizations

mburri:
The embedded visualisation on OpenData.swiss stopped working after we implemented more restrictive CSP policies.
Example page: https://opendata.swiss/de/dataset/programmstruktur-von-radioprogrammen-der-srg.

An analysis showed that opendata.swiss pages embed '/browse' pages. This was not anticipated when the stricter CSP policies were implemented. We expected only published visualisations to be embedded on third-party sites.
It also turned out that this was intentional and part of a broader feature designed some time ago. Instead of including the whole application, only a reduced 'browse' page should have been included using the URL parameter 'odsiframe=true'.
This would render a page with the basic dataset and a 'Create a Visualisation' button that opens the application in a new tab. Unfortunately, this change was not completely implemented, and will probably not be added to opendata.swiss soon.

This PR now addresses these issues.
- The CSP header for /browse now allows *.admin.ch and *.opendata.swiss as frame-ancestors. This enables these pages to be embedded in an iframe.
- The /browse page now detects whether it is embedded in an iframe and uses the reduced layout if necessary. 

This implements the `odsiframe` parameter without setting the parameter in the URL.
The drawback is that this only works on the client, whereas the 'odsiframe' parameter would also work on the server. As the page is rendered on the server, switching from the complete to the reduced layout only occurs after the page has been sent to the client, resulting in brief visible flickering.

### DIFF
--- a/app/pages/browse/index.tsx
+++ b/app/pages/browse/index.tsx
@@ -3,6 +3,7 @@ import { GetServerSideProps } from "next";
 import { SelectDatasetStep } from "@/browse/ui/select-dataset-step";
 import { AppLayout } from "@/components/layout";
 import { ConfiguratorStateProvider } from "@/configurator/configurator-state";
+import { useIsEmbedded } from "@/hooks/useIsEmbedded";
 
 export const getServerSideProps: GetServerSideProps = async ({ query }) => {
   return {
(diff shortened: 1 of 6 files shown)


Kronmar-Bafu (approved):
workable solution to the existing problem

app/pages/browse/index.tsx
```diff
@@ -13,8 +14,9 @@ export const getServerSideProps: GetServerSideProps = async ({ query }) => {
 };
 
 export function DatasetBrowser({ hideHeader }: { hideHeader: boolean }) {
+  const isEmbedded = useIsEmbedded();
   return (
-    <AppLayout hideHeader={hideHeader}>
+    <AppLayout hideHeader={hideHeader || isEmbedded}>
```
hupf, line 19: It is suboptimal, that part of this is evaluated on the server and part of it on the client and we can't use the actual `useOdsIframe` hook here... but I guess it is what how it is...

PR #2603, opened 2026-09-09
fix: xss vulnerabilities

mburri:
fixes some possible xxs vulnerabilties

see https://gitlab.lindas.admin.ch/bafu/bafu-visualize/visualize/-/work_items/787 for more details and how to reproduce


app/charts/map/map-custom-layers-legend.tsx
```diff
@@ -99,15 +104,7 @@ export const MapCustomLayersLegend = ({
...
-                      <Box
-                        sx={{
-                          "& > *": {
-                            // We do not let the tooltip HTML override the font size
-                            fontSize: `${theme.typography.caption.fontSize} !important`,
-                          },
-                        }}
-                        dangerouslySetInnerHTML={{ __html: layer.description }}
-                      />
+                      <CustomLayerDescription description={layer.description} />
```
hupf, line 116: @mburri That fixes the XSS issue, but it also breaks every formatting markup. Shouldn't we sanitize the description instead, like we did in other places? Or was there a decision, that no formatting is allowed here?

If formatting should still work, I'd introduce a sanitizing component, maybe leveraging the already present `sanitizeSchema`/`rehypeSanitize`? Otherwise DOMPurify would be the goto IMO.

Also, I assume there is another XSS in app/rdf/query-search.ts with the `highlightedTitle` and `highlightedDescription`.

